PR #235, opened 2014-07-04
Plain and match

vietjtnguyen:
The driving idea behind these changes was to support [`fzf`](https://github.com/junegunn/fzf) with some shell scripting for quickly fuzzy searching entries based on their title and returning the selected entry. My notes are stored so that the title line is a series of keywords as tags. I wanted to be able to quickly search through these keywords and then return a whole entry from the selection. Here is an example of the behavior I'm going for: [https://asciinema.org/a/10628](https://asciinema.org/a/10628).

With the changes in this pull request I get the intended behavior using (more or less) the following `bash` command:

```
jrnl --plain -shortmatch "$(jrnl --short --plain -a | fzf)"
```

`fzf` will let you fuzzy find through lines provided via `stdin`. `jrnl` sends these lines to `fzf` in `--short` and `--plain` format. The plain format is introduced because `fzf` will display the color escape sequence rather than actually coloring the text.

When a selection is made by `fzf` it returns the selected line to `stdout`. That output is then fed to `jrnl` to filter the entries to those whose short format is exactly the optional argument provided to `-shortmatch`. The `--plain` in the outer `jrnl` call is to prevent `jrnl` from line wrapping and prefixing lines with pipes so that the text can be copied and pasted more faithfully.

Though the plain and shortmatch features were written with this particular use case in mind, I believe that are fairly stand alone as well.

I think the plain format _should_ be a feature regardless to ensure faithful formatting of the original entry for things like piping.

The shortmatch feature is not particularly elegant, but I couldn't figure out another way of getting a very specific entry since `-to` and `-from` were behaving in an unexpected way (at least before [6f4ab85](https://github.com/maebert/jrnl/commit/6f4ab856d3145cc85084dfcbbe591bcd249b43c2)).

I'm interested to hear what you think!


gregorybodnar:
I like the idea of having access to the raw text of an entry for the purposes of scripting. I would get value out of that part of the PR. As it stands, I don't mind the tag highlighting, since `fzf --ansi` will handle the escape codes correctly. A fully rendered entry presumes a minimum terminal size, which tends to be dynamic in my daily use. An unrendered string would respond better to arbitrary environments.

I'd split out the shortmatch part of the PR, though. It's quite a different thing and might benefit from having its own PR and discussion.

PR #255, opened 2014-07-28
2.0-rc1: Add config option, display path & contents

yarko:
Two things caused me to want this:
- yaml config:  easier to read, easier to update, less error prone (thus I now have 6 broad-topic jrnl's)
- xdg to locate config;

Besides showing my jrnl's  (`jrnl -ls`), I'd like a reminder of which are encrypted, which are open, what the editor options are to each, and so forth...

If I want to make a change, I'd now (with xdg) need a little help finding the config.

I suppose this could be combined with an edit option (I chose not to, but I could see keeping a backup copy as a useful part, should someone care to do that).


Shir0kamii:
Hi,

I started using jrnl recently and were annoyed to see a journal.txt in my home directory. I searched if a compliance for XDG specifications was planned and I found this issue. Has this been abandonned or are you still working on this ?

barraponto:
@Shir0kamii you can configure where to save your journals. XDG compliance is about where to store `.jrnl_config`

Shir0kamii:
@barraponto I believe the `journal.txt` file should belong to the user data dir by default, even if you can override it in the `.jrnl_config` (Though I didn't saw this config file back then)

wren:

This looks like it might have value, but it's hard to see what's going on with the current merge conflicts.  Would you mind rebasing to the latest master?

PR #301, opened 2014-10-27
Add jrnl.plugins to the list of installed packages.

gcgoogle:
At the moment installing jrnl 2.0rc1 doesn't work because it doesn't install the plugins.  This patch fixes that.


maebert:
Merged in manually.

PR #314, opened 2014-11-22
Now the list of import types in --help is generated from the plugins.

gcgoogle:
Now the list of import formats shown in --help is generated from the list of plugins rather than being hard-coded.


maebert:
Merged manually - thanks!

PR #319, opened 2014-12-02
Highlight titles in pretty output

noqqe:
Hi,

I found that its maybe useful to have the title of each entry also a little bit highlighted. 

screen shot 2014-12-02 at 18 38 35

Hope you like it. 

PS: Could someone please be so kind to write a better RegEx for title parsing within the next 986 years?


maebert:
Hi, that's a really good idea and I'm going to put this in - however I want to get all of the displaying logic out of the `Entry` class and into an `Exporter` -- I'll close this for now and bring the title highlighting back in.

noqqe:
When? :P

Issue #259, opened 2014-07-31
Case conflicts (Day One)


bitic:
Hi,

Using jrnl 1.8.2 I launched the command: jrnl --edit

and it opened my Day One entries (my only journal) in a big file in Emacs (my editor).

As I didn't want to edit the whole journal, I quit Emacs (C-x C-#). 

This duplicated every entry in my journal with a Case Conflicted file.

I'm using
Python 2.7.3 in Debian 3.2.60-1+deb7u1 x86_64

nikvdp:
@maebert This issue was fixed in PR #325, should be safe to close

PR #325, opened 2015-01-10
Correct issue #259, also fix issue with specifying DayOne journals in install

nikvdp:
Fix two issues with DayOne compatibility:

**Duplicate entries when using Linux to edit DayOne journals**
Python's uuid.uuid1 returns a lowercase uuid string by default, but DayOne
saves it's filenames using an uppercase UUID by default.

On OS X (and I believe windows) filesystems case is preserved, but not
enforced, so using jrnl on a DayOne jrnl file works as expected.  On Linux
however, filesystem case is strictly enforced. As a result when jrnl and DayOne
interact duplicate files get created--one copy with uppercase UUIDs and one
with lowercase.

**Installer doesn't finish if you point it an an existing DayOne journal**
If you specify the top-level diectory of your DayOne journal during install
jrnl dies because it is expecting a file.


maebert:
*slow clap* Thank you, sir!

PR #326, opened 2015-01-12
Bug fix for arguments with quotes for the editor config value. Fixes #322

### LINKED ISSUE #322: Can't pass some options to vim
I'd like to use vim as my editor, but start it with spell checking turned on. I can do `vim -c 'set spell'`, but if I use that command as the editor setting, I get an error when vim starts:

    "spell'" [New File]
    Error detected while processing command line:
    E20: Mark not set

Is it possible to do what I want to currently; am I misunderstanding something? Or is this a jrnl bug? I see `subprocess.call(config['editor'].split() + [tmpfile])` in https://github.com/maebert/jrnl/blob/master/jrnl/util.py#L133. Maybe that `split()` is a naive way of preparing the command, because it shouldn't split in the middle of quotes?

jdavis:
I came across the same issue that was described in #322. I didn't particularly like the solution presented as it was using regular expressions when the subprocess module can already solve this. Instead, I just allow the `editor` config value to be either a string or a list. When it is a string, it does the `editor.split()` like it was doing before, but when it is a list, it just uses that.

Here is what my config value was that was causing it to fail:

```js
{
  "editor": "vim +'set ft=markdown'"
}
```

It was splitting on whitespace so the arguments ended up being: `["vim", "+'set", "ft=markdown'"]` which obviously isn't correct.

Now the fixed config with arguments can now be:
```js
{
  "editor": ["vim", "+set ft=markdown"]
}
```

And of course the original

```js
{
  "editor": "vim --some-arg-here"
}
```

will still work.

I tried to update the docs but don't know if there are any more places you wanted it updated.

Hopefully you like the solution. Let me know if there's anything that needs changed. Cheers!

### DIFF
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -18,7 +18,7 @@ The configuration file is a simple JSON file with the following options and can 
 - ``journals``
       paths to your journal files
 - ``editor``
-    if set, executes this command to launch an external editor for writing your entries, e.g. ``vim``. Some editors require special options to work properly, see :doc:`FAQ <recipes>` for details.
+    if set, executes this command to launch an external editor for writing your entries, e.g. ``vim``. Some editors require special options to work properly, see :doc:`FAQ <recipes>` for details. This value can either be a string such as the earlier ``"editor": "vim"`` example or ``"editor": ["vim", "+set ft=markdown"]`` if you want to pass arguments to the editor.
 - ``encrypt``
     if ``true``, encrypts your journal using AES.
 - ``tagsymbols``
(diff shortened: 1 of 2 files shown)


stuartc:
+1 I ran into this yesterday, would be useful in place of using a binstub in my path.

maebert:
Excellent! Could you rebase this against `2.0.0-rc1` please?

jdavis:
@maebert Done! I rebased this against `2.0-rc1` but the PR is still pointing to master so it's going to look ugly in this PR. The PR isn't updating, but if you do a manual merge from my branch, you'll see it is up-to-date with `2.0-rc1`: https://github.com/jdavis/jrnl/commits/editor-fix

maebert:
Excellent! Merged it manually, it's on 2.0 now!

jdavis:
@maebert Excellent!

Issue #330, opened 2015-02-06
Create new journals from the command line


mar-one:
I recently started using __jrnl__ for my coding/research projects to keep track of ideas/experiments/etc. Is there a more convenient way of initializing a new journal in the current directory other than going and editing the *.jrnl_config* file?

KarimPwnz:
I think we can implement an argument for this through #1068's options. It would be similar to how `-today-in-history` uses `-month` and `-year` (see https://github.com/jrnl-org/jrnl/pull/1145).

CC: @micahellison 

**Edit:** oh, I see that @micahellison already mentioned this in #1068.

PR #331, opened 2015-02-06
Add functionality for creating new journals from cmd

alshedivat:
See issue #330.


maebert:
This is great! I have three requests:

1. Can you rebase this against `2.0.0-rc1`?
2. can we change the parameters to only take the name (we'll then ask for the path)
3. Can we unify the creation process with `jrnl.install.install`? 

Afterwards happy to merge this in for the 2.0 release!

mar-one:
Sure:
1. I will close this PR and open a new one to `2.0-rc1`.
2. OK, that's easy.
3. What do you mean by unify? Do you mean change a bit the interface of `jrnl.install.install`, so it can take an additional argument (the journal name), and then call it from `run` if `-create` argument is provided?

PR #336, opened 2015-03-22
Fix improper charset handling in PY2 path of u(x)

ametheus:
Knowing fully that I may have just added another layer of impropriety, the above line fixed the charset errors I was getting.

I'll illustrate this change with an example string `Charšet` (note the caron above the S), entered (e.g. through stdin) in UTF-8 encoding.
To the best of my knowledge, the previous version would first have encoded this string to `Char\xc5\xa1et` (i.e., it encoded each byte not in ascii range as a hex escape code), and then have parsed this string to `CharÅ¡et` (i.e. after "r" it sees Unicode code point U+00c5 and U+00a1).

My version simply takes this str for what it is: an UTF-8 representation of the unicode string `Charšet`.

### DIFF
--- a/jrnl/util.py
+++ b/jrnl/util.py
@@ -66,7 +66,7 @@ def set_keychain(journal_name, password):
 
 def u(s):
     """Mock unicode function for python 2 and 3 compatibility."""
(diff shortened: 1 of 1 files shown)


maebert:
Thanks a lot! Can you rebase this against `2.0.0-rc1` please? :-)

PR #342, opened 2015-04-11
Add -s alias for --short, missing dependencies pyyaml and pyxdg

flight16:
Set up a fresh virtualenv and these two modules were used but missing from setup.py.


flight16:
This pull request implements #345 and adds a few missing dependencies from setup.py.

flight16:
Will work on a fix for the breaking test.  I swear tests passed when I ran before the commit.

flight16:
I am getting extremely weird behavior for the following two scenarios.  The only difference is the single quotes around the second date that cause it to pass.

<pre>
  Scenario: --short and -s display the short version of entries (only the title)  # features/core.feature:51
    Given we use the config "basic.yaml"                                          # features/steps/core.py:49 0.000s
    When we run "jrnl -on 2013-06-10 --short"                                     # features/steps/core.py:73 0.008s
    Then the output should be "2013-06-10 15:40 Life is good."                    # features/steps/core.py:147 0.000s
    When we run "jrnl -on 2013-06-10 -s"                                          # features/steps/core.py:73 0.003s
    Then the output should be "2013-06-10 15:40 Life is good."                    # features/steps/core.py:147 0.000s
      Assertion Failed: Output has 2 lines (expected: 1)
      Captured stdout:
      2013-06-10 15:40 Life is good.
      2013-06-10 15:40 Life is good.



Failing scenarios:
  features/core.feature:51  --short and -s display the short version of entries (only the title)

0 features passed, 1 failed, 0 skipped
0 scenarios passed, 1 failed, 0 skipped
4 steps passed, 1 failed, 0 skipped, 0 undefined
Took 0m0.012s
</pre>

<pre>
And the next passes with single quotes around the date

  Scenario: --short and -s display the short version of entries (only the title)  # features/core.feature:51
    Given we use the config "basic.yaml"                                          # features/steps/core.py:49 0.000s
    When we run "jrnl -on 2013-06-10 --short"                                     # features/steps/core.py:73 0.007s
    Then the output should be "2013-06-10 15:40 Life is good."                    # features/steps/core.py:147 0.000s
    When we run "jrnl -on '2013-06-10' -s"                                        # features/steps/core.py:73 0.003s
    Then the output should be "2013-06-10 15:40 Life is good."                    # features/steps/core.py:147 0.000s

1 feature passed, 0 failed, 0 skipped
1 scenario passed, 0 failed, 0 skipped
5 steps passed, 0 failed, 0 skipped, 0 undefined
Took 0m0.010s

</pre>

I have been debugging for an hour stepping through code but can't find the problem.  The interesting parts are features/steps/core.py check_output():

<pre>
def check_output(context, text=None):
    text = (text or context.text).strip().splitlines()
    out = context.stdout_capture.getvalue().strip().splitlines()
</pre>

and StringIO.py

<pre>
    def getvalue(self):
        """
        Retrieve the entire contents of the "file" at any time before
        the StringIO object's close() method is called.

        The StringIO object can accept either Unicode or 8-bit strings,
        but mixing the two may take some care. If both are used, 8-bit
        strings that cannot be interpreted as 7-bit ASCII (that use the
        8th bit) will cause a UnicodeError to be raised when getvalue()
        is called.
        """
        _complain_ifclosed(self.closed)
        if self.buflist:
            self.buf += ''.join(self.buflist)
            self.buflist = []
        return self.buf
</pre>

In the bugged version, in StringIO on the first step that runs jrnl, buflist is empty, but on the second run it contains the output from the find command.  This doesn't happen in the good version.  In both versions StringIO is the same instance in both steps.

It is even stranger that adding quotes around the date fixes the problem, but changing -s to --short doesn't.

I'm new to Python and behave so, although I'm not happy about it it, I'd prefer to submit this change with the hack of single quotes to avoid the bug.  @maebert Are you ok with the hack?

setup.py
```diff
@@ -89,6 +89,8 @@ def get_version(filename="jrnl/__init__.py"):
         "PyYAML>=3.11",
```
maebert: This should just be lower case instead!

setup.py
```diff
@@ -89,6 +89,8 @@ def get_version(filename="jrnl/__init__.py"):
...
         "keyring>=3.3",
         "passlib>=1.6.2",
+        "pyyaml>=3.09",
+        "pyxdg>=0.25",
```
maebert: `pyxdg>=0.19` is already defined above!

maebert:
The problem is a little bit tricky. Basically, when do the `the output should be` step it checks the output of the FULL scenario. So when you run it twice, the output will be two (identical lines). With the quotes it fails because for some unrelated reason the argument in quotes does not get passed on to jrnl (I just fixed that on `2.0` - now your tests should fail properly whether you use quotes or not ;-)

You'll have to use two different tests here.

```cucumber
    Scenario: --short displays the short version of entries (only the title)
        Given we use the config "basic.yaml"
        When we run "jrnl -on 2013-06-10 --short"
        Then the output should be "2013-06-10 15:40 Life is good."

    Scenario: -s displays the short version of entries (only the title)
        Given we use the config "basic.yaml"
        When we run "jrnl -on 2013-06-10 -s"
        Then the output should be "2013-06-10 15:40 Life is good."
```

I did this and merged this in manually. Thanks!

flight16:
I come from a Perl and C++ background and you can probably tell I'm new to Python.  Thank you very much for taking the time to explain the errors and point out the mistakes in my PR.

maebert:
No, thank you for contributing! This was a tricky problem indeed and took me some time to figure out, and heck, I _wrote_ the code ;-)

PR #349, opened 2015-05-05
2.0 -- yaml export fixes, dayone fixes

### LINKED ISSUE #322: Can't pass some options to vim

### LINKED ISSUE #348: `shlex.split` breaks on Windows paths
#322 replaced this [line](https://github.com/maebert/jrnl/blob/master/jrnl/util.py#L133)
```py
subprocess.call(config['editor'].split() + [tmpfile])
```
with
```py
subprocess.call(shlex.split(config['editor']) + [tmpfile])
```

However, `shlex` is only designed to work on POSIX systems on so breaks on Windows. On Windows, if you specify an editor using 'normal' Windows paths (i.e. using backslashes, like 'C:\Program Files (x86)\Notepad++\notepad++.exe`), it gives the un-helpful error:
```cmd
FileNotFoundError:[WinError 2] The system cannot find the file specified
```
To the (Windows) user, this is strange because paths to journal files can be given 'as normal' (i.e. with backslashes) without issues.

Four possible solutions come to mind:
1. revert #322
2. use `shlex` on POSIX systems and not on non-POSIX systems
3. use double backslashes or (single) forward slashes in the configuration
4. find a replacement for `shlex` that works on both POSIX systems (e.g. Mac) and Windows

I don't like 1, because although it is simplest for me personally, I know it was meant to fix a problem for others.

I don't particularly like 2, because it increases the complexity of the code and makes it nearly impossible for one person to test both code branches (that inability to test is what brought us here).

3 is workable, but is non-intuitive at best.

My preferred options would be 4, but I only know how to do the first three at the moment. 

@pcarranza @philipsd6 @maebert  Suggestions?

MinchinWeb:
This pull request fixes/adds:

- `shlex.split` on Windows (fixes #322, fixes #348)
- CLI
    - if no editor is set, running `jrnl` allows you to create an entry, but no prompt was being given
    - colorizes warning and error messages throughout
- improves Dayone handling:
    - moves to new-style date handling (variable length, square bracket enclosed)
- improves YAML export handling
    - only increases headings by one level
    - includes the entry's UUID (if available) in the export
    - removes lines from the entry if they only list tags


maebert:
dis gon b gud

Wow. So much goodness in here indeed. Merging it in now, thank you!

MinchinWeb:
@maebert , your responses get better and better! Glad to help

PR #350, opened 2015-05-10
[Dayone] remove extra spaces from the titles of edited DayOne entries

MinchinWeb:
Otherwise, a leading space was being introduced


MinchinWeb:
Now also maintains existing metadata (location, weather, creator) on Dayone entries across edits.

MinchinWeb:
Manually resolved merge conflict with #355

flight16:
Thanks for the leading space fix.  I encountered this issue, too.  Looking forward to it getting merged back into the 2.0 branch.

wren:
Is this PR still relevant? It seems like fixes for Day One before their 2.0 is outdated, but I honestly don't know if people still use the older Day One client.

MinchinWeb:
@wren I actually use a DayOne Classic compatible client on Windows. (DayOne has never released official client for Windows).

wren:
@MinchinWeb Which client? Are there any similar client for Mac? Do you know if Day One supports their old client anymore?

MinchinWeb:
I'm using [Journaley](https://github.com/yyoon/Journaley). I don't have a Mac, so I've never gone looking for something similar there. I rather doubt Day One is still supporting their old clients; it seems they haven't updated it to work with Dropbox's new API.

This also helps with old entries, written in the original DayOne (iPad/iPhone) client when it was still current.

MinchinWeb:
@ wren this has been rebased against the current master, add all test are passing. It should be ready to be merged!

These are all bug fixes, so I'd love it if it made the 2.0 cutoff.

MinchinWeb:
We may need to review what we're doing with line endings. This is one of the raw strings that was being testing against:

`'2010-06-10 15:00 A life without chocolate is like a bad analogy.\r\n\n2013-06-10 15:40 He said "[this] is the best time to be alive".\r'`

(notice the total mis-match of line endings.) Some of it, I'm sure, is due to Windows, and some due to Git, but it was making for tests that passed on Travis, but failed locally.

wren:
I'm going to close this for a bit, while we wait for the day one-related tests to come back.

PR #355, opened 2015-05-23
Fix: DayOne mixed-case tags don't work

flight16:
When filtering on DayOne tags that contain upper case, no entries are
returned. maebert/jrnl#354


maebert:
Merged manually, thanks!

Issue #356, opened 2015-05-25
Poor DayOne performance for large journals


flight16:
I tried using jrnl with my DayOne journal of 2500 entries on a brand new, 2015 MacBook Pro with SSD: 4.5 seconds of wall time to list the most recent entry (CPU time was close).

```
$ time jrnl dayone --short -n 1
2015-05-18 13:41 My latest entry.

real    0m4.469s
user    0m4.007s
sys 0m0.225s
$
```

The problem appears to be that jrnl loads the entire file into memory, sorts, and then extracts the last entry.  Using an index file or persisting data to an easier format to search (eg. sqlite) should improve performance greatly.

Related:  #334

flight16:
This is going to be a bigger change than I initially expected.

Jrnl's design assumes entries are loaded into the Journal object on open, then they are removed when filtered.

What an optimized DayOneJournal would do is lazily read entries and apply the a filter when .entries is accessed.  But that makes it a read-only attribute, and it is being written to in many places.

I think breaking .entries out into a get_entries(filter) and set_entries(entries) would be the best choice because it means file reading and filtering can be deferred until it is actually needed, and the (new) index file can get the most usage.

@maebert , does that refactor sound ok?

flight16:
@maebert This is how I'm thinking to implement the index so it can be shared between PlainJournal and DayOneJournal.  It requires a small refactoring to filter().  I am interested in your comments.

    Journal: # What currently exists
        def open(...) # No longer reads the file.
        def set_filter(...) # Replaces filter(), affects filtered_entries attribute
        filtered_entries # Read-only attribute; result is affected by set_filter(); lazy-loaded
        entries # Read/write attribute; always returns all entries; lazy-loaded

    IndexedJournal: # New class 
       # Overridden, implemented using the index.  Calls entry_from_index(key) of subclasses
        filtered_entries 
        index_entry(entry) # Adds entry to index
        deindex_entry(entry) # Removes entry from index

        # Required to be implemented when subclassed 
        # An index key is a json object that can be used to extract the specific entry
        # from the raw journal file.  It will be stored in the sqlite index file.
        def entry_to_index_key(entry) 
        def entry_from_index_key(key)
        def find_entries_to_reindex() # Called every run on Journal open.

    DayOne(Journal, IndexedJournal): # Existing class
        def entry_from_index_key(key) # key is DayOne's UUID
        def entry_to_index_key(entry) # return entry.uuid
[...]

Issue #334, opened 2015-02-17
Pickleing?


schollz:
Hi maebert!

I love this code. I was about to write my own, but I think yours accomplishes everything I could possibly want! 

I'd like to help out to improve, if I can. I was thinking about contributing something to help speed up the code. I found that a 600k file took about 11 seconds everytime I wanted to access it. I dropped in a bunch of printed time statements, but couldn't find the main culprit. Do you know the limiting step?

My guess is that that the slow step might be the loading/sorting of the entries. In this case I think it may be useful to store/load a "state" file via Pickle to help the loading speed.

Thanks!!

maebert:
Hi, thanks for your feedback! Is your journal encrypted?

I think parsing and sorting takes a lot of time indeed. If you're familiar with profiling in python, go right at it! If not, great time to learn it! Huy Nguyen had a nice article on line-by-line profiling a while back: http://www.huyng.com/posts/python-performance-analysis/

flight16:
I just tried using jrnl with my DayOne journal of 2500 entries on a brand new, 2015 MacBook Pro.  4 seconds of CPU time to list the most recent entry:

<pre>
$ time jrnl dayone --short -n 1
2015-05-18 13:41 I played with Eston today and made him lunch.

real	0m4.469s
user	0m4.007s
sys	0m0.225s
$
</pre>

I agree that some sort of caching could be done.  In my opinion, sqlite would be best, because it could be selectively updated when entries change.  The solution would be such that it could be applied to any journal type.

I will start poking around at this, but don't expect anything in the next couple weeks.

maebert:
Hmmm. I think for journal types with individual files it maybe makes more sense to build an index file with the dates and filenames. Currently writing an entry or loading the most recent ones requires loading ALL entries into memory, which is entirely unnecessary.

flight16:
I'm with you on there being no need to load everything into memory like is happening now.

[...]

PR #357, opened 2015-05-28
Improve cli.py readability and reduce complexity.

flight16:
Refactored implicit run modes into explicit, individual commands.

It is much easier to see, at a glance, what is being executed when each command is split up into its own local function.  It is also less error-prone.

This is pre-work to simplify the implementation of #356 and #334.

### DIFF
--- a/jrnl/cli.py
+++ b/jrnl/cli.py
@@ -18,6 +18,7 @@ import jrnl
 import argparse
 import sys
 import logging
+import os
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


jrnl/cli.py
```diff
@@ -53,26 +54,183 @@ def parse_args(args=None):
...
-    import_ = False
-    if args.import_ is not False:
-        compose = False
-        export = False
-        import_ = True
-    elif args.decrypt is not False or args.encrypt is not False or args.export is not False or any((args.short, args.tags, args.edit)):
-        compose = False
-        export = True
-    elif any((args.start_date, args.end_date, args.on_date, args.limit, args.strict, args.starred)):
+class CommandEnum:
```
maebert: Python 2.x needs `class CommandEnum(object)` here for new-style classes.

flight16: Thanks for the comment.  I made CommandEnum a new-style class to keep things simple.  You will find the new commit in this pull request.

wren:

This feature looks useful and we'd be interested in merging it. Do you think you could resolve the merge conflict for the PR?

PR #359, opened 2015-06-02
Remove dead code

flight16:
I came across these two pieces of code that weren't being used while working on a refactor for DayOneJournal's index.


maebert:
Merged manually, thanks!

PR #376, opened 2015-08-28
Make line wrapping more clever about whitespace

### LINKED ISSUE #222: Line wrapping should be more clever about whitespace
So, stumbled upon this project on HN, just installed it.

I could easily fall in love. Ignoring a random limitation that bugs me (not jrnl's fault, but .. a ' is a) common in english text and b) not usable without quotes on the command line) I think this is ugly and could be improved:

````
$ jrnl 
[Nothing saved to file]
2014-06-28 19:34 So I really like this idea.
| I've lost track of what I did and wanted to do quite some time ago, let's see
|  if this tool can be used to bring order to chaos. I'm not overly convinced, 
| but if this doesn't work, the _tool_ is not to blame I guess....
````

Note that the line wrap ends up starting a line with a whitespace. I like the wrap at 73 (or whatever) characters, but spaces probably should be swallowed in that case. Is this intentional?

### PR DESCRIPTION
**Pull request by philipsd6**
This PR fixes #222, which has been bugging me forever. It also makes the line indentation character configurable, so you can put the BOX DRAWINGS LIGHT VERTICAL unicode character (│) in your `~/.jrnl.config`:

```json
{
  "linewrap": 79,
  "indent": "│",
}
```
This makes the jrnl output look pretty slick, assuming your font/terminal supports unicode characters.

### DIFF
--- a/jrnl/Entry.py
+++ b/jrnl/Entry.py
@@ -50,8 +50,9 @@ class Entry:
             body = "\n".join([
                     textwrap.fill(line,
                         self.journal.config['linewrap'],
-                        initial_indent="| ",
-                        subsequent_indent="| ") or "|"
+                        initial_indent=self.journal.config['indent'] + " ",
+                        subsequent_indent=self.journal.config['indent'] + " ")
+                            or self.journal.config['indent']
                     for line in self.body.rstrip(" \n").splitlines()
                 ])
         else:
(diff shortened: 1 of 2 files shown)


fortunat:
Thank you for your decision .

philipsd6:
Well, my only decision was to make this pull request, we need @maebert to actually decide to integrate it!

I really hope you do this sometime soon @maebert, even though this project is obviously very low-brow for a neuroscientist... ;)

And when you finally get around to taking a look at this, maybe you want to consider sharing maintainer status with another interested party with the appropriate skillset? (Not me!) I'd not like to see this project languish, or get lost in a forest of forks...

maebert:
Alright, turns out the solution is much simpler indeed! See the fix for #222

Issue #358, opened 2015-05-31
DayOne: Editing existing entries with tags removes the tags when saving


flight16:
1. Create a DayOne entry using DayOne.app and add a tag using the tag button at the top of the entry (don't include the tag in the body).
2. Click save and see that the entry is now tagged.
3. Close DayOne
4. Run jrnl dayone -n 1 --edit and edit the entry you just added, and save it.
5. Open DayOne and note that the latest entry is no longer tagged.

flight16:
I think this might have been caused by fd46ffea23eb1cb98121bde3a3993a4fa56e1afb, but I haven't had time to investigate yet.

I think this shows there are two tests that are lacking:
0. A test to flex the --edit code path which launches an external editor and make sure data round-trips OK.
0. A test to verify the format of saved DayOne .xml to make sure it conforms to DayOne's expected format.

MinchinWeb:
I've been able to confirm the issue.

I think it is somewhat related to #351.

It actually exists much earlier. It exists as of ea6060d and as of 3912693, which dates to last September. Which makes me wonder if it has ever worked...

---

The problem seems to be here -> [DayOneJournal.py](https://github.com/maebert/jrnl/blob/8a6563d958776cef47fe3f88564602464319f18f/jrnl/DayOneJournal.py#L132) which calls this function -> [Entry.py](https://github.com/maebert/jrnl/blob/8a6563d958776cef47fe3f88564602464319f18f/jrnl/Entry.py#L25)

---

Does 8f12d62 fix it?

MinchinWeb:
waiting on merge of PR #350

wren:
@MinchinWeb Sorry, but it looks like #350 has merge conflicts again.

Honestly, I'm having some trouble with what to do with Day One functionality. I'm hoping we can talk it through a bit, since it seems to be an important feature for you.

The way I see it, the Day One functionality in jrnl needs some work to get it, and its tests, up to snuff. And then it also needs someone that can seriously commit to maintaining it long term. When I first started fixing up jrnl, I have to admit that the Day One functionality seemed like a top candidate for cutting out (for a variety of reasons that I'm happy to get into, if you're interested).

That being said, I want to respect the existing community of users, and especially existing contributors. If there's a feature that you value this much, I want to make sure it gets the attention it deserves.

[...]

Issue #159, opened 2014-04-18
DayOne issue and improvement


hs6705gc:
Hi!

Thanks for the solution with the previous issue!
I now manage to list and add entries to the DayOne journal.

There is an issue however:
When a new entry is added a CRLF is added to the entry! DayOne recognizes
such entries as headers and display them in bold text! So all entries added 
with jrnl displays in bold!                                                 

Then I have a feature request! There is a location added to each entry.
Obviously jrnl does not implement this but it would be nice if there was an
option in the .jrnl_config to set a static location that was added to the
entry each time! Parameters to consider is:
Administrative area
Country
Latitude
Locality
Longitude
Place Name

Code inserted if the parameters exists in then .jrnl_config file:

``` xml
    <key>Location</key>      
    <dict>
        <key>Administrative Area</key>
        <string>Ostergotlands lan</string>
        <key>Country</key>            
        <string>Sverige</string>          
        <key>Latitude</key>
        <real>58.383400000000000</real>
        <key>Locality</key>
        <string>City</string>          
        <key>Longitude</key>
        <real>15.577170000000000</real>
        <key>Place Name</key>
        <string>Street</string>     
    </dict>
```

Thanks in advance,
Henrik

MinchinWeb:
Currently, if you edit a Dayone entry, any existing location metadata is dumped. At a minimum, it would be nice to keep this existing metadata.

PR #377, opened 2015-09-06
Don't drop extra DayOne metadata when editing a DayOne journal

nikvdp:
DayOne stores a lot of extra data in it's journal entries such as location, and weather, etc., but jrnl doesn't retain this data when editing a DayOne journal entry. 

Added an extra class for DayOne entries which stores the data, and updated the `editable_str` and `parse_editable_str` methods to allow storing of the extra data as zlib compressed base64 data.


MinchinWeb:
Just for reference, another approach to dealing with this issue (#358, see also #159) is containing in PR #350.

wren:
@MinchinWeb It looks

Issue #380, opened 2015-09-24
Feature Request: "head"-style viewing on a per-entry basis.


zdravi:
I've adopted a method of keeping notes in jrnl that involves the format:

```
Short synopsis
@+specific @-types @*of @#tags

Body of the journal
```

This works well for me, but the issue I've run into is not being able to view the synopsis and tags without cluttering up the title. --short obviously drops everything but the title, but every other output format results in the entire entry being dumped to screen. 

To work around this, I'm working on a changeset at that adds the "-t" flag, which works in a similar manner to "head -n" on a per-entry basis. It currently counts "lines" as entered by the user in the journal itself instead of lines as split by jrnl on display.

If I'm the only person interested in this feature, then that's fine too, but I thought I'd throw it out there in case anyone else likes having uncluttered titles.

zdravi:
I'm working on this in https://github.com/zdravi/jrnl/tree/slice-top, but after that false start I'm going to hold off on making a PR unless someone's actually interested in looking over it/using it.

maebert:
Hi,

I'd prefer to not add more command line options, but I'm working on a more generic export option that would allow this!

PR #381, opened 2015-09-24
 Feature Request: "head"-style viewing on a per-entry basis. #380 

zdravi:
The changeset mentioned in #380


zdravi:
And of course as soon as I click submit PR, I run into a bug with my own code.

Issue #392, opened 2016-01-04
Can't encrypt


heitorsilva:
I'm not being able to encrypt my file, after editing the txt on Sublime Text 3, on Windows 10.

Getting the following error:

```
Traceback (most recent call last):
  File "c:\python27\lib\runpy.py", line 162, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "c:\python27\lib\runpy.py", line 72, in _run_code
    exec code in run_globals
  File "C:\Python27\Scripts\jrnl.exe\__main__.py", line 9, in <module>
  File "c:\python27\lib\site-packages\jrnl\cli.py", line 257, in run
    encrypt(journal, filename=args.encrypt)
  File "c:\python27\lib\site-packages\jrnl\cli.py", line 76, in encrypt
    password = util.getpass("Enter new password: ")
  File "c:\python27\lib\site-packages\jrnl\util.py", line 31, in getpass
    return gp.getpass(prompt)
  File "c:\python27\lib\getpass.py", line 95, in win_getpass
    msvcrt.putch(c)
TypeError: must be char, not unicode
```

maebert:
Fixed for 2.0 with 06e3f0c101cc402139a7741773ce2e881bf7d373

makais:
Is it possible for this fix to be back-ported to version 1, or for a version 2 build to be made available?  Not being able to encrypt on Windows is a showstopper.

PR #393, opened 2016-01-06
Fix for getpass on Windows

heitorsilva:
I've opened the issue #392, which I'm not able to use --encrypt on Windows, because I get a 'must be char, not unicode' error.

This is the fix for that... Works beautifully on a Windows 7!


heitorsilva:
Also works on Windows 10!

maebert:
I fixed this for the upcoming 2.0 release - thank for pointing this out!

PR #417, opened 2016-04-07
Fully strip title and body after SENTENCE_SPLITTER search

philipsd6:
This fixes the problem where an exported entry will have extra whitespace depending on if it was the last entry or a previous entry.

For example, a git hook that commits changes will always commit a new whitespace change for the previous entry with every new entry.


maebert:
Good catch, merging this.

PR #418, opened 2016-04-07
Add exporter to output entries inside unicode box character boxes

philipsd6:
Just for fun, I created an exporter that formats entries into unicode boxes. It's even more fun when used with [busyloop/lolcat](https://github.com/busyloop/lolcat)!

I only got this to work using `PYTHONIOENCODING=utf-8` though. 

It looks like this:
image


maebert:
Haha, brilliant. Need to think about this for a second, because I'm working on exporters too. But I definitely like the idea!

wren:
I like this! We need a bit more time to look through everything, but this shouldn't go stale just yet.

Issue #401, opened 2016-01-25
Feature request: configurable indent character for unicode line drawing characters


philipsd6:
Instead of having a hardcoded pipe symbol used for the indent character, I would like to be able to use a unicode line drawing character, such as the BOX DRAWINGS LIGHT VERTICAL unicode character (│). 

I have implemented this in my fork: philipsd6/jrnl@ad4d2ed71bcae05cbec6a5ee9c6e97de998f7f5a

It looks like this:
![image](https://cloud.githubusercontent.com/assets/908201/12562515/95550f06-c373-11e5-97df-800d012d1e2e.png)

My branch requires #399 and #400 to be merged first before it will merge cleanly into 2.0-rc1. The functionality itself is backwards compatible (as submitted in #376)

I've picked `indent_character` for the configuration variable, but I'm open to changing it, since it seems a little long.

PR #419, opened 2016-04-07
User defined indent character

philipsd6:
I really like unicode box characters... This PR allows the user to define what character to use for the indent. I like the light vertical box char -- on my terminal, it displays as an unbroken line on the left margin -- much nicer than the pipe char.

### DIFF
--- a/features/data/configs/basic.yaml
+++ b/features/data/configs/basic.yaml
@@ -9,3 +9,4 @@ linewrap: 80
 tagsymbols: "@"
 template: false
 timeformat: "%Y-%m-%d %H:%M"
+indent_character: "|"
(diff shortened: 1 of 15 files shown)


maebert:
You're on a roll, sir!

philipsd6:
Sadly, this latest commit f493e37 actually made things blow up! Here's why: `install.py` sees the new 
`indent_character` key in the config dict and automatically updates the config file, and then prints this:

    "[Configuration updated to newest version at {}]"

Which breaks the feature that checks to see if the output matches, because it doesn't have that update message…

I'll fix this by adding the new key to all of the config files in `features/data/configs/`

philipsd6:
BTW, this resolves #401

maebert:
LGTM!

PR #420, opened 2016-04-07
Clean up text wrapping once and for all --- Please!

### LINKED ISSUE #222: Line wrapping should be more clever about whitespace

philipsd6:
The current behavior of text wrapping continues to be borken, despite my previous attempts to submit fixes. This PR solves *all* the pretty printed text wrapping issues, very pythonically I think, and fixes #222 more simply than any other existing fix I've seen, such as 060565c.

After this PR, with perfect whitespace and indent handling:
image

Before, on the current 2.0-rc1 branch, with `drop_whitespace=True`:
image

Before, on the master (1.0) branch, with `drop_whitespace=False`:
image

### DIFF
--- a/jrnl/Entry.py
+++ b/jrnl/Entry.py
@@ -78,11 +78,11 @@ class Entry:
             title = textwrap.fill(date_str + " " + self.title, self.journal.config['linewrap'])
             body = "\n".join([
                 textwrap.fill(
-                    (line + " ") if (len(line) == 0) else line,
+                    line,
                     self.journal.config['linewrap'],
                     initial_indent="| ",
                     subsequent_indent="| ",
-                    drop_whitespace=True)
+                    drop_whitespace=True) or "| "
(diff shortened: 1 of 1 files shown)


maebert:
Yes!!! :bow:

PR #423, opened 2016-05-05
Change README url to point to actual documentation

Davidslv:
Instead of pointing to homepage.

Thank you


maebert:
Yup, thanks!

PR #435, opened 2016-08-19
editor: use shell to launch editor to allow spaces in editor command

gauteh:
If I want to set a filetype in vim on opening the journla, e.g.:

  vim -c 'set ft=notes' -c Goyo -f

the set and ft=notes will be split and interpreted as two files in vim.
Using the shell to execute the command makes it possible to have editor
commands with quoted/escaped spaces in them.


wren:

This is a good catch! Would you mind rebasing this PR against the new master so we can take a closer look and merge?

Thank you!

gregorybodnar:
I was able to test this out by cherry-picking 460179efdde3a660355ba8d0cc808208640425a8 onto 2.0-rc3. Obviously, there's a conflict preventing direct merging, but it adequately covers #581 and makes sense for Windows users, which will convert a sequence to string anyway before execution.

If  @gauteh is unable to update the PR to satisfy the conflict, I can use the branch I tested on as a new PR.

Issue #441, opened 2016-10-03
Configure more characters to be allowed inside tags


sprax:
Currently, the tag strings @M&M and @4:20 create tags M and 4, because the ampersand and colon are not in the set of characters allowed inside tags, namely [-+*#/\w].
Likewise, the strings @:-), and @^_^  create no tags at all.
No doubt some users would prefer that a colon terminate a tag rather than be included within it, but letting them add "extra" characters to those already permitted inside tags could give them control of this preference.  I have coded this up, using a "tag_extras" field in the config file, and will submit a PR.

wren:
Agreed. We should fix this. 

@sprax Any chance you could update your PR to fix the merge conflicts?

wren:
If the associated PR doesn't get updated, we should still fix this.

wren:
My initial inclination is to do anything starting with `@` and surrounded by whitespace should be a tag. There's some questions around whether someone can end a sentence in a tag, though, in which case we have to exclude some characters. 

Example: 
`This is @mytag.`

micahellison:
It sounds like we'd have to drive this through a config field like the OP mentioned, or else it would be a breaking change -- for instance, if someone ends sentences with a "@vacation" tag, a change to this behavior might lead to other tags that include "@vacation." and "@vacation," when the user really only wanted the one tag.

PR #442, opened 2016-10-03
Tag config

sprax:
PR for this issue:
Configure more characters to be allowed inside tags #441


wren:

This is a great catch and a great fix. Would you mind rebasing onto the updated master so we can get it merged in?

Thank you!

PR #447, opened 2016-12-17
Fix failing behave tests

radomirbosak:
The keyring package broke backward compatibility in version 8.0 by
moving some keyring backends to another package, keyrings.alt, as
documented in the [changelog](https://pythonhosted.org/keyring/history.html#id22).

This change broke behave tests which were trying to use
keyring.backends.file.PlaintextKeyring - no longer existing in keyring
package.

This commit adds the keyrings.alt package as dependency so that the
PlaintextKeyring class can be used.

### DIFF
--- a/features/steps/core.py
+++ b/features/steps/core.py
@@ -5,7 +5,8 @@ import os
 import codecs
 import json
 import keyring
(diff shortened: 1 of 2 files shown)


maebert:
Excellent, thank you!

PR #448, opened 2016-12-17
Add ability to search in entry title and body

### LINKED ISSUE #384: List of --short entries on a full text search query on a journal
It would be nice if there was a full text search feature which returned just a short listing of entries. While I realize you can _kinda_ get a proxy for this with something like `jrnl -n 10 | grep awesome` or the like with the addition of the `|` in front of the entries is not quite what I'm going for. I'm looking 

Effectively, is there a way to get the equivalent of a full text search switch (much like tags) which returns a list of --short entries  where the search terms is in the body (or title of the jrnl entry).

ie. something like

```
jrnl -n 3 --short --search "awesome"
```

which yields a list 

```
2015-11-10 23:10 This entry mentions awesome in the title.
2015-10-10 22:12 This one only mentions it in the body of the text.
2015-09-09 20:12 This entry mentions it only in passing.
```

For some weird reason, this feels like a major missing feature (though, of course, it probably is only an issue for those of us using encryption and not being total fans of grep.  =p 

Daryl.
PS> If I've just missed something obvious in terms of getting this output with a command line pipe, do let me know. Have not had anywhere near enough coffee this morning.

radomirbosak:
jrnl was missing ability to search in entry titles and message body.

This commit adds a `-search TEXT` option which, if specified, causes
jrnl to display only those entries which have `TEXT` in either their
title or message body.

Specifying the `-search` option also implies that the compose mode
won't be used.

Fixes #384.


maebert:
That's a great feature. Sorry for only commenting now. Could you rebase that against `2.0-rc1` though? That's where new features should go.

radomirbosak:
Ok, I'll rebase against `2.0-rc1`, but it might take some time as there are merge conflicts.

maebert:
Thanks a lot!

radomirbosak:
Rebased and reopened as #470

Issue #398, opened 2016-01-20
Entries as separate files


pca2:
Love jrnl. Would it be possible to add support for storing jrnl entries as separate files in a designated folder? That would make it much easier to edit entries from mobile apps.

aisapatino:
@pca2 there's a similar issue (suggesting file per day, rather than per entry) here: #170

Issue #170, opened 2014-06-24
Store Journals in folder (with multiple files)


matthiasbeyer:
Hi,

I think it would be great if the program would use several files instead of just one file. One file per day would be feasable, I think!

matthiasbeyer:
Will this be implemented? It's open for over half an year now... and it seems nothing has happened...

notbalanced:
Hey, I've got something working on my fork of jrnl, in the "folder" branch. There are some bugs with deleting/editing entries, still, so not really ready for prime time, but if you'd like to try it out...feel free.

flight16:
This issue is important to me, too.  I just found jrnl and it looks like everything I wanted, except the directory/file structure.  Is this still being considered by @maebert ?  Is there anything I can do to help?

matthiasbeyer:
@flight16 :+1:

flight16:
@notbalanced I had a look at your change, and it appears you added the changes to the base Journal class.  Just curious, what issues did you run into when editing and deleting entries?  Were these implementation bugs or design issues?

I am not the author of jrnl, but I think we could cleanly solve this issue by enhancing cli.py and the config to support arbitrary journal formats.  There is already DayOneJournal that is hard-coded.  All that needs to be done is to create a mechanism to optionally tag a journal as a specific format, and then instantiate the associated class.  So I see this issue as "Support custom journal formats".

@maebert Would you accept a pull request for the above enhancement?

notbalanced:
@flight16 Hi, my implementation is actually a new Folder class derived from the Journal class (similarly to what is done with the DayOne class.) In the cli.py, I do instantiate the Folder class if the journal in the .jrnl_config is a folder that **doesn't** have ".dayone" in the folder name.

 It works just fine for adding entries to the journal. But when you use the --edit flag, and then delete an entry, it doesn't get deleted. (You'd have to go into the txt file and delete it manually.) I run into this problem because I can have multiple entries in one file (unlike the dayone journal that only has 1 entry per file). I haven't had time to figure out a way to identify which entry gets deleted, so that it doesn't get written when you update the file.

There is also the issue that if you only have 1 entry in a file and delete that entry, that the code doesn't delete the file.

Finally, I haven't had the time to run the tests on the update. I'm pretty sure I've broken something there, so I need to update those tests and add additional tests to cover this new journal type.

matthiasbeyer:
Please note that I'm still interested in this.

wren:
[...]

PR #458, opened 2017-01-14
Issue #170 - Add folder based journal

notbalanced:
Adds feature for issue #170 (and #398) where you configure your journal to be a directory and entries are added as sub-directories and files: yyyy/mm/dd.txt. Multiple entries in a day will go in the same file, but a
new entry for a specific day will create a new file (and directory structure).


setup.py
```diff
@@ -89,6 +89,7 @@ def get_version(filename="jrnl/__init__.py"):
         "keyring>=7.3",
         "passlib>=1.6.2",
         "pyxdg>=0.25",
+        "asteval>=0.9.8",
```
MinchinWeb: I don't think this (`asteval`) is being used...

notbalanced: Yeah, you're right. It isn't in my checkin, but if you look at this
[checkin](https://github.com/maebert/jrnl/commit/8b64cc95d9e5171b7b914451f8538783b77c2c71), it adds `asteval` in [template.py](https://github.com/maebert/jrnl/commit/8b64cc95d9e5171b7b914451f8538783b77c2c71#diff-46697bd48828cb658c9837ea175fab2fR2).

When the Travis build ran, it was failing, so I added it to fix that dependency.

I didn't know how to handle that. Is there a better way of doing that?

MinchinWeb: No, this is probably a great way to handle it. Great catch.

(If anyone else is following up on this, `template.py` has been moved to [jrnl/plugins/template.py](https://github.com/maebert/jrnl/blob/2.0-rc1/jrnl/plugins/template.py#L2))

wren:
This seems very useful!

notbalanced:
I'd still like to see this merged in with jrnl.

wren:
Hi, @notbalanced. I would also like to see this merged in. Do you have a chance to resolve the merge conflicts?

notbalanced:
Merge conflicts have been resolved and I've updated the tests so that they pass correctly.

wren:
Thank you, @notbalanced! I'll take a look through this at my next jrnl work session (next Saturday).

PR #464, opened 2017-02-12
Update iA Writer bundle id

pwyliu:
Hi, 

I've recently started using Jrnl and I like it a lot! Thanks so much for making this great software.

I'm using iA Writer as my editor, and I noticed the instructions in the jrnl docs are a little bit out of date. I just bought the newest version and it seems the bundle identifier has changed. This PR updates the id and adds some instructions on how to find a bundle id, in case it changes again.


maebert:
Excellent, thanks!

PR #469, opened 2017-03-07
Fix behave tests by adding asteval python dependency

radomirbosak:
Behave tests were failing in `2.0-rc1` branch with 
```
ImportError: No module named 'asteval'
```

This commit adds the `asteval` dependency to `setup.py` to fix the failing tests. I chose the required asteval by setting it to my current installed version, because I didn't know what other version to put there. Feel free to change it.

Behave still fails for python 2.7, but that seems to be an unrelated issue.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -89,6 +89,7 @@ setup(
         "keyring>=7.3",
         "passlib>=1.6.2",
         "pyxdg>=0.25",
+        "asteval>=0.9.8",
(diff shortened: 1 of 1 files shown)


radomirbosak:
The other problem for python 2.7 that I mentioned should be fixed by #471

maebert:
Thanks!

PR #470, opened 2017-03-07
Add ability to search in entry title and body

### LINKED ISSUE #384: List of --short entries on a full text search query on a journal

radomirbosak:
This merge request replaces #448 , which wasn't rebased against branch `2.0-rc1`.

* The first commit fixes behave tests for `python3`.
* The second commit adds a feature to search the journal entries (their title and text body) by using the `-S` and `-search` CLI switches. (`-s` is already used for `--short`)

Fixes #384

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -98,6 +98,17 @@ the last five entries containing both ``@pineapple`` **and** ``@lubricant``. You
 
   ``jrnl @pinkie @WorldDomination`` will switch to viewing mode because although **no** command line arguments are given, all the input strings look like tags - *jrnl* will assume you want to filter by tag.
 
+Searching
+---------
+
+To search for a string ``"excellent idea"`` in all journal entries, use the ``-search`` argument::
+
+    jrnl -search "excellent idea"
+
+The ``-search`` argument can be combined with other filters, such as ``-from`` and ``-until``.
+
+Note that the searching is case-insensitive and doesn't accept any wildcards or regular expression syntax.
+
 Editing older entries
 ---------------------
 
--- a/jrnl/Journal.py
+++ b/jrnl/Journal.py
@@ -165,7 +165,7 @@ class Journal(object):
         tag_counts = set([(tags.count(tag), tag) for tag in tags])
         return [Tag(tag, count=count) for count, tag in sorted(tag_counts)]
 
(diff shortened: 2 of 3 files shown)


maebert:
I like that, but what's the rationale for using `-p` as a shorthand? Any other applications that are using this?

maebert:
This looks great, but what's the rationale for using `-p` as a shorthand?

radomirbosak:
I was thinking `-p` as in **p**lain search or **p**hrase (as suggested in #313 ). My first choice would be naturally `-s`, but that one is already used. Feel free to suggest/change the searching shorthand to anything else.

Other option would be using `-S` for search. What do you think?

radomirbosak:
Ok, I changed the short option for search from `-p` to `-S`. I think it makes more sense.

craftyguy:
Is there any chance of seeing this in a release soon? I've been using it, and it works great.

empireshades:
This seems to fail only on the python 2.7 tests. In light of it being end of life'd ([https://www.anaconda.com/end-of-life-eol-for-python-2-7-is-coming-are-you-ready/]), can we simply remove it from the testing list?

empireshades:
Bumping this as per my previous message (above):

> This seems to fail only on the python 2.7 tests. In light of it being end of life'd (https://www.anaconda.com/end-of-life-eol-for-python-2-7-is-coming-are-you-ready/), can we simply remove it from the testing list?

PR #471, opened 2017-03-07
Fix python2 error in template_explorer.py

### LINKED ISSUE #456: Latest 2.0-rc1 branch gives error on jrnl execution
Pulled down the latest from the 2.0-rc1 branch and got the following error when trying to run jrnl:

```
  File "C:\Util\Python27\lib\runpy.py", line 162, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "C:\Util\Python27\lib\runpy.py", line 72, in _run_code
    exec code in run_globals
  File "d:\Projects\jrnl-nb\jrnl\__main__.py", line 4, in <module>
    from . import cli
  File "jrnl\cli.py", line 15, in <module>
    from . import plugins
  File "jrnl\plugins\__init__.py", line 13, in <module>
    from .template_exporter import __all__ as template_exporters
  File "jrnl\plugins\template_exporter.py", line 49, in <module>
    __all__.append(__exporter_from_file(template_file))
  File "jrnl\plugins\template_exporter.py", line 42, in __exporter_from_file
    "template": template
TypeError: type() argument 1 must be string, not unicode
```

radomirbosak:
If template_explorer.py was imported in python2, it would fail because
it uses unicode_literals which would ultimately cause passing an unicode
string as the first argument to the 'type' built-in function.

This commit fixes this by converting the unicode to string before it is
passed to the 'type' function.

Fixes #456

### DIFF
--- a/jrnl/plugins/template_exporter.py
+++ b/jrnl/plugins/template_exporter.py
@@ -36,7 +36,7 @@ def __exporter_from_file(template_file):
     """Create a template class from a file"""
     name = os.path.basename(template_file).replace(".template", "")
     template = Template.from_file(template_file)
-    return type("{}Exporter".format(name.title()), (GenericTemplateExporter, ), {
+    return type(str("{}Exporter".format(name.title())), (GenericTemplateExporter, ), {
         "names": [name],
         "extension": template.extension,
         "template": template


maebert:
You're on a roll!

PR #476, opened 2017-03-10
Catch keyboard interrupt when opening journal

### LINKED ISSUE #450: User unfriendly stack traces on aborted input requests
As  I understand it, judging by the site friendly appearance, `jrnl` meant to be rather a consumers app (even though for power-users who are able to use CLI). But alas, currently its behaviour isn't that user-friendly as it could be. For example, I would handle system signals (like `SIGTERM`) to avoid stack trace outputs to the user terminal, like it happens when user cancel any interactive input request:

```sh
$ jrnl -n 10                                                                                                                                                                                                                       
Password: [^-C here] 
  File "/usr/local/bin/jrnl", line 11, in <module>
    load_entry_point('jrnl==1.9.7', 'console_scripts', 'jrnl')()
  File "/usr/local/Cellar/jrnl/1.9.7/libexec/lib/python2.7/site-packages/jrnl/cli.py", line 177, in run
    journal = Journal.Journal(journal_name, **config)
  File "/usr/local/Cellar/jrnl/1.9.7/libexec/lib/python2.7/site-packages/jrnl/Journal.py", line 39, in __init__
    self.open()
  File "/usr/local/Cellar/jrnl/1.9.7/libexec/lib/python2.7/site-packages/jrnl/Journal.py", line 105, in open
    journal = util.get_password(keychain=self.name, validator=validate_password)
  File "/usr/local/Cellar/jrnl/1.9.7/libexec/lib/python2.7/site-packages/jrnl/util.py", line 35, in get_password
    password = pwd_from_keychain or getpass()
  File "/usr/local/Cellar/jrnl/1.9.7/libexec/lib/python2.7/site-packages/jrnl/util.py", line 28, in getpass
    return gp.getpass(prompt)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/getpass.py", line 71, in unix_getpass
    passwd = _raw_input(prompt, stream, input=input)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/getpass.py", line 133, in _raw_input
    line = input.readline()
KeyboardInterrupt
```

Maybe if it's really crucial for support purposes or whatever reasons, it would be reasonable to turn it off by default and provide some config option / CLI argument / environment variable to enable debug mode.

Thanks in advance. Cheers! 👍

radomirbosak:
If the journal is encrypted, jrnl asks for a password using the
getpass.getpass function. If user presses the keyboard interrupt (e.g.
ctrl+c) during the password prompt, jrnl crashes and prints a stack
trace. This however is not very user friendly.

This commit adds a check for KeyboardInterrupt Exception and prints a
message whenever this exception occurs and exits. Stack is no longer
printed.

Fixes #450

### DIFF
--- a/jrnl/cli.py
+++ b/jrnl/cli.py
@@ -206,7 +206,11 @@ def run(manual_args=None):
             mode_compose = False
 
     # This is where we finally open the journal!
-    journal = Journal.open_journal(journal_name, config)
+    try:
+        journal = Journal.open_journal(journal_name, config)
+    except KeyboardInterrupt:
+        util.prompt("[Interrupted while opening journal]".format(journal_name))
+        sys.exit(1)
 
     # Import mode
     if mode_import:


maebert:
Good catch, thanks!

PR #477, opened 2017-03-10
Remove --use-mirrors switch from travis setup

radomirbosak:
The --use-mirrors pip switch [does not work](https://pip.pypa.io/en/stable/news/) since pip 7.0.0. This caused
travis builds to fail.

This commit removes the --use-mirrors switch.


maebert:
You're [insert deity here]-send.

PR #499, opened 2017-08-30
Adds JSON importing feature and HTML exporting feature

timetoplatypus:
I realized after exporting my journal to a JSON file and accidentally deleting the original journal, there wasn't an option in the jrnl CLI to import my JSON file back into a journal except by doing so manually. This pull request adds an option in the jrnl CLI to import from a JSON formatted file.

Additionally, I wanted a way to export my journal to a static webpage with minimal/no client-side dependencies that I know will look "ok"/"pretty". So this pull request also adds another option to the jrnl CLI that gives the user an option to export their jrnl to a light HTML file with inlined CSS; the intent is that the user can export their journal and drop it straight into a static content server (i.e. Apache webserver) without any extra work to make it look presentable

I did not make any changes related to the jrnl application version, Travis, etc. etc. as I wanted to leave that to the upstream maintainer (@maebert) to decide whether these changes are appropriate to merge into the jrnl application. If they are, I'd like to see if these additions can be pushed out to the jrnl being served through the pip repositories too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,9 @@ If you just call `jrnl`, you will be prompted to compose your entry - but you ca
 
 Installation
 ------------
+Install _jrnl_ from source:
+
+    sudo python setup.py install
 
 Install _jrnl_ using pip:
 
--- a/jrnl/Entry.py
+++ b/jrnl/Entry.py
@@ -109,3 +109,31 @@ class Entry:
             body=body,
             space=space
         )
+
+    def to_html(self):
+        html = "<!DOCTYPE html>\n"
+        html += "<html>\n"
+        html += "\t<head>\n"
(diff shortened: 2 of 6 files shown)


maebert:
Hi, thanks for this PR! this is definitely a great idea. I'll admit right away that I haven't got a lot of time to maintain jrnl recently, but in general all new contributions should go against the [2.0-rc1](https://github.com/maebert/jrnl/tree/2.0-rc1) branch. Could you rebase that against that?

timetoplatypus:
No problem, I'll try to do that (hopefully this coming weekend). I haven't worked with Travis before but I think I know what caused that failure so I'll try and fix that too

PR #501, opened 2017-09-16
PyCrypto is dead. Switch to PyCryptodome

jnozsc:


jnozsc:
@maebert

sigmap:
I hope this gets merged. It'd be a good temporary fix until 2.0

tcyrus:
I am currently unable to install `jrnl[encrypted]` due to dlitz/pycrypto#218, which will not be fixed because the project is no longer maintained.

tcyrus:
If anybody is trying to install on Windows with pycrypto, this solution worked for me:
https://github.com/dlitz/pycrypto/issues/218#issuecomment-434551388

tcyrus:
@maebert

crookshanked:
Too bad about this not getting merged in. The above info seems to not fully work for me with the work around. It installs pycrypto, and then jrnl installs... creates the journal and encrypts... then asks for the password and errors saying pycrypto is not installed.

wren:
Hi, we switched away from PyCrypto in v2.0. Sorry that this didn't get merged in! Thank you for the PR anyway!

PR #506, opened 2017-11-14
Switch to hashmark Markdown headers on export

### LINKED ISSUE #487: Markdown headers first and second level as # and ##
Markdown export is a handy feature. Could we get an option to export first and second level headers with the more often used # and ## instead of underlines?

MinchinWeb:
Simple fix that I hope is helpful!

Closes #487


wren:
@MinchinWeb Looks like this is failing some tests. Can you please look into it?

MinchinWeb:
Will do!

PR #516, opened 2017-12-06
Fix #512

### LINKED ISSUE #512: jrnl writes to default journal if a colon is used after a journal's name
You can't write to non-default journals without specifying a date, and when you do, it writes to the default journal instead. For example, if I have journals "default" and "projects", I would see the following behavior:

C:\Users\Me>jrnl projects now: This will work properly...
[Entry added to projects journal]

C:\Users\Me>jrnl projects: This won't, and will write to default instead.

juanpagfe:
Fixes non default entry creation bug when date is not given. Fix #512

### DIFF
--- a/jrnl/cli.py
+++ b/jrnl/cli.py
@@ -153,7 +153,7 @@ def run(manual_args=None):
 
     # If the first textual argument points to a journal file,
     # use this!
-    first_arg = args.text[0].replace(':', '')
+    first_arg = args.text[0].replace(':', '') if len(args.text) > 0 else None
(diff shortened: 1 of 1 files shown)


juanpagfe:
I don't know what is wrong with the travis check. Apparently is nothing about my code, but I am not sure.  Can anyone help me to figure it out?

juanpagfe:
By the way, it works in my computer.

wren:
Hi, if you want to repopen this, we'd like to take a look at it.

PR #526, opened 2018-02-16
Create CODE_OF_CONDUCT.md

maebert:


micahellison:
We just noticed this! We implemented the same text from the GitHub boilerplate also. Great minds think alike. :)

PR #527, opened 2018-02-19
Update advanced.rst

saltycraig:
Update to clarify that jrnl creates .jrnl_config file under $XDG_CONFIG/jrnl/ folder (I mistakenly interpreted the original to mean the config file should be called $XDG_CONFIG/jrnl, and was super confused that jrnl did not read it).


maebert:
Thanks, good call!

PR #543, opened 2018-06-24
added config option for locale to make date output work nicer

domfi:
Should solve issue: https://github.com/maebert/jrnl/issues/521


domfi:
Sorry. Does not work as intended, further work needed...

Issue #502, opened 2017-09-21
Multiple concurrent jrnl invocations lose entries


semi:
```
$ jrnl -1
2017-09-21 15:00 my last entry

$ (echo first; sleep 1) | jrnl & echo second | jrnl
[Entry added to default journal]
[Entry added to default journal]
$ jrnl -2
2017-09-21 15:00 my last entry

2017-09-21 15:00 first
```

What happened to entry 'second'?

I just lost two day's worth of journal because I went to save and quit an entry that I had forgotten open in an editor.

Other than that, jrnl has been a real gem, thank you so much for developing it!

sorsasampo:
I have been bitten by this as well. This is quite serious issue due to data loss.

sorsasampo:
Keep this open. Data loss is bad.

wren:
@sorsasampo Hi! The stalebot got a bit overzealous with the backlog. This issue should already be fixed in v2.0. 

Can you try the v2.0-rc2 version and verify if you still experience this issue? You can install with `pip install jrnl==2.0.0rc2`. 

Thanks!

sorsasampo:
@wren, this particular issue seems to be fixed with 2.0.0rc3.post2.

BenKettlewell:
I upgraded to the latest version but this isn't working as you'd expect.

I edited two entries from the same journal and made a couple changes. Saved and quit both and some, but not all the changes were retained.

PR #545, opened 2018-06-26
Avoid data loss by postponing open() (#502)

semi:
Postpone the reading of existing entries until a new entry
has been obtained and is ready to be appended to the journal.

This patch reduces the likelihood of losing entries when the
user has launched concurrent jrnl sessions to add multiple
entries. A proper solution would require locking the journal
file for the time between reading and writing, but this is
already good enough for typical interactive use cases.

### DIFF
--- a/jrnl/Journal.py
+++ b/jrnl/Journal.py
@@ -36,8 +36,6 @@ class Journal(object):
         self.search_tags = None  # Store tags we're highlighting
         self.name = name
 
-        self.open()
-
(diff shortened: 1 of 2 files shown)


semi:
I notice that the bug has been fixed in 2.0 branches, but here is a quick patch that I wrote to fix it in 1.9.8 for my personal use.

maebert:
Great, thanks!

PR #550, opened 2018-07-18
Handle KeyboardInterrupt when installing journal

silenc3r:
Don't throw exception on Ctrl-C, simply abort.

### DIFF
--- a/jrnl/cli.py
+++ b/jrnl/cli.py
@@ -13,7 +13,7 @@ from . import Journal
 from . import util
 from . import install
 from . import plugins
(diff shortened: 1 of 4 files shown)


wren:

This is a good bit of polish to add to the project. Would you mind rebasing so we can merge?

wren:
I think this is a good bit of polish that we should get merge, if possible.

silenc3r:
I've moved handling of exceptions to the outermost function.

wren:
@silenc3r Thank you!

PR #555, opened 2018-08-08
Allow argument list for config['editor']

hoclun-rigsep:
This helps whenever your editor commandline needs to escape spaces,
for example:

        vim -c set\ filetype=markdown

        can now go in your config file as:

            { "editor": ["vim", "-c", "set filetype=markdown"] }

### DIFF
--- a/jrnl/util.py
+++ b/jrnl/util.py
@@ -130,7 +130,10 @@ def get_text_from_editor(config, template=""):
     with codecs.open(tmpfile, 'w', "utf-8") as f:
         if template:
             f.write(template)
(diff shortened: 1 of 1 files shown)


maebert:
Nice! Could you rebase that against the 2.0 branch?

hoclun-rigsep:
It doesn't rebase clean I think. I don't know when I'll get to it but I'll try.

hoclun-rigsep:
It doesn't rebase clean I think. I don't know when I'll get to it but I'll try.

Issue #415, opened 2016-03-28
Tempfiles on Windows


maebert:
Via Email:

> I don't know if this is the right way to go about this, but I'm having a little trouble getting jrnl configured on my Windows 8 machine.  I set vim as the editor in the configuration file, but when I run jrnl, it opens an instance of vim in read only mode.  Looking at the traceback, it tries to run `os.remove(tmpfile)` but prompts a windows error saying "The process cannot access the file because it is being used by another process: `c:\users\username\appdata\local\temp\jrnldokd2b.txt`"  I'm still a noob at trying to solve issues like this on my own, but it looks like `mkstemp` returns an open file which then causes problems when subprocess.call and os.remove try and mess with file.  Is this right or am I barking up the wrong tree?  And how would you solve the issue?  Should I try explicitly closing the file after creating it and getting the path?

maebert:
@MinchinWeb You're on a windows box, right? Any idea what might cause this?

MinchinWeb:
Indeed I am.

I ran into a similar (the same) issue when I was trying to set up Sublime Text as my editor. What is happening is that jrnl console script [1] is creating a python process [2], which in turn is creating the editor process [3], whether that's VIM, Notepad++, or Sublime Text 3.

The python process [2] maintains a handle on the temp file after it calls the editor. Some editors (like Notepad++) seem to ignore it, and will allow you to save your edits to the temp file. Other editors (like VIM and Sublime Text) seem to notice the other file handle and refuse to save changes to the file due to it being open by another process.

Possible solutions:

1. Use Notepad++ as your editor
2. have `jrnl` drop the file handle (but not delete the file) before calling the editor, and reopen the file handle after the editor has closed.

MinchinWeb:
PR#416 implements option 2 by closing the filehandle before opening the editor. VIM and Sublime Text 3 now work on my end as editors.

lamarqua:
Bump.

```
E:\Home>jrnl
Traceback (most recent call last):
  File "C:\Software\Python\Python35-32\Scripts\jrnl-script.py", line 9, in <module>
    load_entry_point('jrnl==1.9.8', 'console_scripts', 'jrnl')()
  File "c:\software\python\python35-32\lib\site-packages\jrnl\cli.py", line 207, in run
    raw = util.get_text_from_editor(config)
  File "c:\software\python\python35-32\lib\site-packages\jrnl\util.py", line 136, in get_text_from_editor
    os.remove(tmpfile)
[...]

PR #564, opened 2018-11-14
Added @hollystyles' fix for Issue #415

Austin-Scott:
Fixes the problem with permissions when trying to use an external editor to edit old entries on Windows.


maebert:
Nice one, thanks!

Issue #594, opened 2019-05-18
Failed export test - name 'len' is not defined


micahellison:
Running the tests with behave on the 2.0-rc1 branch and found this test failing:
```
  Scenario: Exporting using custom templates  # features/exporting.feature:30
    Given we use the config "basic.yaml"      # features/steps/core.py:66
    Given we load template "sample.template"  # features/steps/core.py:100
    When we run "jrnl --export sample"        # features/steps/core.py:90
```

You can easily reproduce this by running:
```
jrnl --export sample
```
Which leads the output:
```
NameError
   "-" * len(entry.title)
          ^^^
name 'len' is not defined
```

It looks like it has something to do with the TemplateExporter class and code in template.py that is consuming code from /templates/sample.template - a line of hyphens that's supposed to be the same length as the title.

wren:
I can confirm on my system that this test was failing before, and this PR fixes it.

PR #595, opened 2019-05-18
#594 allowing template interpreter to interpret common Python commands

micahellison:
It seems like the intent in sample.template is for all templates to be able to arbitrarily execute Python commands, so I initialized the asteval Interpreter with its default commands -- referenced here: https://newville.github.io/asteval/basics.html

Alternatively, if templates should only be able to call len, we could add that to the symtable.

### DIFF
--- a/jrnl/plugins/template.py
+++ b/jrnl/plugins/template.py
@@ -39,7 +39,8 @@ class Template(object):
         return self._expand(self.blocks[block], **vars)
 
     def _eval_context(self, vars):
-        e = asteval.Interpreter(symtable=vars, use_numpy=False, writer=None)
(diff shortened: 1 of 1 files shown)


micahellison:
So.... I just realized that I wasn't working with the latest code and saw the "export changes" commit from 4/29/19 - looks like that breaks this but also lots of other stuff. Not sure where to go from here.

wren:
I think this is the commit @micahellison is referencing: https://github.com/maebert/jrnl/commit/81dfebb2c0ef4d4ebf16e8c29fdec5f4a9d9afb3

I'm having a similar problem. This PR fixed what was at the time the only failing test (running `behave` locally), but the commit above broke quite a few tests (~25) for me.

PR #611, opened 2019-07-08
Move to mkdocs

maebert:
This PR moves the documentation to [MkDocs](https://www.mkdocs.org). This has several advantages:

- Significantly easier and faster build process than sphinx
- Docs are written in Markdown instead of ReST, which lowers the barrier to entry for new contributors
- Direct deployment to github-pages

### Development

```sh
pip install mkdocs
make html # opens the docs and live-reloads
make docs # builds docs and deploys them to gh-pages
```

### This PR also

- [x] Consolidates doc edits from master, rc1, and rc2
- [x] Simplifies and modernizes the landing page
    landing
- [x] Modernizes the assets
- [x] Includes wonderful social banners for sharing on Twitter, Facebook, etc:
    image
- [x] Switches to a modified ReadTheDocs theme that fixes some readability issues (turns out the web has changed quite a bit since 2012...):
    docs

PR #612, opened 2019-07-08
Use Poetry for dependency management and deployments

### LINKED ISSUE #610: cryptography is pinned to version 1.4
I was unable to build and install cryptography==1.4 using libssl-dev 1.1.1 to compile it. I was able to work around this by relaxing the version requirement by specifying cryptography>=1.4, which is using cryptography==2.7.

Is there a reason to use an older version of cryptography? I was able to read and write encrypted journals with 2.7 without issues (so far).

maebert:
Since `jrnl` was first written in 2012, a lot has happened in the wonderful world of dependency management and python deployments. This PR moves the deployment process to [Poetry](https://poetry.eustace.io), which has the following advantages over `pip` or `pipenv`

- Requirements can be locked (similar Ruby gems or Yarn), creating less dependency conflicts for users and developers
- Poetry handles virtual environments, no need to create and maintain one yourself
- It's super easy to publish from Travis straight to PyPi (set to deploy from master on tagged commits)

With poetry, the new development workflow is

```sh
# 1. Set up environment
git clone ...
cd jrnl
poetry install

# 2. Add new dependencies
poetry add my-dependency

# 3. Run tests (or any other command that requires the installed dependencies)
poetry run behave
```

I'll update the documentation for this after #611 is merged.

Also fixes #610.


wren:
This is a good update, but it changes a few dependencies. I think that introduces risk to v2.0 since we already have people testing, and I don't want to lose that momentum. I'd love to get this merged in first thing for v2.1, though.

maebert:
Yup, agree, let's get 2.0 in first and then rebase this against master.

PR #621, opened 2019-07-16
Unpin crypto #610

wren:
This is the smallest fix possible for #610. I want to avoid changing the dependency management until we have time to vet all of the updated dependencies.


gregorybodnar:
That was how I approached the fix when I reported it and I've not yet had any crypto issues.

PR #626, opened 2019-07-22
Reading

### LINKED ISSUE #344: Clarify usage output between export and reading sections
When reading the usage output, I didn't expect to be able to use --short while reading a journal, but it worked.  This makes sense knowing the code, but just from reading the usage it is unexpected.

I think '--short' and '--tags' should be listed under 'Reading' with a note under 'Export' that 'Reading''s options can also be used.

<pre>
usage: jrnl [-h] [-v] [-ls] [-from DATE] [-until DATE] [-on DATE] [-and]
            [-starred] [-n [N]] [--short] [--tags] [--export TYPE] [-o OUTPUT]
            [--encrypt [FILENAME]] [--decrypt [FILENAME]] [--edit]
            [[...]]

optional arguments:
  -h, --help            show this help message and exit
  -v, --version         prints version information and exits
  -ls                   displays accessible journals

Composing:
  To write an entry simply write it on the command line, e.g. "jrnl
  yesterday at 1pm: Went to the gym."

  

Reading:
  Specifying either of these parameters will display posts of your journal

  -from DATE            View entries after this date
  -until DATE, -to DATE
                        View entries before this date
  -on DATE              View entries on this date
  -and                  Filter by tags using AND (default: OR)
  -starred              Show only starred entries
  -n [N]                Shows the last n entries matching the filter. '-n 3'
                        and '-3' have the same effect.

Export / Import:
  Options for transmogrifying your journal

  --short               Show only titles or line containing the search tags
  --tags                Returns a list of all tags and number of occurences
  --export TYPE         Export your journal. TYPE can be json, markdown, or
                        text.
  -o OUTPUT             Optionally specifies output file when using --export.
                        If OUTPUT is a directory, exports each entry into an
                        individual file instead.
  --encrypt [FILENAME]  Encrypts your existing journal with a new password
  --decrypt [FILENAME]  Decrypts your journal and stores it in plain text
  --edit                Opens your editor to edit the selected entries.
</pre>

mzfr:
Fixes #344

### DIFF
--- a/jrnl/cli.py
+++ b/jrnl/cli.py
@@ -38,11 +38,11 @@ def parse_args(args=None):
     reading.add_argument('-on', dest='on_date', metavar="DATE", help='View entries on this date')
     reading.add_argument('-and', dest='strict', action="store_true", help='Filter by tags using AND (default: OR)')
     reading.add_argument('-starred', dest='starred', action="store_true", help='Show only starred entries')
+    reading.add_argument('-s', '--short', dest='short', action="store_true", help='Show only titles or line containing the search tags')
+    reading.add_argument('--tags', dest='tags', action="store_true", help='Returns a list of all tags and number of occurences')
     reading.add_argument('-n', dest='limit', default=None, metavar="N", help="Shows the last n entries matching the filter. '-n 3' and '-3' have the same effect.", nargs="?", type=int)
 
     exporting = parser.add_argument_group('Export / Import', 'Options for transmogrifying your journal')
-    exporting.add_argument('-s', '--short', dest='short', action="store_true", help='Show only titles or line containing the search tags')
-    exporting.add_argument('--tags', dest='tags', action="store_true", help='Returns a list of all tags and number of occurences')
     exporting.add_argument('--export', metavar='TYPE', dest='export', choices=plugins.EXPORT_FORMATS, help='Export your journal. TYPE can be {}.'.format(plugins.util.oxford_list(plugins.EXPORT_FORMATS)), default=False, const=None)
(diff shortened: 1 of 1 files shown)


wren:
It looks like some tests are failing. Can you be sure those are updated so this is ready for review?

mzfr:
I am having a bit difficult time understanding why the test are failing.

jrnl/cli.py
```diff
@@ -63,7 +63,7 @@ def guess_mode(args, config):
         compose = False
         export = False
         import_ = True
-    elif args.decrypt is not False or args.encrypt is not False or args.export is not False or any((args.short, args.tags, args.edit)):
+    elif args.decrypt is not False or args.encrypt is not False or args.export is not False or args.edit:
```
wren: Did you mean to commit this change? This is a change to guess mode, and the original issue was documentation-only.

mzfr: yes @maebert 	pointed this out in [comment](https://github.com/jrnl-org/jrnl/pull/617#pullrequestreview-260446315) of my old PR. That is why I made these changes.

wren: Sorry about the back and forth, but I'm still not sure about this line. It looks like it breaks the functionality. Do the tests pass if you take this one out?

peter-schmidbauer: The "export mode" is a bit misleading. Short or tags mode only get triggered when "export mode" is True, otherwise it will always default to the normal reading mode. So either you need to revert this change or fix the if-else at the bottom of cli.run() here: https://github.com/jrnl-org/jrnl/blob/master/jrnl/cli.py#L248-L260.

The whole cli.py could use a severe revamp/code refactoring, it's quite a mess.

wren:
Closing to clear up the review queue. Feel free to resubmit once you've addressed the comments.

PR #375, opened 2015-08-19
Add functionality for creating new journals from cmd

alshedivat:
Improved PR #331 rebased against `2.0-rc1`.

PR #627, opened 2019-07-22
Add support to create new journal from cmd

### LINKED ISSUE #330: Create new journals from the command line

mzfr:
Fixes #330
Also, we should be able to close #375

### DIFF
--- a/jrnl/cli.py
+++ b/jrnl/cli.py
@@ -15,6 +15,7 @@ from . import install
 from . import plugins
 from .util import ERROR_COLOR, RESET_COLOR, UserAbort
 import jrnl
+from os.path import abspath, join, exists
 import argparse
 import sys
 import logging
@@ -28,6 +29,7 @@ def parse_args(args=None):
     parser.add_argument('-v', '--version', dest='version', action="store_true", help="prints version information and exits")
     parser.add_argument('-ls', dest='ls', action="store_true", help="displays accessible journals")
     parser.add_argument('-d', '--debug', dest='debug', action='store_true', help='execute in debug mode')
+    parser.add_argument('-create', help='create a new journal', metavar='NAME')
 
     composing = parser.add_argument_group('Composing', 'To write an entry simply write it on the command line, e.g. "jrnl yesterday at 1pm: Went to the gym."')
     composing.add_argument('text', metavar='', nargs="*")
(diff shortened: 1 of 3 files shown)


wren:
It looks like some tests are failing. Can you be sure those are updated so this is ready for review?

MinchinWeb:
keep this in the pipeline?

mzfr:
I can complete this if anyone could help me understand why the test cases are failing. I am having a bit hard time understanding that.

wren:
@mzfr I can take look next Saturday, if it hasn't been resolved by then.

jrnl/cli.py
```diff
@@ -151,14 +153,34 @@ def run(manual_args=None):
...
+    # If the current folder contains .jrnl file, read the journal name from it
         util.prompt("No default journal configured.")
+    elif exists('.jrnl'):
+        with open('.jrnl', 'r') as fp:
+            journal_name = fp.read().strip()
+    else:
+        journal_name = 'default'
+
+    if journal_name not in config['journals']:
+        util.prompt("No '{}' journal configured.".format(journal_name))
```
wren: The addition of quotes around default is what's making the current test fail. Removing those, or updating the test will make it pass. 

There is a problem in that there aren't any tests for this new functionality, though. Is there any chance you can add some tests?

mzfr:
@wren I can work on adding new test cases but I'll need some help in understanding how to add them. Also, it would be nice if you can provide some explanation on what kind of test to be added.

alichtman:
> I can work on adding new test cases but I'll need some help in understanding how to add them.

Read the [tutorial / docs for behave](https://behave.readthedocs.io/en/stable/tutorial.html).

> it would be nice if you can provide some explanation on what kind of test to be added.

Test the creation of a journal (which should succeed) and the creation of a journal that already exists (which should fail).

mzfr:
@wren @alichtman so I went through the tutorial and I was able to write the test for `when journal already exists`

```
Scenario: create a new journal that already exists
        Given we use the config "basic.yaml"
        When we run "jrnl -create default"
        Then we should see the message "Journal with `test-random` name already exists."
```

But I am confused with the case when the journal doesn't exist because when we would run `jrnl -create random` it would ask `Path to your journal file (leave blank for /home/mzfr/jrnl/journal.txt):` and then it would prompt for `Enter password for journal (leave blank for no encryption):`. So I am confused about how to test for those prompts.


Also another question, I have to make a new file named `creating-jrnl.feature`to add these tests? or can I add it in `core.feature`

peter-schmidbauer:
@mzfr I'd rename the Scenario you wrote to "Creating a journal that already exists fails" or something like that. I would also use a double-dash (`--create`) for the command to bring it in line with --edit, --encrypt and so on (imho -ls should also be changed to --ls, but that's a different question).

As to your questions: You can test user input with the behave line `When we run "jrnl --create random" and enter """your commands here"""` with each command on a new line. See `upgrade.feature` for some examples. 

I'd append the tests to `core.feature` for now.

wren:
Closing to clear up the review queue. Feel free to resubmit once you've addressed the comments.

PR #628, opened 2019-07-22
Fzf recipe

gregorybodnar:
Add a recipe that uses [fzf](https://github.com/junegunn/fzf) as a real-time filter to select an entry from `jrnl --short` and display the full entry.

In this trivial example, an error will be returned if fzf is aborted with no selection, since `jrnl -on` expects an argument. A more complete shell function can guard against that, which I do in my fish shell version. I'm tempted to leave this as simple and flawed because it works well enough and can be typed by hand.

### DIFF
--- a/docs/recipes.rst
+++ b/docs/recipes.rst
@@ -81,7 +81,7 @@ Select a previous entry to display
 
 .. code-block:: sh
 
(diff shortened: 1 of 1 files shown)


gregorybodnar:
Closing this PR and opening a new one that's based on the new Markdown docs.

Issue #367, opened 2015-06-27
Random entry within other read requirements


momandine:
I added a tiny feature to return a random file entry, within the date requirements. I returns 1 entry if not specified, or n entries if the -n flag is used.

You can look at my fork if you're interested: https://github.com/momandine/jrnl

Thanks!

maebert:
Interesting! What's your use case for that?

momandine:
I added it to my .bash_profile so I can get a random snippet of my past
every time I open my terminal. jrnl became my replacement for OhLife, which
shut down last fall. I might set up an email system too.

In any case I find happening across old entries I'd forgotten about really
interesting.

mzfr:
@maebert 	Do we need this in our project. If yes then I can work on it

maebert:
I'd be more interested in a clever shell script that provides that functionality. jrnl is meant to be composable!

mzfr:
So you mean a simple bash script that just parse the .jrnl file and randomly shows you the entry ?

gregorybodnar:
I don't think you'd need to directly parse the .jrnl file. I use fzf to filter the list provided by `--short` and grab the timestamp. That gets passed to another jrnl invocation to display the full entry. Adding a random selection in place of fzf would be simple.

maebert:
That would be a great little script to add to the [Recipes](https://jrnl.sh/recipes.html#id1)!

insanerwayner:
I was just thinking about this the other day. I was thinking about writing a bash script to get a random entry. I was thinking I would want the following features:

- Random entry by default
- Random entry within a date range or other filters (tags, etc)
- FlashBack: Basically like timehop, or "This Day in History", to show entries from this date from years past.

I may still write the bash script, but it'd be awesome if these features were written and encorporated into the program by much better programmers than myself. I'm a scripter(sysamin) at best.

PR #629, opened 2019-07-22
Add recipe to display random entry

gregorybodnar:
This is based on the request in #367. Instead of modifying jrnl directly, it composes shell tools to get a similar result.


wren:
@gregorybodnar Thank you for the PR!

Can you be sure to pull the latest from master? We recently changed the docs to be markdown files, so the edits you made to the old *.rst files are causing conflicts.

gregorybodnar:
I think the easiest thing to do is to close this PR and open a new one. Trying to rebase against files that have been removed is going to get ugly quickly.

PR #633, opened 2019-07-24
Add recipe to display random entry

gregorybodnar:
New PR to address the request in #367 , but based on new Markdown docs.


wren:
After some testing, it looks like this works only with certain time formats. I think that's fine as long as we specify what that format is. Can you update to include an example in the recipe?

PR #634, opened 2019-07-24
Add recipe to select an entry with fzf

gregorybodnar:
Add a recipe that uses FZF to select an entry for display.

### DIFF
--- a/docs/recipes.md
+++ b/docs/recipes.md
@@ -93,6 +93,15 @@ log_question 'What did I achieve today?'
 log_question 'What did I make progress with?'
 ```
 
+### Filtering with FZF
+
+Use the search powers of [fzf](https://github.com/junegunn/fzf) to select an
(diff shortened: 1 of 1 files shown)


wren:
After some testing, it looks like this works only with certain time formats. I think that's fine as long as we specify what that format is. Can you update to include in the recipe?

gregorybodnar:
Thanks for checking on that. The pipeline needs to be customized for timestamp rendering, which is usually pretty straightforward, given the pattern used by jrnl. Having said that, `jrnl -on` needs to be able to parse the timestamp rendered, no matter how broken `timeformat` is.

wren:
Accepting this. It's not quite what I was talking about, but we can clarify it a bit later. I'd like an example of the format that a user would have in their config in order for this recipe to work, so that they can just copy and paste to use the recipe.

Issue #632, opened 2019-07-24
Add live post-upgrade tests


micahellison:
We've seen a handful of issues from testers (thank you!) when upgrading journals from 1.x.x to 2.0. I think we can handle all the issues that have been reported, but I'm still worried about but issues we can't anticipate. Working with completely arbitrary and unescaped plain text that has to be divided into records isn't easy, after all.

One way to deal with this would be to add some post-upgrade tests, and warn the user (or just print an error and crash!) if any of the tests fail.

This is the main test that comes to mind for me: If we dump the imported entries to text, then parse that text as if it's the new journal, do we end up with the same number of entries?

Feel free to comment with other ideas.

MinchinWeb:
I think a 'count entries' test is a good basic test.

Do we make a backup of the jrnl file before upgrading?

wren:
I agree that count is mostly good enough for this migration. Maybe we can also check that the dates line up, in order to catch if any of the dates were changed or anything like that.

For the backup file, I think warning the user to backup their journal before upgrading should be good for this. I don't want to take the burden of backing up a user's journal for them without adding way more checks and balances.

micahellison:
So, it turns out that jrnl DOES make a backup when upgrading! It saves the old journal in the same file path with a .backup extension. It might be worthwhile for the console to let the user know when this happens, though.

As for tests, I agree that it should check the dates. That would rule out situations where the upgrade process would mangle the old dates.

micahellison:
With help from @wren I made a pull request (#650) for this and was able to test that every entry is exactly the same after upgrading (title, body, date, etc.). It will abort the upgrade if there are any differences.

PR #650, opened 2019-08-24
[GH-632] confirming that each journal can be parsed during upgrade

micahellison:
And aborting upgrade if not.

This change also moves upgrade-related tests into their own upgrade.feature file.

### DIFF
--- a/jrnl/install.py
+++ b/jrnl/install.py
@@ -86,7 +86,7 @@ def load_or_install_jrnl():
     if os.path.exists(config_path):
         log.debug('Reading configuration from file %s', config_path)
         config = util.load_config(config_path)
-        
+
         try:
             upgrade.upgrade_jrnl_if_necessary(config_path)
         except upgrade.UpgradeValidationException:
@@ -94,7 +94,7 @@ def load_or_install_jrnl():
             sys.exit(1)
 
         upgrade_config(config)
-        
+
         return config
     else:
         log.debug('Configuration file not found, installing jrnl...')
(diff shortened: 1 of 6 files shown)


jrnl/upgrade.py
```diff
@@ -85,17 +86,27 @@ def upgrade_jrnl_if_necessary(config_path):
...
+
+    # loop through lists to validate
+    failed_journals = [j for j in all_journals if not j.validate_parsing()]
+    if len(failed_journals) > 0:
+        util.prompt("\nThe following journal{} failed to upgrade:\n{}".format(
+            's' if len(failed_journals) > 1 else '', "\n".join(j.name for j in failed_journals))
+        )
+
+        util.prompt("Aborting upgrade.")
+        return
```
wren: We should use an exception here instead of returning. Once this function returns, a bunch of other code will still run (like upgrading the config), and not fully abort the upgrade.

micahellison: Agreed - just made the change.

jrnl/install.py
```diff
@@ -90,7 +90,10 @@ def load_or_install_jrnl():
         try:
             upgrade.upgrade_jrnl_if_necessary(config_path)
         except upgrade.UpgradeValidationException:
-            util.prompt("Aborting upgrade. Exiting.")
+            util.prompt("Aborting upgrade.")
+            util.prompt("Please tell us about this problem at the following URL:")
+            util.prompt("https://github.com/jrnl-org/jrnl/issues/new?title=UpgradeValidationException")
```
wren: 🚀

PR #692, opened 2019-10-25
Pretty print journal entries

### LINKED ISSUE #508: Pretty print journal entry titles as bold
There is already a visual distinction created by the `|` at the beginning of the journal entry lines, but when you want to quickly skim through the entries looking at titles, it's nice to have a stronger distinction in my opinion. Making the titles bold accomplishes this nicely. I already put together the PR since it's literally just a few lines, but the contributing guidelines say to make an issue first.

alichtman:
Changes appearance of all jrnl viewing commands, such as `$ jrnl --short` and
`$ jrnl -n {NUM}`.

Fix #508, #717


alichtman:
This feature has some slight bugs that I'm still working out.

alichtman:
The last failing test case is exporting with custom templates. This is also failing for me on master, so I don't think it's my PR causing this.

```
  Scenario: Exporting using custom templates  # features/exporting.feature:23
    Given we use the config "basic.yaml"      # features/steps/core.py:68 0.003s
    Given we load template "sample.template"  # features/steps/core.py:102 0.001s
    When we run "jrnl --export sample"        # features/steps/core.py:92
NotImplementedError
   "-" * len(entry.title)
'Constant' not supported
NotImplementedError
   "-" * len(entry.title)
    When we run "jrnl --export sample"        # features/steps/core.py:92 0.022s
    Then the output should be                 # features/steps/core.py:171 0.000s
      """
      My first entry.
      ---------------

      Everything is alright

      Life is good.
      -------------

      But I'm better.
      """
      Assertion Failed: ['---------------', 'None']
      Captured stdout:
      My first entry.
      None

      Everything is alright

      Life is good.
      None

      But I'm better.

```

peter-schmidbauer:
I'm not the biggest fan of the color choices to be honest. Why not let the user define the colors (separately for date, title and body) in the config?

alichtman:
> Why not let the user define the colors (separately for date, title and body) in the config?

Good call. Do you think it's better to have them set colors like `red, blue, green, ...` or ANSI escapes?

peter-schmidbauer:
Definitely colors. Not sure what the ideal way to map them to escapes is though. We should probably look into moving away from using the raw escapes all together and towards a package like [colorama](https://pypi.org/project/colorama/).

features/steps/core.py
```diff
@@ -5,8 +5,10 @@
 from jrnl import cli, install, Journal, util, plugins
 from jrnl import __version__
 from dateutil import parser as date_parser
+from ansiwrap import strip_color
 from collections import defaultdict
 import os
+import re
```
alichtman: This should be removed.

alichtman:
I should add two tests:

1. Identifying invalid color configs
2. Upgrading the config from no colors -> colors

jrnl/util.py
```diff
@@ -162,6 +162,21 @@ def scope_config(config, journal_name):
...
 
+def verify_config(config):
+    """
+    Ensures the keys set for colors are valid colorama.Fore attributes.
+    :return: True if all keys are set correctly, False otherwise
+    """
+    all_valid_colors = True
+    for key, color in config["colors"].items():
+        if not getattr(colorama.Fore, color.upper(), None):
+            # TODO: Not sure whether both of these should stay or not.
```
alichtman: This should be addressed before merging.

alichtman:
@wren I believe this is ready for review.

wren:
@alichtman Great! I'll review this weekend.

One thing from first glance, though: Is there an option to disable color? I know a lot of people pipe output to various scripts, and the escape sequences can complicate that.

alichtman:
Yep, I can add that.

jrnl/util.py
peter-schmidbauer: Probably just the print(), but use print(..., file=sys.stderr) to keep it in line with the other jrnl output.

jrnl/util.py
```diff
@@ -180,9 +198,20 @@ def get_text_from_editor(config, template=""):
...
+def colorize(string, color, bold=False):
+    """Returns the string colored with colorama.Fore.color. If the color set by
+    the user is "NONE", it returns the string without any modification. Otherwise,
+    If the color set by the user doesn't exist in the colorama.Fore attributes,
+    the colorization is done with WHITE."""
+    upper_color = color.upper()
+    if upper_color == "NONE":
+        return string
+
+    color_escape = getattr(colorama.Fore, upper_color, colorama.Fore.WHITE)
```
peter-schmidbauer: I think the default should just be no color at all. This way we keep backwards compatibility and we don't confuse users who have set their terminal background color to white.

jrnl/install.py
```diff
@@ -57,6 +58,10 @@ def module_exists(module_name):
     'highlight': True,
     'linewrap': 79,
     'indent_character': '|',
+    'colors': {
+        'date': 'red',
+        'title': 'blue',
```
peter-schmidbauer: I'd add one for the body and for tags too, because why not? ;)

jrnl/util.py
alichtman: White is the default for an invalid color, but I can change the color defaults to none when the config is created.

peter-schmidbauer: Ah, I just realized somebody with an old config would just get the new default values as well.
Yea that would probably be better.

alichtman: Done.

alichtman:
I've found one bug in the config updating algorithm that will need to be addressed. It only looks at top-level keys, and not nested keys.

https://stackoverflow.com/a/27266178

A solution like this will be needed if the subkeys of `colors` will ever be updated...

I'll open a separate issue.

alichtman:
image

So here's an interesting bug...

I understand why the body drops all colorama formatting after the tag (due to the `colorama.Style.RESET_ALL` in `colorize()`), but I can't figure out why the title doesn't have the same problem.

peter-schmidbauer:
By the way, I think stripping the color from all outputs during tests is not ideal. I'd instead have all previous tests use no color, and add a new configuration with colors and some more specific tests. It's a new feature and deserves its own tests imho.

alichtman:
I tried for a few hours to figure out how to get behave to work with ANSI escapes and made no progress. I agree with you, but I don't think it's possible.

jrnl/util.py
```diff
@@ -180,9 +196,58 @@ def get_text_from_editor(config, template=""):
...
+    color.tags, while colorizing the rest of the text based on color.
+    :param entry: Entry object, for access to journal config
+    :param text: Text to be colorized
+    :param color: Color for non-tag text, passed to colorize()
+    :param bold: Bold flag text, passed to colorize()
+    :return: Colorized str
+    """
+    config = entry.journal.config
+    if config['highlight']:  # highlight tags
+        if entry.journal.search_tags:
```
alichtman: The code duplication here bothers me but I don't know how to clean it any more at the moment.

alichtman:
image

Looks functional to me.

jrnl/util.py
```diff
@@ -162,6 +162,22 @@ def scope_config(config, journal_name):
...
+def verify_config(config):
+    """
+    Ensures the keys set for colors are valid colorama.Fore attributes, or "None"
+    :return: True if all keys are set correctly, False otherwise
+    """
+    all_valid_colors = True
+    for key, color in config["colors"].items():
+        upper_color = color.upper()
+        if upper_color == "NONE":
+            pass
```
peter-schmidbauer: This should be "continue"

jrnl/util.py
alichtman: Figured out how to condense this, but caused a regression in the test for unreadable dates that I'm working on fixing at the moment.

jrnl/util.py
alichtman: Good catch.

alichtman:
That last commit breaks one regression test, and I'm not sure why.

`features/regression.feature:51  Journals with unreadable dates should still be viewable`

Doesn't happen when the commit isn't applied, but I can't find anything in that commit which should cause it to break...

alichtman:
Found the issue. This last commit clobbers `\n` characters in the bodies of journal entries.

image

alichtman:
This is all ready for review. No bugs that I'm aware of.

peter-schmidbauer:
I was unsatisfied with not being able to test it so I did a little digging... Behave does seem to escape the backslashes automatically kind of like r-strings (e.g. `print(r"\n")`). 
It seems like it's easy to "unescape" them again by doing `text.encode("utf-8").decode("unicode_escape")`:

grafik

However that doesn't seem to play nice with unicode characters...

grafik

However, the codecs package provides a better way that works with unicode:

grafik

With this you can transform the text from behave and then compare it to the stdout capture without stripping any colors.

Thanks to [this answer](https://stackoverflow.com/a/57192592/5266392) on Stackoverflow!

alichtman:
That's a nice trick! I got stuck on that problem... I'll take a look at revamping the tests.

peter-schmidbauer:
It's a bit cryptic, so make sure to link to the SO post in a comment. However, in my tests it seemed to work fine. 

Slightly off topic, I just noticed a potential bug:

grafik

The tags in the title seem to have a slightly different color (config for tags is yellow) for some reason and they also seem to add a space behind them.

alichtman:
Ugh. This space bug has been haunting me... I thought I had it all fixed in that last set of commits. Lemme see if I can repro it.

alichtman:
This one is tricky to fix. The way I have this set up makes it difficult to distinguish between these two cases: `@tag. Text` and `@tag - text`. This branch currently breaks a regression test due to this.

peter-schmidbauer:
How about `re.split(r'(?<=\s)([{@}][-+*#/\w]+)', text)`?

peter-schmidbauer:
I'm not sure where you need that re.split, but `re.split(r'([@][-+*#/\w]+)', text)` does what you expected.

alichtman:
> I'm not sure where you need that `re.split`

It's broken out from `Entry.tag_regex()`, which is used in `highlight_tags_with_background_color()`.

```python
text_fragments = re.split(entry.tag_regex(config['tagsymbols']), text)
```

The bug fix will go in either this method, `colorized_text_generator`, or `colorize`.

I'm going to have to think about how this should be working some more. I've spent way too much time on this feature over the past few days and need to take care of some other stuff. If you see a way to solve the issue that causes the tests that I wrote for this feature to fail, definitely lmk. I've got nothing at the moment.

alichtman:
Sweet, the newline bug is fixed. Now the only bug left to fix is the missing space after a tag that is not followed by punctuation. See the first line of the body for an example.

image

alichtman:
Decided to bold all tags and titles. Works great in PyCharm, doesn't work in my terminal.

image


image

This is a problem for another day, though.

wren:
Hey, we just merged a bunch of stuff into master. This just needs a quick rebase on the new master, but then should be ready to go. Thanks, and sorry!

alichtman:
Disclaimer: I did that merge in the GitHub UI. I'm pretty sure it's done correctly but we shall see...

alichtman:
Resolve merge conflict in `poetry.lock`

alichtman:
@wren Any update on this?

wren:
@alichtman I have a full jrnl work day scheduled for Saturday, so I'll be able to go fully through this then. Lots a big one, so I want to make sure to put it through its paces.

jrnl/util.py
```diff
@@ -4,21 +4,30 @@
...
+from string import punctuation, whitespace
 import logging
 
 log = logging.getLogger(__name__)
 
-WARNING_COLOR = "\033[33m"
-ERROR_COLOR = "\033[31m"
-RESET_COLOR = "\033[0m"
+PY3 = sys.version_info[0] == 3
+PY2 = sys.version_info[0] == 2
```
pspeter: Don't add all these constants back in, they were purposefully removed previously.

jrnl/Entry.py
```diff
@@ -49,8 +49,8 @@ def tags(self):
 
     @staticmethod
     def tag_regex(tagsymbols):
-        pattern = fr'(?u)(?:^|\s)([{tagsymbols}][-+*#/\w]+)'
-        return re.compile(pattern)
+        pattern = r'(?<!\S)([{tags}][-+*#/\w]+)'.format(tags=tagsymbols)
+        return re.compile(pattern, re.UNICODE)
```
pspeter: Same with some of these changes. re.UNICODE is not needed and you reverted the fstring.

features/steps/core.py
```diff
@@ -205,7 +217,9 @@ def check_output_time_inline(context, text):
 @then('the output should contain "{text}"')
 def check_output_inline(context, text=None):
     text = text or context.text
-    out = context.stdout_capture.getvalue()
+    out = process_ANSI_escapes(context.stdout_capture.getvalue())
+    if isinstance(out, bytes):
+        out = out.decode('utf-8')
```
pspeter: Same here

jrnl/util.py
pspeter: I also meant the ones below this comment (only excluding the colors)

jrnl/Entry.py
alichtman: I was under the impression that `(?u)` and `re.UNICODE` were the same, so that was intentional. Otherwise, these were merging errors. Sorry about that.

pspeter: That's true actually, they are the same and you can safely remove both.

alichtman: Initially, it was done in this commit: 95b23d433c124de685a1bd9891fdff8f1edb308a

Fixed all the merge errors. Sorry about that.

jrnl/util.py
alichtman: Whoops. Fixed.

wren:
This is looking good so far. I didn't quite get this one finished, but I'll wrap up on Monday night. Thanks for your patience!

Issue #664, opened 2019-09-27
UpgradeValidationException


anoopd:
I have installed latest jrnl using pip and tried to run it  ....... i had two journals `work` and `fitness`  which the new one asked `will be upgraded` and i gave a `yes` but i got

     Upgrading plain text 'default' journal stored in /Users/anoopd/journal.txt...
      Created a backup at /Users/anoopd/journal.txt.backup

    Upgrading plain text 'work' journal stored in ~/work.txt...
    Created a backup at ~/work.txt.backup

     Upgrading plain text 'fitness' journal stored in ~/fitness.txt...
     Created a backup at ~/fitness.txt.backup

     The following journal failed to upgrade:
     default
      Aborting upgrade.`

wren:
This sounds to me like you don't have a default journal. Which I hadn't considered before. If you copy one of your other journal to the default journal location (e.g. `cp ~/work.txt ~/journal.txt`), does that allow the migration to complete? I suspect that should get you unstuck.

If that does work, please report back. It would point to a bug in how we handle journals that exist in the config, but not on the file system.

anoopd:
Sorry to mention , i have already tried that by doing `touch ~/work.txt ` and  `touch  ~/journal.txt` even before i posted this issue .

wren:
@anoopd `touch ~/journal.txt` is going to give you an empty file for a journal, which will likely have an error. Can you try `cp ~/work.txt ~/journal.txt` and see if that works? We do still need to fix the bug, but this command might get you going in the meantime.

azat715:
I got the same error.
`Following plain text journals will upgraded to jrnl 2.0.0:
    default -> /home/archman/Documents/configs/journal.txt
Continue upgrading jrnl? [y/N]
y

Upgrading plain text 'default' journal stored in /home/archman/Documents/configs/journal.txt...
  Created a backup at /home/archman/Documents/configs/journal.txt.backup
DEBUG    jrnl.Journal opened LegacyJournal with 30 entries
DEBUG    jrnl.Journal Imported 30 entries from LegacyJournal to PlainJournal

The following journal failed to upgrade:
default
Aborting upgrade.
Please tell us about this problem at the following URL:
https://github.com/jrnl-org/jrnl/issues/new?title=UpgradeValidationException`

anoopd:
@wren   Thank you , i deleted  `journal.txt` and copied the work.txt as journal.txt and it is working now ! Thank you

wren:
[...]

Issue #630, opened 2019-07-24
Strict date parsing


micahellison:
We have a variety of issues related to date parsing in v2, with the jrnl code interpreting data such as markdown (#443 and #614) and chat logs (#625) as dates simply because they are wrapped in square brackets. In rc3, we tried to tighten up the logic by ignoring anything less than 6 characters, but it's still not a comprehensive fix.

I think, in the perfect world, jrnl would parse dates strictly rather than loosely. That is, it would read the `timeformat` in the config file and use `strptime` with that format to parse it. If it can't be parsed, it's not treated as a date, and instead is just part of the journal entry.

However, this would make changing timeformats impossible without a conversion tool. If you wanted a new format and just changed it in your config file, jrnl would end up reading the whole thing as a single entry with today's date (at least given its current behavior). Not ideal.

There are also concerns related to Python versions - in [this SO thread](https://stackoverflow.com/questions/26165659/python-timezone-z-directive-for-datetime-strptime-not-available) it's said that strptime doesn't support the %z convention until 3.2, and that particular convention was raised in the bug #444.

So if we take this approach, we'd need to add:

- a command the user can enter to change the date format
- very clear documentation about NOT changing the date format in the config
- some live post-date-conversion tests to be absolutely sure the journal doesn't get nuked when changing the date format 
- some basic input validation on the new date format entered by the user (e.g. can it be converted to a Python date and back again with that format while ending up as the same date?)

There are some other, simpler approaches to this problem that I'll make separate issues for, but I think this could be a worthwhile long term goal for jrnl.

zorab47:
It adds check complexity, but allowing multiple date formats to be specified in the config might be a solution. And each would be attempted before ignoring the value.

peter-schmidbauer:
Using parsedatetime to parse the dates from the saved journal breaks a lot of custom date formats (i.e. the German format %d.%m.%Y) because it will parse them incorrectly (i.e. the day as month and vice versa). This is a clear regression and makes v2.0 unusable for me. 

Personally I don't think the journal config should be edited by hand (at all) and instead should be edited through commands (kind of like `git config`). That way, it would be possible to change the format later on.

Alternatively, at least use `strptime` by default everywhere and only fall back to parsedatetime when that doesn't work, as otherwise the custom date format is kinda pointless.

wren:
[...]

PR #694, opened 2019-10-26
Fix parsing Journals using a little-endian date format

pspeter:
Fixes parts of #630 and one source of failure during upgrading to 2.0 (#664).

Ever since version 2.0, when parsing a journal file, jrnl would not use the custom date format string anymore. Instead, it relied on the dateutil library to get the parsing right. This change was made to allow people to change their date format without having to manually change their file. However, this broke some existing date formats like %d.%m.%Y, as it would falsely interpret the month as day and vice versa (when %d <= 12). This commit adds some tests using the little endian date format and fixes the behavior by trying to parse the dates using the custom format first, only falling back to dateutil when needed.

Note that this does not yet allow users to make new entries using the little endian format yet. It will still use the dateutil parser for this. As I usually create new entries with no timestamp or just "today" or "yesterday" as date, this is an okay trade-off for me, but still not ideal. At least upgrading and querying works again properly this way.

### DIFF
--- a/features/custom_dates.feature
+++ b/features/custom_dates.feature
@@ -1,0 +1,35 @@
+Feature: Reading and writing to journal with custom date formats
+
+    Scenario: Loading a sample journal
+        Given we use the config "little_endian_dates.yaml"
+        When we run "jrnl -n 2"
+        Then we should get no error
+        And the output should be
+            """
+            09.06.2013 15:39 My first entry.
+            | Everything is alright
+
+            10.06.2013 15:40 Life is good.
+            | But I'm better.
+            """
+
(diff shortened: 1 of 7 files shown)


peter-schmidbauer:
Any chance we could see this in a 2.0.2 (or 2.1) release? This is blocking me from really using jrnl at the moment along with #699.

wren:
I'm out of the country right now, but should be back in a few days. v2.1 is
already in beta, and soon to be released, so it's too late for that one.

I haven't had a chance to look at this yet, but if all looks good, I don't
see why we can't bump it up into v2.1.1

Issue #463, opened 2017-02-10
Full-text search


shaobozi:

maebert:
Hey Shabozi, can you maybe give an example of what you mean?

shaobozi:
like the command >: jrnl -s "blah" ; will list all entries with the word "blah" .

maebert:
Ah, now I understand.This'll come in the next big release, see also #313.

radomirbosak:
I rebased my earlier PR which solves this: #470 

@shaobozi Note that `-s` is already used for turning on short output (`--short`).

@maebert , do you plan to postpone any search-related features until 2.2 ? What is the current schedule for adding new features?

craftyguy:
Is this feature still planned? Being able to search for any text in a journal would be super helpful, and it would save me from having to @tag every word :P

jyoo:
> Ah, now I understand.This'll come in the next big release, see also #313.

Is this feature still slated for release? Been almost 2 years.

wren:
Yup, this is a good idea for a feature. We'll (re)schedule it for a milestone once we get the rest of the project organized. 

PRs welcome!

dradux:
This has been a blocker for me to use the app in any serious way for over 2 years. Perhaps I am not the target user but search seems to be the cornerstone to journals/notes (assuming you intend to use them in any meaningful way in the future). You will likely need to find something you wrote in the past and relying on adding a tag for everything you _may_ want search on does not seem feasible. Do typical users fall back to grep for this or build their own lucene index?

wren:
@dradux You'll be happy to know that Search is one of our top priorities and now scheduled for an upcoming milestone.

I can't speak for most users (although I'd love to hear other people's approaches), but my personal approach is to open the whole journal in neovim when I want to do any searching of this kind. I do this because I keep most of my journals encrypted, so grep isn't super useful. I use ripgrep for the unencrypted journals sometimes. Either way, it'd be useful to have something built in for users. 

As for building an index, the plain text format, and the encryption feature come together to make building and keeping any sort of metadata (including an index) very tricky in jrnl. I consider both of those features integral to jrnl, so metadata has been a difficult topic to approach, and search will likely depend on it in some way or another (unless we want to search the entire journal every time, which can lead to some performance problems in larger journals).

[...]

PR #698, opened 2019-10-29
Add --delete for interactive removal of entries

### LINKED ISSUE #434: --delete ?
Instead of using --edit then delete/save/quit editor, maybe it would be easier to have a --delete command that would show the selected entries and ask confirmation to delete them, without going through the editor.
Especially convenient when managing todo lists.

$ jrnl @todo buy bread
$ jrnl @todo wash dishes
$ jrnl @todo --delete # I'm done with today's duty
2016-07-30 11:43 @todo buy bread

2016-07-30 11:43 @todo wash dishes

Are you sure you want to delete the selected entries? [y/N]

(And jrnl is a great app by the way!!)

alichtman:
- Add switch for deleting entries. Works with filters, etc.
- Fix some minor style issues

Fix #434

```bash
# First, create a journal named "test"
$ python3 -m jrnl test First entry
$ python3 -m jrnl test Second entry
$ python3 -m jrnl test Third entry
$ python3 -m jrnl test --delete
```

### Checklist


alichtman:
Turns out the python library I used for prompting doesn't support windows: https://github.com/magmax/python-inquirer/issues/63

I could load an alternative library in that case, disable the feature, or find a different library that works on all platforms.

alichtman:
```cucumber
    # The input for this test is <SPACE><ENTER>y
    Scenario: --delete flag allows deletion of single entry
        Given we use the config "deletion.yaml"
        When we run "jrnl --delete"
        And we type " "
        And we type
            """

            y
            """
        When we run "jrnl -on 2019-10-29 -s"
        Then the output should not contain "2019-10-29 11:11 First entry."
```

This structure for the test doesn't work... Testing this feature as it is may not be possible.

image

peter-schmidbauer:
PyInquirer seems like a heavy dependency for a relatively simple feature like this. 
Also, your PR adds another place which parses datetimes without using config["timeformat"], which we should try to avoid as time.parse() doesn't work with all formats. 

Why don't we just do it like `rm -i` and keep it simple?
Something like:

```python
def prompt_delete_entries(entries: List[Entry]):
  print("Confirm each entry you want to delete [N/y]:")
  to_delete: List[Entry] = []
  for entry in entries:
    agreed = input("jrnl: Delete entry '{}'? ".format(entry.pprint(short=True)))
    if agreed == "y":
      to_delete.append(entry)

  return to_delete
```

And then: 

```python
jrnl.entries = [e for e in jrnl.entries if e not in to_delete]
jrnl.write()
```

That would also be much easier to test.

alichtman:
I believe it's a much better user experience with the interactive selection provided by `PyInquirer`. If I want to quickly scan a list of potentially hundreds of journal entries and select only one or two to delete, I'd much rather have the selection checkbox than have to [y/n] every single entry up until the one I'd like to remove.

alichtman:
> Adding a complete command-line UI just for `--delete` creates an inconsistent user experience in my opinion.

I see a natural progression to using this with `--edit`, as well. Any features that involve selecting some subset of journal entries before performing some other action with them (aside from maybe viewing) should have a nice selection UI.

alichtman:
> your PR adds another place which parses datetimes without using config["timeformat"], which we should try to avoid as time.parse() doesn't work with all formats.

This is a fair point though. I have to take a closer look at how to handle this.

peter-schmidbauer:
I may be wrong, but I think the main use case for this would be to delete the last entry. This would be very easy with my proposal, especially when adding a `-y|--yes` option to skip the questions:
`jrnl -1 --delete -y`

It might even make sense to make that the default when none of the selection options are specified (instead of putting up the whole journal for deletion). 

If you are looking for a specific text to delete in the journal then the feature proposed in #463 would help. But if you need to really scan through the whole thing then maybe just going with `jrnl --edit` and dealing with it in your editor would be the better choice anyways.

Sorry if I'm being a downer here, I'm just personally not a big fan of 'GUIifying' a CLI ;)

alichtman:
> Even if you are looking for a specific entry, then you probably saw that entry earlier and can look for it using its date

I'd rather combine these steps into one command. Instead of going fishing for an entry, copying the date or a tag or something, and then running `jrnl <search> --delete`, I'd like to start fishing from `jrnl --delete`.

> But if you need to really scan through the whole thing then maybe just going with jrnl --edit and dealing with it in your editor would be the better choice anyways.

I'm always scared of deleting partial entries / more than I want to when I pull up the raw journal to edit in vim. I always have to scan what I'm deleting multiple times just to be sure...

One nice aspect of the interactive selection option is that it's not possible to delete partial entries / corrupt the journal. Your solution works for small lists of entries, but I'd prefer mine for larger lists.

I do understand your concerns, however. Still thinking about how to best handle this. It'd be great to get some input from other contributors and users.

wren:
This is a tough one. I think the introduction of an interactive UI is a bit too much for now, but you raise some very good points about how cumbersome it can be to deal with querying large amounts of journal entries for the right ones to delete. I need to think about this one a bit more.

@micahellison What do you think?

micahellison:
I'm of a couple minds about this also. I was about to post that we should use filtering rather than an interactive tool, but then it hit me that one reason to delete an entry could be that a duplicate was made on the same specified time. No amount of filtering could rule out that duplicate, right? Having an entry-by-entry confirmation could deal with that situation, but it could also be really annoying for when a user is really sure that they want to delete a range of entries.

And yet, maybe this is what editing is for. After all, people have already been using this tool for seven years, and the edit functionality has been the way to do this. Maybe `--delete` could be for the simple use cases (such as the most recent entry, as @pspeter mentioned), and we recommend users use edit for more complicated situations.

As an aside, I think there is a place for interactive UIs that interact with jrnl and/or jrnl's data. I would love for this tool to be easily available for people that aren't comfortable with the command line. But I think we should take great care in figuring out their points of contact with jrnl (e.g. a plugin system), and setting them up in a way that minimizes their impact on jrnl's core functionality. So even with tricky use cases like the above, my instinct is to lean away from interactivity.

alichtman:
> But I think we should take great care in figuring out their points of contact with jrnl (e.g. a plugin system), and setting them up in a way that minimizes their impact on jrnl's core functionality.

I would agree that having a plugin system is probably the right way to go with this. Also, I'm excited to build a plugin system, so there's that :)

What if we have filtering with confirmation by default (somewhat along the lines of @pspeter's solution), and offer an interactive deletion plugin. Anyone that doesn't want to use it doesn't have to.

wren:
Yup, sounds good! We can have a `--delete` option (no gui) with a prompt to confirm, with a `--force` (no one-letter option for now).

alichtman:
Ah... yeah that rebase doesn't look right. I'm going to have to figure out how to get this all fixed up. 😨

alichtman:
Still needed:

- Tests
- `--force` option

wren:
We can split the `--force` flag into a separate PR, if you'd like.

alichtman:
Yeah, let’s do that.

wren:
@alichtman Hi! This PR still looks really promising. Are you planning on coming back to it? Or would you like someone else to pick it up?

alichtman:
If someone else could take it from where I left off, that'd be great. I've got a lot on my plate at the moment and can't fit this in, but this feature definitely deserves to get finished.

wren:
@alichtman No worries. Thanks for taking it this far!

dbxnr:
@alichtman Would you be interested in collaborating on a fork of the project? This seems like a really good feature and I'd like to do some more work on it with you

wren:
@dbxnr Hi! As this is an open-source project, you're well within your rights to fork the project and do what you'd like with it. But please don't use our project's space to recruit for--or advertise--your other project. It's distracting for us here, and bad open-source community etiquette.

alichtman:
> This seems like a really good feature and I'd like to do some more work on it with you

@dbxnr, if you're interested in improving this feature, I'd recommend opening another PR (since this one was just merged).

PR #703, opened 2019-10-30
 [Stacked on #702] Password confirmation dialog

peter-schmidbauer:
Stacked on #702, rebase on master before merging.

When creating a new password (using `--encrypt`), jrnl will now ask you to repeat the password. It will only encrypt it after you entered the same password correctly twice. 
I've read some people complaining about permanently encrypting their journal because they mistyped their password, this should prevent that.


peter-schmidbauer:
Reopening due to rebasing issues

PR #705, opened 2019-10-31
remove py2 remnants and use mocks in tests

pspeter:
This PR should completely remove all the Python 2 compatibility features and other leftovers.
Closes #701, closes #615, fixes #700, might fix #667.

I also removed the wonky testing utilities in jrnl.utils and instead used unittest.mock.patch to simulate user inputs in the tests. It's a bit unusual as well this way, but at least we don't have to mess with replacing std.stdin and sys.stderr and what not. If somebody knows a simpler way to simulate user input that works with builtins.input, getpass.getpass as well as sys.stdin.read, let me know.

This is a bit of a bigger PR, but the tests all pass.

## Todo
- [x] Remove from \_\_future\_\_ imports
- [x] Switch util functions to builtins print()/input() 
- [x] Rework tests to mock user input
- [x] Remove all mentions of unicode
- [x] Change str.format() to f-strings
- [x] Change codecs.open to builtins.open

### Checklist
- [x] The code change is tested and works locally.
- [x] Tests pass. Your PR cannot be merged unless tests pass
- [x] There is no commented out code in this PR.
- [x] Have you followed the guidelines in our Contributing document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](../pulls) for the same update/change?
- [x] Have you added an explanation of what your changes do and why you'd like us to include them?
- [x] Have you written new tests for your core changes, as applicable?

PR #706, opened 2019-10-31
Password confirmation

### LINKED ISSUE #481: Ask to confirm password
When changing or setting the password for the first time I think it would be helpful to type it twice.

pspeter:
~~Stacked on #705.~~ Rebased on Master.

When creating a new password (using --encrypt), jrnl will now ask you to repeat the password. It will only encrypt it after you entered the same password correctly twice.
I've read some people complaining about permanently encrypting their journal because they mistyped their password, this should prevent that.

This does not work during the install yet, but will in #708. 

Fixes #481. 

### Checklist

### DIFF
--- a/features/encryption.feature
+++ b/features/encryption.feature
@@ -3,29 +3,55 @@
             Given we use the config "encrypted.yaml"
             When we run "jrnl -n 1" and enter "bad doggie no biscuit"
             Then the output should contain "Password"
-            and the output should contain "2013-06-10 15:40 Life is good"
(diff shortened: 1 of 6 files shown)


wren:
v2.1.1 is getting a bit big, so I'm moving this to v2.1.2.

wren:
@pspeter Can you add some non-happy-path tests? Like what happens if you mistype the password on the second try?

pspeter:
@wren Done 👍

PR #708, opened 2019-11-02
[Stacked on #706] Refactor password logic to prevent accidental password leakage

pspeter:
~~Stacked on #706 and #705, merge those before merging this PR. I can also rebase on master afterwards before merging to avoid duplicate commits.~~

After working with the passwords a bit in #706, I noticed how they were sometimes stored in the same dict as all the other journal specific configuration. I also noticed how some of the test files even had password fields in them. 

I was worried that real passwords would accidentally get written to the jrnl config file in the future (and subsequently leaked through peoples dotfile repositories...). To prevent that, I refactored a lot of the code to constrict the password handling logic only to the EncryptedJournal classes and away from the other configuration that gets saved to disk. 

However, I had to change the install progress a little bit to make this work. It will now not ask you for a password right away but rather if you want to encrypt or not. Only if you reply 'y' will it ask you for the password later on.

~~Currently this still breaks the install, which we do not have tests for by the way.~~

### Checklist

### DIFF
--- a/jrnl/util.py
+++ b/jrnl/util.py
@@ -4,8 +4,10 @@ import sys
 import os
 import getpass as gp
 import yaml
+
 if "win32" in sys.platform:
     import colorama
+
     colorama.init()
 import re
 import tempfile
@@ -13,6 +15,7 @@ import subprocess
 import unicodedata
 import shlex
 import logging
+from typing import Optional, Callable
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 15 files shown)


jrnl/util.py
```diff
@@ -44,37 +35,49 @@
...
+def create_password(journal_name: str, prompt: str = "Enter password for new journal: ") -> str:
+    while True:
+        pw = gp.getpass(prompt)
+        if not pw:
+            print("Password can't be an empty string!", file=sys.stderr)
+            continue
+        elif pw == gp.getpass("Enter password again: "):
+            break
+
+        gp.getpass("Passwords did not match, please try again")
```
pspeter: This should be a `print(..., file=sys.stderr)`

pspeter:
```suggestion
        print("Passwords did not match, please try again", file=sys.stderr)
```

pspeter: Fixed in #706

Issue #615, opened 2019-07-08
Drop support for 2.7


maebert:
Since Python is officially dropping support for 2.7 on January 1st, 2020, I would like to drop support for Python 2 in jrnl 2.0 too. Happy to comb through the `v2.0-rc2` branch and make the adjustments, it'll make future maintenance a lot easier.

@wren, @MinchinWeb, @micahellison Any concerns or objections?

wren:
@maebert I agree that we should drop Python 2 support by the end of the year, but trying to get this into v2.0 feels like scope creep. 

The CI tests are already running only on Python 3, so new PRs don't have to support it. And the current v2.0-rc2 build is currently passing the Python 3 tests, so this doesn't seem like a high priority (as long as it's done by 2020). 

Since users have already been stuck on v1.9.8 for a long time, we should avoid anything that could possibly delay v2.0 any longer. v2.0 really needs to be locked down to the existing code, and only include critical bug fixes (with lower priority bug fixes in v2.0.1). We can then follow up with refactors and features after that.

This will get the bug fixes and features already in v2.0 out to users more quickly, which is the highest priority.

micahellison:
I'm also feeling like a solid testable release is worth prioritizing over an overhaul of code, especially if it's mainly dealing with deprecated-but-not-broken language from 2.7, since that sounds like a regression risk. Though since we are only running tests in 3, is there anything we should do on PyPi or Homebrew related to this change?

MinchinWeb:
I have no personal objections to dropped Python 2.7. If people are running a Python 2-only environment, v1.9.8 remains a proven option.

maebert:
Great. We can officially drop support for 2.7 with the release, even though it might still "accidentally" work, and rework the code afterwards. If we move to Poetry with #162 we can specify compatible python versions rigidly for PyPi, for homebrew we need to change the formula (which we need to do anyway for the next release).

peter-schmidbauer:
I agree that support for Python 2 and up to 3.5 should be dropped, however right now the pyproject.toml specifies only versions 3.7+ are supported, is that on purpose? 3.7 has been out for less than 1.5 years, I think we can and should support 3.6 as well.

maebert:
Perfectly reasonable. We're not usinc asyncio or annotations, so 3.6 support should be easily maintainable for a while.

wren:
I have been wanting to add annotations to the codebase, but it's not on the
[...]

PR #710, opened 2019-11-04
Support Python 3.6+

pspeter:
As discussed in #615, we can and should support Python from version 3.6 and up. This PR:
 - adds all supported Python versions to Travis,
 - tells poetry and therefore pypi that 3.6+ is supported.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,8 @@
 dist: xenial   # required for Python >= 3.7
 language: python
-python: "3.7"
+python:
+  - 3.6
+  - 3.7
 git:
   depth: false
 before_install:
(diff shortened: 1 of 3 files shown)


peter-schmidbauer:
Looks like the custom template export fails in 3.8. I'm not familiar with this part of the code, but it seems to throw an Error somewhere during parsing: 
> NotImplementedError
>    "-" * len(entry.title)
> 'Constant' not supported


I'm not sure what to think of the custom template feature in general. It seems stamped together with regex and complicated parsing logic. Do we want to maintain this? I don't even have a clue where the error might be coming from.

wren:
Since it's causing so much trouble, can you drop 3.8 from this PR and file a separate issue for it? That'll let us get 3.6 support in, and deal with the template issues with a bit more time.

pspeter:
> can you drop 3.8 from this PR and file a separate issue for it?

Done!

PR #744, opened 2019-11-14
Exit jrnl if no text entered into editor

### LINKED ISSUE #589: Closing Text editor before composing leads to dump of entire journal contents in terminal.
I use jrnl with gedit under Ubuntu. 

When I call jrnl in my terminal, I get a blank document in gedit, as expected. If I compose something, and then close the window, I get the expected message that `[Entry added to default journal].`

However, if I don't put any characters into the gedit editor and just close it, the *entirety* of my existing journal gets printed *in my terminal*.

Odd. Is that supposed to happen?

alichtman:
Fix #589

Pretty self-explanatory. ~I was trying to think of a creative way to test this but I don't think we can automate exiting the editor cleanly (and judging by the lack of tests for anything relating to an editor, I think everyone else has come to the same conclusion that I did.)~

### Checklist


pspeter:
I see two ways to test this: 

1. (easy) Mock the whole util.get_text_from_editor() function to return a predefined string ("" in this case)
2. (bit harder) Mock the subprocess.call() call to put that predefined string into the tmpfile.

alichtman:
> Mock the whole `util.get_text_from_editor()` function to return a predefined string (`""` in this case)

This solution makes sense to me, but I'm not sure how to actually implement it. Will do some digging...

features/steps/core.py
```diff
@@ -66,6 +67,12 @@ def set_config(context, config_file):
             cf.write("version: {}".format(__version__))
 
 
+@when('we open the editor and exit')
```
pspeter: I would make this test a bit more general so we can reuse it for additional tests. Something like the `run_with_input` method below. 

```
@when('we open the editor and enter')
@when('we open the editor and enter {text}')
```
And then patch the get_text_from_editor with `(context.text or text)`

features/steps/core.py
```diff
@@ -182,6 +189,11 @@ def check_json_output_path(context, path, value):
     assert struct == value, struct
 
 
+@then('the output should be empty')
+def check_empty_output(context, text=None):
+    assert (text or context.text) is None
```
pspeter: You need to check `context.stdout_capture` here, not `context.text` or `text`. You don't need `text` at all actually.

features/steps/core.py
```diff
@@ -66,6 +67,12 @@ def set_config(context, config_file):
             cf.write("version: {}".format(__version__))
 
 
+@when('we open the editor and exit')
+def open_editor_and_exit_without_entering_text(context):
+    with patch('jrnl.util.get_text_from_editor', return_value=""):
```
pspeter: Patching the subprocess.call is a bit cleaner because you would still get the proper console output. Alternatively you can create a function that prints the expected output and add it to `patch(side_effect=<print_func>)`.

alichtman:
Yep, agreed. Will fix when I've got some time.

features/steps/core.py
alichtman: > Alternatively you can create a function that prints the expected output and add it to patch(side_effect=<print_func>)

Seems messy to me, and would require changing both the test and the code if the message was ever modified. I'd rather patch `subprocess.call`

alichtman:
It's unclear to me why this test outputs the following on stderr:

image

On this branch, when I open the editor, write nothing and exit, I get this:

image

features/steps/core.py
```diff
@@ -67,9 +67,11 @@ def set_config(context, config_file):
...
 
 
-@when('we open the editor and exit')
-def open_editor_and_exit_without_entering_text(context):
-    with patch('jrnl.util.get_text_from_editor', return_value=""):
+@when('we open the editor and enter')
+@when('we open the editor and enter {text}')
+def open_editor_and_enter(context, text=""):
+    text = (text or context.text)
+    with patch('subprocess.call', return_value=text):
```
pspeter: If you look at the get_text_from_editor function, it doesn't use the return value of subprocess.call, but instead lets it write something to a tempfile and then reads that tempfile. So your patched function would have to write `text` to that tmpfile. 
You can do this by creating a helper function that takes a `List[str]` (like subprocess.call), takes the last string of that list, which will be the path to the tmpfile, and then writes `text` to that file. You can then use this function with `patch('subprocess.call', side_effect=_mock_editor_function)`.

alichtman: Ah, yep, you're right. Missed that detail.

alichtman: Do you know a good way to pass `text` to that patched function? It won't be in the `subprocess.call` args, but it needs to be accessible from within the patched function that's overriding `subprocess.call`.

pspeter: You could define the function within `open_editor_and_enter()`, then it has access to `text`.

alichtman: I've made those changes in my last commit, however, the patched function doesn't ever seem to be called. Still looking into why.

features/steps/core.py
```diff
@@ -66,6 +67,22 @@ def set_config(context, config_file):
...
 
 
+@when('we open the editor and enter')
+@when('we open the editor and enter {text}')
+def open_editor_and_enter(context, text=""):
+    text = (text or context.text)
+    print("open_editor_and_enter called")
+    def _mock_editor_function(command):
+        print("_mock_editor_function called")
+        tmpfile = command.split()[-1]
```
pspeter: command is already a list
```suggestion
        tmpfile = command[-1]
```

features/steps/core.py
alichtman: It's possible that I'm not patching the right reference to `subprocess.call`, but I've tried it with `jrnl.utils.subprocess.call` as well, but no luck.

pspeter: In travis, the function does seem to be called. The print statements show up in the test output.

alichtman: It's not happening locally for me. I've got no idea how that's possible. I guess I'll just work off the Travis runs for testing :/

alichtman: image

So, the empty file does get written, but for some reason, this code block in `jrnl/cli.py` is never hit:

```python
raw = util.get_text_from_editor(config, template)
if not raw:
	sys.exit()
```

pspeter: That's weird. utils doesn't do a `from subprocess import call` but instead looks up `subprocess.call` during the function call. So patching `subprocess.call` should be correct. Not sure why it's not working on your end. 

I have no idea why raw is not empty either at this point..

jrnl/cli.py
```diff
@@ -196,6 +196,8 @@ def run(manual_args=None):
                     print(f"[Could not read template at '{config['template']}']", file=sys.stderr)
                     sys.exit(1)
             raw = util.get_text_from_editor(config, template)
+            if not raw:
+                sys.exit()
```
pspeter: By the way, I would move this down to the bottom of the big if statement, so that it also exits when not using an editor and writing nothing.

features/steps/core.py
alichtman: Yeah, I just did that. Waiting for Travis.

features/steps/core.py
```diff
@@ -66,6 +67,25 @@ def set_config(context, config_file):
             cf.write("version: {}".format(__version__))
 
 
+@when('we open the editor and enter')
+@when('we open the editor and enter {text}')
```
pspeter: This should fix it 😄 
```suggestion
@when('we open the editor and enter "{text}"')
```

jrnl/cli.py
alichtman: I'm not entirely sure where you mean.

features/steps/core.py
alichtman: > I hate computers sometimes.

This is all I have to say about that.

jrnl/cli.py
pspeter: Where it currently just sets mode_compose to False in line 211

features/steps/core.py
alichtman: That makes behave think it's undefined for some reason.

```python3
You can implement step definitions for undefined steps with these snippets:

@when(u'we open the editor and enter ""')
def step_impl(context):
    raise NotImplementedError(u'STEP: When we open the editor and enter ""')
```

features/steps/core.py
```diff
@@ -66,6 +67,25 @@ def set_config(context, config_file):
             cf.write("version: {}".format(__version__))
 
 
+@when('we open the editor and enter')
+@when('we open the editor and enter "{text}"')
```
pspeter: Apparently you also have to add
`@when(u'we open the editor and enter ""')` to make {text} optional.

alichtman: I just added a "we open the editor and enter nothing" case, which takes advantage of the `text=""` default in the method signature.

pspeter: Apparently you can also make it optional by setting it to {text?} but only if the package `parse-type` is installed...
https://behave.readthedocs.io/en/latest/tutorial.html#python-step-implementations

pspeter:
There we go :D

features/steps/core.py
alichtman: Fixed it.

alichtman:
When this gets merged, can we squash this all into one commit?

alichtman:
I honestly don't think that's necessary. Who's going to be running `jrnl ""`?

pspeter:
`""` denotes an empty string. It's the same scenario as above when somebody closes the editor without writing anything, only without the editor. But without an editor, this line will be called:

https://github.com/jrnl-org/jrnl/blob/4318da5b0642f495916c65f5601e86796b85b7c7/jrnl/cli.py#L202

But this can also be exited without writing anything, leaving you with an empty string again.

alichtman:
Oh, so this would happen if someone tried to `cat` an empty file into `jrnl` on stdin?

pspeter:
That too, I assume. But even if you don't pipe anything into jrnl, it will start an input stream at that line and let you type something in your terminal. You can input multiple lines and then stop the stream with ctrl+d. 

grafik

alichtman:
This test breaks for some unknown reason. Will take a look at it again when I have time. I have to do some other work...

features/core.feature
```diff
@@ -20,6 +20,20 @@ Feature: Basic reading and writing to a journal
         When we run "jrnl -n 1"
         Then the output should contain "2013-07-23 09:00 A cold and stormy day."
 
+    Scenario: Writing an empty entry from the editor
+        Given we use the config "editor.yaml"
+        When we open the editor and enter ""
+        Then we should see the message "[Nothing saved to file]"
+
+    Scenario: Writing an empty entry from the command line
```
alichtman: If we decide we don't care about this last test, we can just delete this block and merge it. I don't think this test is super important and likely won't spend more time debugging it.

features/core.feature
```diff
@@ -20,6 +20,20 @@ Feature: Basic reading and writing to a journal
...
+        Given we use the config "basic.yaml"
+        When we run "jrnl ''"
```
pspeter:
```suggestion
        When we run "jrnl" and enter ""
```

alichtman: That causes the test to hang until `Ctrl-c` or `Ctrl-D` is entered. It does make the test pass when it is, however.

pspeter: I see, the mocking doesn't work with an empty string currently

alichtman: I updated it to work with an empty string. I'm not sure why `sys.stdin.read()` isn't being mocked here.

image

alichtman:
@wren: This PR is also ready for review.

PR #753, opened 2019-11-20
Fix `pyproject.toml`

alichtman:
Something weird happened when merging #692 and changes to the
`pyproject.toml` file were not applied. I can't find the commit that overwrote those changes, and they do appear in the PR...

Regardless, this should fix that.

- `$ poetry add ansiwrap`
- Manually make `colorama` not Windows exclusive
- `$ poetry update`

Actually, now that I look more closely, it doesn't look like my PR was ever merged.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1,4 +1,15 @@
 [[package]]
+category = "main"
+description = "textwrap, but savvy to ANSI colors and styles"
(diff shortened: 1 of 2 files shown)


alichtman:
Oh, sorry. I was looking on `master`. My bad.

Issue #739, opened 2019-11-12
Add Python 3.8 support


pspeter:
I tried adding travis tests for 3.8 before in #710 but they failed for some reason, so for now I just added them for 3.6 and 3.7. We should add 3.8 support ASAP though.

wren:
The test suites are now running 3.8 (linux, mac, and windows). They're under the "Allowed Failures" section for now, but we should figure out what's up with 3.8 (which is failing consistently across all platforms).

mathematicalmichael:
> The test suites are now running 3.8 (linux, mac, and windows). They're under the "Allowed Failures" section for now, but we should figure out what's up with 3.8 (which is failing consistently across all platforms).

just an FYI: I didn't pay attention to the dependencies and ended up building a docker image based on alpine python 3.8.1 and jrnl 2.1.post2 (via pip) and haven't had any trouble with it. I went through the changelog and saw that 3.6+ support only comes in 2.1.1 and thought it was worth checking out what I have been using.

https://github.com/mathematicalmichael/docker-env/blob/master/jrnl/Dockerfile was last built Jan 20 and I've been relying on it since.

Issue #619, opened 2019-07-16
Windows Support


wren:
We need to talk about Windows support on this project. We say we support it, but don't treat it equally as Mac or Linux. 

I want to remedy this. We should either: 
1. Treat Windows as a first-class citizen, or
2. Drop support for it. 

I would like to officially support it, but it's going to require a few updates. I'm hoping this issue can serve as a place to gather requirements for what needs to be done to fully support Windows. We can then make some issues and get some work done. 

Here's the issues I can think of at the moment (in no particular order): 

- **Run the tests on Windows** - There's currently no way for us to tell if a PR broke anything on Windows without a lot of manual testing. But Travis doesn't properly support a Windows Python environment for testing. Maybe we need a second pipeline for running Windows tests? Maybe move from Travis? Or something else?
- **Distribute jrnl on a Windows-based package manager** - We distribute on Homebrew, but nothing that focuses on Windows (Pip can be a bit unwieldy on Windows). Maybe Chocolatey would be a good fit for this?

Does anyone have any other things to point out for this?

micahellison:
This came up in #177 also - there's talk of Cygwin, MinGW, and later WSL, but limiting to those would definitely alienate some if not most potential Windows users.

As far as distribution, I like using Chocolatey, but I think most Windows users that might be interested are just going to want to go find a place where they can download an EXE, which we can handle by attaching the binaries to the releases in GitHub.

That leaves the build pipeline. Apparently Travis "works" with Windows and Python but doesn't necessarily "support" it? I'm not sure if that's acceptable, though. It looks like the main drawback is that it has to install python every time it spins up a VM to build, so all builds take at least a couple minutes. I like Travis otherwise and have no idea what it would be like to have two separate build pipelines. Maybe someone else has opinions or ideas about this?

MinchinWeb:
Software for Windows is typically distributed as an EXE, typically an installer, but sometimes as a "portable" version. Package managers aren't really a thing on Windows, at least the same way they are on Linux. The closest thing Windows has to an official package manager is the Windows Store. Chocolately seems largely targeted as system administrators, and (at least the free version) doesn't actually host any of the installers, but is rather a collection of Powershell scripts (think bash) that download the official installers off the official sites, and then try to automate the install to not require user input.

Python is a good example of the above: you can download and install Python via an EXE installer off of Python.org, there is an official option to install Python from the Windows Store, or there's an unofficial version on Chocolately that run the Python.org installer under the hood.

[...]

PR #759, opened 2019-11-28
More Travis-CI Testing

MinchinWeb:
This is to expand the Travis testing. Specifically, it adds:

- Python 3.8 (on Linux) as an "allowable failure". See #739. It looks like an easy fix to get this working. Once working, this can removed from the allowable failure list.
- the developmental version of Python (on Linux), as an allowable failure. This is designed as a forward-looking test to help us be aware of anything in the next Python version that might break `jrnl`. It also has the bonus of helping the Python core devs by exposing the cutting edge Python code to lots of real life Python code.
- a test not in UTC. When I was re-adding the DayOne tests (#742), I discovered that the code path changes depending on if the timezone is UTC or now. However, I expect few user's machines are set to UTC, so it seems like an oversight to to only test in UTC.
- MacOS. Because why not? Runs under Python 3.7.4. We have had MacOS-specific bugs in the past (see #130).
- Windows, as an allowable failure. This is currently tested under Python 3.7, as on Python 3.8 `poetry` tries to install `cryptography` as a locally-built wheel (rather than a wheel downloaded from PyPI), which fails. This test currently times out after 10 minutes, as one of the `behave` steps is to open the editor and get text, and I think it is *really* opening the editor, rather than pretending to. This is odd, because it doesn't do this on my local machine.... c.f. #619

I haven't implemented it here, but the format of Travis should make it simple to add stages to the testing sequence, for example to run all the code through `black` or check that the dependencies have all been updated.

### Checklist
- (n/a) The code change is tested and works locally.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,4 @@
 dist: xenial   # required for Python >= 3.7
-language: python
 jobs:
   include:
     - name: "Python 3.6 on Linux"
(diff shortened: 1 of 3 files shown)


MinchinWeb:
Glad to help!

PR #769, opened 2019-12-07
Black Formatter

MinchinWeb:
This adds the `black` formatter as Linting stage to the Travis-CI setup. The existing code was also adjusted, as needed, to pass the linter.

This linting stage has to pass before the regular tests are run.

### Checklist
- [x] There is no commented out code in this PR. (Except for added comments, but then those aren't really code, are they?)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -113,6 +113,7 @@ jobs:
       before_deploy:
         - poetry config http-basic.pypi "$PYPI_USER" "$PYPI_PASS"
         - poetry version "$TRAVIS_TAG"
+        - echo __version__ = \"$TRAVIS_TAG\" > jrnl/__version__.py
         - poetry build
       deploy:
         - provider: script
@@ -124,4 +125,3 @@ jobs:
         - git add pyproject.toml
         - git commit -m "Incrementing version to ${TRAVIS_TAG}"
         - git push https://${GITHUB_TOKEN}@github.com/jrnl-org/jrnl.git master
-
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,6 +53,13 @@ jobs:
     - os: windows
 
   include:
+    - stage: Lint & Tests
+      name: Lint, via Black
+      python: 3.8
+      # black is automatically installed by peotry
+      script:
+        - black --check . --verbose
+
     # Python 3.6 Tests
     - name: Python 3.6 on Linux
       python: 3.6
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,7 +58,7 @@ jobs:
       python: 3.8
       # black is automatically installed by peotry
       script:
-        - black --check . --verbose
+        - black --check . --verbose --diff
 
     # Python 3.6 Tests
     - name: Python 3.6 on Linux
--- a/features/steps/core.py
+++ b/features/steps/core.py
@@ -5,8 +5,11 @@ from jrnl import cli, install, Journal, util, plugins
 from jrnl import __version__
 from dateutil import parser as date_parser
 from collections import defaultdict
-try: import parsedatetime.parsedatetime_consts as pdt
-except ImportError: import parsedatetime as pdt
+
+try:
+    import parsedatetime.parsedatetime_consts as pdt
+except ImportError:
+    import parsedatetime as pdt
 import time
 import os
 import json
@@ -17,7 +20,7 @@ import shlex
 import sys
 
 consts = pdt.Constants(usePyICU=False)
-consts.DOWParseStyle = -1 # Prefers past weekdays
+consts.DOWParseStyle = -1  # Prefers past weekdays
 CALENDAR = pdt.Calendar(consts)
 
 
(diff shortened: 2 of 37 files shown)


wren:
Thanks for the PR. I'll review this on Tuesday (my next jrnl work day).

Small point of order: the contributing guidelines ask that we open a new issue for new features. This gives us a chance to talk about things before you put the work into implementing it. In this case, I do happen to like Black already, but please try to follow the guidelines in the future.

MinchinWeb:
I created issue #770 as a place to have some discussion around this!

.travis.yml
```diff
@@ -53,8 +53,17 @@ jobs:
     - os: windows
 
   include:
+    - stage: Lint
```
wren: I don't think linting should be a separate stage. If we run this in the test stage, it'll run in parallel with everything else. But in its own stage, it will delay the other tests from even starting until it has completed. I know black is fast, but Travis still has to spin up the vm and clone the repo, etc, so there's can be up to a minute delay waiting for this.

features/steps/core.py
```diff
@@ -119,24 +127,26 @@ def run_with_input(context, command, inputs=""):
         text = iter([inputs])
 
     args = ushlex(command)[1:]
+    # fmt: off
```
wren: Why are we turning it off for this file?

.travis.yml
```diff
@@ -53,8 +53,17 @@ jobs:
...
+      name: Lint, via Black
+      python: 3.8
+      install:
+        - pip install black
```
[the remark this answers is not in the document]
MinchinWeb: This surprises me but I haven't much with `poetry`. Does this mean that `black` is installed for everyone who install `jrnl`?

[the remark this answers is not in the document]
MinchinWeb: (If so, I don't think `black` should be installed for "normal" users)

.travis.yml
```diff
@@ -53,8 +53,17 @@ jobs:
...
+      script:
+        - black .
```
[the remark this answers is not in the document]
MinchinWeb: done

.travis.yml
MinchinWeb: I've changed it. One advantage of a "Linting" stage is it can "fail fast" to let people know that (hopefully minor) changes are needed.

features/steps/core.py
MinchinWeb: Because otherwise the next three lines become rather mangled and less readable. I.e. this:

```
with patch("builtins.input", side_effect=_mock_input(text)) as mock_input,\
     patch("getpass.getpass", side_effect=_mock_getpass(text)) as mock_getpass,\
     patch("sys.stdin.read", side_effect=text) as mock_read:
```

becomes this:

```
with patch("builtins.input", side_effect=_mock_input(text)) as mock_input, patch(
    "getpass.getpass", side_effect=_mock_getpass(text)
) as mock_getpass, patch("sys.stdin.read", side_effect=text) as mock_read:
```

To get `black` to leave the context manager alone (i.e. the `with` statement), the whole block has to be turned off as `black` requires it's on/off to be at the same indentation level.

.travis.yml
wren: Yup, that's definitely an advantage, but that still happens even if the jobs are being run in parallel (the build will fail with the first failed job). The rest of the jobs will still run, but that can also be useful in case there's more than 1 kind of problem with the build.

.travis.yml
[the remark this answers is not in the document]
wren: If you look at `pyproject.toml`, you'll see that there are two sets of dependencies, the regular dependencies, and the dev dependencies. In our build pipeline, we install the dev dependencies since we need behave to run the tests (behave is also not installed for regular users).

features/steps/core.py
wren: Ugh, yeah, that's ugly. Looks like it might be a while for that to get fixed, too (https://github.com/psf/black/issues/557).  :(

wren:
Hey, this is going to be very prone to merge conflicts, and I don't want you to try to hit a moving target. Can you take out the changes made by black itself, and just do the travis file and the `fmt` comments? I can run black when this gets merged so there are no conflicts.

wren:
It is otherwise ready for approval, though. Thank you!

MinchinWeb:
@wren I just updated it to the current 'develop' branch. Mostly as a way to get Travis-CI to run (maybe..). It should be a clean merge at the moment.

MinchinWeb:
Black doesn't want to make any changes locally, but want to on Travis, so I added the `--diff` option to figure out why.

features/steps/core.py
```diff
@@ -60,12 +60,15 @@ def read_journal(journal_name="default"):
 def open_journal(journal_name="default"):
     config = util.load_config(install.CONFIG_FILE_PATH)
     journal_conf = config["journals"][journal_name]
-    if (
-        type(journal_conf) is dict
-    ):  # We can override the default config on a by-journal basis
+
+
+    if type(journal_conf) is dict:
+         # We can override the default config on a by-journal basis
```
wren:
```suggestion
        # We can override the default config on a by-journal basis
```

PR #773, opened 2019-12-14
[DayOne] YAML export improvements

MinchinWeb:
This is split out of #159 et al to just contain those parts that extend the YAML exporter to also export DayOne metadata.

The code has been "black formatted".

Currently, `jrnl` may be stripping this metadata, so another PR (also split from #159) will be added to fix that.

### Checklist
- [x] Tests pass. Your PR cannot be merged unless tests pass -- ish. I have 6 failing tests, but I think it's Windows issues rather than anything I changed here.
- [x] Have you written new tests for your core changes, as applicable? -- I don't think any apply really...

### DIFF
--- a/jrnl/plugins/yaml_exporter.py
+++ b/jrnl/plugins/yaml_exporter.py
@@ -17,8 +17,8 @@ class YAMLExporter(TextExporter):
     def export_entry(cls, entry, to_multifile=True):
         """Returns a markdown representation of a single entry, with YAML front matter."""
         if to_multifile is False:
-            print("{}ERROR{}: YAML export must be to individual files. "
-                  "Please specify a directory to export to.".format("\033[31m", "\033[0m"), file=sys.stderr)
+            print("{}ERROR{}: YAML export must be to individual files. Please \
+                specify a directory to export to.".format(ERROR_COLOR, RESET_COLOR, file=sys.stderr))
             return
 
         date_str = entry.date.strftime(entry.journal.config['timeformat'])
@@ -29,13 +29,13 @@ class YAMLExporter(TextExporter):
         # see also Entry.Entry.rag_regex
         multi_tag_regex = re.compile(r'(?u)^\s*([{tags}][-+*#/\w]+\s*)+$'.format(tags=tagsymbols))
 
-        '''Increase heading levels in body text'''
+        """Increase heading levels in body text"""
         newbody = ''
         heading = '#'
         previous_line = ''
         warn_on_heading_level = False
-        for line in entry.body.splitlines(True):
-            if re.match(r"^#+ ", line):
+        for line in body.splitlines(True):
+            if re.match(r"#+ ", line):
                 """ATX style headings"""
                 newbody = newbody + previous_line + heading + line
                 if re.match(r"^#######+ ", heading + line):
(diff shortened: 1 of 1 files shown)


MinchinWeb:
@wren : the issue to support this is #159 -- [this comment](https://github.com/jrnl-org/jrnl/issues/159#issuecomment-99232692) in particular

jrnl/plugins/yaml_exporter.py
```diff
@@ -80,9 +77,30 @@ def export_entry(cls, entry, to_multifile=True):
         dayone_attributes = ""
         if hasattr(entry, "uuid"):
```
wren: Does every Day One entry have a uuid? If so, can we indent all of the following conditionals so that they're nested inside of this conditional? That way, we don't check the conditions for all entries if we know they aren't related to Day One.

jrnl/plugins/yaml_exporter.py
```diff
@@ -19,11 +19,10 @@ def export_entry(cls, entry, to_multifile=True):
...
         if to_multifile is False:
             print(
-                "{}ERROR{}: YAML export must be to individual files. "
-                "Please specify a directory to export to.".format(
-                    "\033[31m", "\033[0m"
-                ),
-                file=sys.stderr,
+                "{}ERROR{}: YAML export must be to individual files. Please \
+                specify a directory to export to.".format(
+                    ERROR_COLOR, RESET_COLOR, file=sys.stderr
```
wren: 👍

jrnl/plugins/yaml_exporter.py
```diff
@@ -33,17 +32,15 @@ def export_entry(cls, entry, to_multifile=True):
...
 
         """Increase heading levels in body text"""
         newbody = ""
         heading = "#"
         previous_line = ""
         warn_on_heading_level = False
-        for line in entry.body.splitlines(True):
-            if re.match(r"^#+ ", line):
+        for line in body.splitlines(True):
+            if re.match(r"#+ ", line):
```
wren: What is this change for?

jrnl/plugins/yaml_exporter.py
MinchinWeb: It should, but I'm leery of assuming that will always be the case.

It may make sense to indent lines 88 through 99 in any case.

jrnl/plugins/yaml_exporter.py
MinchinWeb: **The first line**: The "body" (the entry's main text) is slightly pre-processed before doing anything with it (see lines 31-32), and this is just completing the switch to that.

**The second line**: I don't think that change should be there....

jrnl/plugins/yaml_exporter.py
MinchinWeb: I've intended the second block of lines in 31a040878c6bc2b302c0e7709e305e3955701b6f

jrnl/plugins/yaml_exporter.py
MinchinWeb: Fixed the second line in 31a040878c6bc2b302c0e7709e305e3955701b6f

MinchinWeb:
@wren I think this one is ready to go!

jrnl/plugins/yaml_exporter.py
```diff
@@ -80,9 +77,32 @@ def export_entry(cls, entry, to_multifile=True):
         dayone_attributes = ""
         if hasattr(entry, "uuid"):
             dayone_attributes += "uuid: " + entry.uuid + "\n"
-            # TODO: copy over pictures, if present
-            # source directory is  entry.journal.config['journal']
-            # output directory is...?
+        if (
```
wren: Sometime in the future, I'd love to split this off into a DayOne-specific class, so it's not clogging up the yaml exporter.

MinchinWeb:
@wren Can this be merged in? It looks like everything is good, and you've signed off on it. Not sure what's next...

wren:
Yup, this will likely go into the next version. Just focusing on the v2.2 beta version right now.

PR #785, opened 2020-01-05
Add UTC support for failing DayOne tests

MinchinWeb:
Travis-CI (typically) runs in UTC, and UTC behaves weird compared to other timezones. This adds UTC support to the DayOne parts of the code.

Thus this re-enables the (most of) DayOne tests (all of which broke previously, because of UTC), building on #742 .

The code change is pulled from #350.

### Checklist

### DIFF
--- a/jrnl/DayOneJournal.py
+++ b/jrnl/DayOneJournal.py
@@ -52,16 +52,11 @@ class DayOne(Journal.Journal):
                     except (KeyError, pytz.exceptions.UnknownTimeZoneError):
                         timezone = tzlocal.get_localzone()
                     date = dict_entry["Creation Date"]
-                    try:
+                    # convert the date to UTC rather than keep messing with
+                    # timezones
+                    if timezone.zone != "UTC":
                         date = date + timezone.utcoffset(date, is_dst=False)
-                    except TypeError:
-                        # The above converts the date to UTC. However, if the
-                        # timezone is already in UTC, you can't call
-                        # timezone.utcoffset() with the is_dst parameter
-                        # ("utcoffset() got an unexpected keyword argument
-                        # 'is_dst'"), but we already have our date in UTC, so
-                        # nothing more needs to be done.
-                        pass
+
                     entry = Entry.Entry(
                         self,
                         date,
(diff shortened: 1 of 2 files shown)


jrnl/DayOneJournal.py
```diff
@@ -52,7 +52,16 @@ def open(self):
                     except (KeyError, pytz.exceptions.UnknownTimeZoneError):
                         timezone = tzlocal.get_localzone()
                     date = dict_entry["Creation Date"]
-                    date = date + timezone.utcoffset(date, is_dst=False)
+                    try:
```
wren: Passing on errors is an anti-pattern I'd like to get us out of whenever possible. 

How about something like this?
```
if dict_entry.get("Time Zone") in pytz.all_timezones:
    timezone = pytz.timezone(dict_entry["Time Zone"])
else:
    timezone = tzlocal.get_localzone()

date = dict_entry["Creation Date"]
if timezone.zone != "UTC":
    date = date + timezone.utcoffset(date, is_dst=False)
```

wren: Sorry, apparently didn't hit submit to the above comment yesterday (found it sitting in a tab today).

MinchinWeb: I agree that generally passing on an a raised error is bad from, but what about cases like this where the raised error is actually telling us things are working as expected? This from also supports the philosophy of "it's easier to ask for forgiveness than permission", which is often typically idiomatic Python.

The [article](https://devblogs.microsoft.com/python/idiomatic-python-eafp-versus-lbyl/) I found on this (first link on Google) gives an example and explicitly uses the `except: pass` formation. The [Python docs](https://docs.python.org/3.8/glossary.html#term-eafp) seem to support this form as well.

So I think we should leave it as it is as being more idiomatic Python.

MinchinWeb: @wren : a decision here is what's holding to #786. I like it this way, and think it's more "Pythonic", but of you still want it changed, I'm happy to make the change. Just key me know.

wren: Yes, please. Thank you!

MinchinWeb:
@wren: changed as requested! This, and #786 should be good to go!

Issue #714, opened 2019-11-06
Config updating algorithm only looks at top-level keys 


alichtman:
The config updating algorithm in `upgrade_config(config)` in `install.py` only looks at top-level keys, and not nested keys.

A solution [like this](https://stackoverflow.com/a/27266178) will be needed if the subkeys of any config keys will ever be updated.

alichtman:
Don’t close.

PR #791, opened 2020-01-10
upgrade_config checks for nested changes

dbxnr:
Fix for #714 - the upgrade_config algorithm recursively replaces missing key/value pairs with default values.

### Checklist


jrnl/install.py
```diff
@@ -61,19 +62,16 @@ def module_exists(module_name):
...
-        save_config(config)
-        print(
-            f"[Configuration updated to newest version at {CONFIG_FILE_PATH}]",
-            file=sys.stderr,
-        )
+    for k, v in new_config.items():
+        new_value = user_config.get(k)
+        if isinstance(v, Mapping) and isinstance(new_value, Mapping):
+            user_config[k] = v
+            upgrade_config(new_value, v)
```
wren: Good use of recursion.

jrnl/install.py
```diff
@@ -113,7 +111,9 @@ def load_or_install_jrnl():
             print("Exiting.", file=sys.stderr)
             sys.exit(1)
 
-        upgrade_config(config)
+        if config.get("version") and config.get("version") < __version__:
```
wren: Why the change in behavior here?

dbxnr: Optimisation, without this check it would try to upgrade the config file on every invocation

dbxnr:
I'm really not sure I can test it with behave while default_config is hard-coded into install.py unless I'm missing something?

There are quite a few environmental variables that probably shouldn't be in there, moving them to jrnl.conf would make it testable... What are your thoughts?

PR #794, opened 2020-01-10
Add import support for Day One v2

dbxnr:
Add import support for DayOne2.

usage `jrnl --import dayone2 some_file.json` converts a DayOne2 formatted json file and appends it to the journals in jrnl.yaml

### Checklist


MinchinWeb:
So does this work by importing the DayOne2 export JSON, or by using that JSON as a "journal file" directly?

dbxnr:
Right now it reads a JSON exported from DayOne2, it should be straightforward to add write capabilities

dbxnr:
I like the chevron syntax, it seems quite intuitive, (although ```jrnl journal.txt > dayone2``` makes more sense) however it makes the ```--import``` and ```--export``` a little redundant...

Import would always require a file, so having a separate flag for path (currently ```-i```) is a little redundant, ```jrnl --import dayone2 [path]``` works well.  Importing implies an export, it's really a conversion, so I'm handling it by automatically saving as the input filename with a .json suffix, unless an output filename is given.

I'll open a separate PR for it, but there's some real spaghetti code going on with the argparser.

wren:
> I like the chevron syntax, it seems quite intuitive, (although `jrnl journal.txt > dayone2` makes more sense) however it makes the `--import` and `--export` a little redundant...

I see what you're saying, but this would make knowing which format to use when importing/exporting quite difficult (since DayOne isn't the only format that we support). There would be no good way to know which format we're writing to when doing something like `jrnl journal.txt > dayone2`. 

> Import would always require a file, so having a separate flag for path (currently `-i`) is a little redundant, `jrnl --import dayone2 [path]` works well.

An import doesn't necessarily require a file (you can pipe from any program directly to jrnl import, or even copy/paste if you'd like), but I do see the point you're making about the `-i` flag. As long as we don't break the ability to pipe in/out of jrnl, we can maybe clean up the interface for when a user wants to provide a filename to jrnl, since that seems likely to be the most often scenario.

> Importing implies an export, it's really a conversion, so I'm handling it by automatically saving as the input filename with a .json suffix, unless an output filename is given.

Yup, you are exporting from DayOne2 to json, and then importing from json to jrnl. Also, please remember that this isn't the only json format jrnl supports, so we can't assume that any json file is a DayOne2 export.

> ... there's some real spaghetti code going on with the argparser.

Yup, that is some of the oldest code in jrnl, and definitely needs some love. We've been considering a more robust framework (probably [Click](https://click.palletsprojects.com), but not sure), but haven't settled on an implementation, yet.

> I'll open a separate PR for it...

It's okay to update this PR, but whatever is easier for you is good with me.

dbxnr:
> There would be no good way to know which format we're writing to when doing something like `jrnl journal.txt > dayone2`.

In this case, `dayone2` is a flag for the file export format - so the default behaviour would be to convert journal.txt into dayone2 format and say is as journal.json  An optional flag could specify the output filename.

For importing, `jrnl new_journal.txt < some_json.json`, the schema could be inferred by the JSON parser, it could try validating aginst accepted formats until one passes.

> We've been considering a more robust framework (probably Click, but not sure), but haven't settled on an implementation, yet.

Dropping the import flag seems to be dependent on this, Click seems like a solid library and it should be able to use a string suffixed with `.txt` followed by `>` or `<` in place of `--import`

dbxnr:
The functionality is pretty much completed now, and the `-i` flag has been dropped (partly because the argparse code is such a mess it was easier to drop an unnecessary flag than refactor code that's possibly going to be removed...)

So to try it out, `jrnl --import dayone2 /features/data/journals/dayone2.json` will create a jrnl formatted file called `dayone2.txt` in that directory.

I still need to write tests for it, and would appreciate if you could take a look at #798 since the tests need that fix.

jrnl/plugins/dayone2_importer.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def parse_json(self):
+        entries = []
+        for entry in self.json["entries"]:
+            entries.append(
+                Entry(
+                    self,
+                    date=datetime.strptime(entry["creationDate"], "%Y-%m-%dT%H:%M:%SZ"),
+                    text=entry["text"],
+                    starred=entry["starred"],
```
MinchinWeb: The UUID and tags should also be parsed and added to the entry. See https://github.com/jrnl-org/jrnl/blob/develop/jrnl/DayOneJournal.py#L62..L66 for how it is done with the DayOne classic journal.

Two questions:

1. If we import the DayOne2 journal, and find a match on the UUID in the existing journal we're importing into, what do we do? Display a diff and ask the user what to do?
2. This drops any attached photos or audio recordings. With the DayOne Classic journals, I mostly ignored this because if there was a photo, it would be in a given folder, relative to the entry, named with the UUID of the entry as the filename for the photo. So I could infer the photo location, while with DayOne2, that isn't possible. Maybe this doesn't need to be solved here, but in that case, we should raise an issue on the tracker here so it can be addressed later.

dbxnr: I'm not sure I understand the purpose of saving the UUID since the date (with second resolution) is effectively a UUID.  Without knowing how DayOne2 calculates their UUIDs, introducing theirs to journals could potentially lead to duplicates in future, although that is admittedly unlikely.  Not having explicit UUIDs littered around from an exported journal would keep the .txt much cleaner as well, and reduce overhead for parsing it out when exporting.

>1. If we import the DayOne2 journal, and find a match on the UUID in the existing journal we're importing into, what do we do? Display a diff and ask the user what to do?

If either a date match or UUID match it found, I suppose it would make sense to print a diff and give a choice of choosing the old, new or aborting the import.

>2. This drops any attached photos or audio recordings. 

I'm not sure how to deal with this. My instinct is that since jrnl is a slim text-based app the most pragmatic option would be to either ignore the photos altogether.  The JSON export contains the image filename along with an MD5 hash, so this could be used to identify the associated file however the only context it has is from the associated journal entry.  I'm concerned that introducing code to manage decontextualized images would lead to feature bloat.

wren:
> > There would be no good way to know which format we're writing to when doing something like `jrnl journal.txt > dayone2`.
> 
> In this case, `dayone2` is a flag for the file export format - so the default behaviour would be to convert journal.txt into dayone2 format and say is as journal.json An optional flag could specify the output filename.

Ah, I see. Unfortunately, `>` is a [shell metacharacter](http://faculty.salina.k-state.edu/tim/unix_sg/shell/metachar.html) and can't be used in this way.

> For importing, `jrnl new_journal.txt < some_json.json`, the schema could be inferred by the JSON parser, it could try validating against accepted formats until one passes.

Unfortunately, since the filename is on the other side of the `<` metacharacter, jrnl would have no way to see what filename the content is coming from. So, we would have no way to tell a json import from a yaml import, or any other filetype apart from each other. In this scenario, we'd have to guess at each file type from the incoming text until we got it right. Which is not a great place to be, and is how--I think--the current arg parser got where it is today.

> > We've been considering a more robust framework (probably Click, but not sure), but haven't settled on an implementation, yet.
> 
> Dropping the import flag seems to be dependent on this, Click seems like a solid library and it should be able to use a string suffixed with `.txt` followed by `>` or `<` in place of `--import`

Due to the reasons above, I just don't see how any library can help implement this proposal. We'll need to keep the import flag in order to continue supporting input and output redirection in the shell.

jrnl/plugins/dayone2_importer.py
MinchinWeb: > I'm not sure I understand the purpose of saving the UUID

The most obvious use would be on import. If your destination journal is a DayOne Classic journal, UUID's are going to be assigned to the entries on creation anyway, so I think it would be best to re-use the existing ones, if possible. If the same entries (possibly changed in DayOne2 on your phone) are re-imported, it makes it easy to line up what entries have changes.

I guess one of the workflows I see myself (continuing) doing is keeping a DayOne2 "journal" on my phone, exporting it to import into `jrnl`, and then exporting it out of `jrnl` via the YAML exporter to feed into Pelican, to turn everything into a static website. 

> My instinct is that since jrnl is a slim text-based app the most pragmatic option would be to ignore the photos altogether. The JSON export contains the image filename along with an MD5 hash, so this could be used to identify the associated file however the only context it has is from the associated journal entry. I'm concerned that introducing code to manage decontextualized images would lead to feature bloat.

My thought he is just to retain the data (e.g. the photo filenames, etc) as part of the Entry when imported, a little like how we try and maintain the weather information. Whether (and how) it gets saved can somewhat be handed off to the Journal and Exporter classes. I'm not envisioning any photo-handling code, although it may (down the road) make sense to add some as part of an exporter.

dbxnr:
```
photos: [
  {
      "exposureBiasValue":0.0,
      "height":0,
      "identifier":"9F0BB56D4EDE4BC29779ED935BBEC3F8",
      "md5":"4e2058da5f377369b4d33d994f1661ae",
      "orderInEntry":0,
      "type":"jpg",
      "width":0
   
}]
```
This is the relevant part of the DayOne2 JSON - would it be useful to append the filename to the end of the entry body, perhaps with a prefix to assist in search?

```
[2020-01-10 12:22] Entry Number One.
@awesome @great 📷9F0BB56D4EDE4BC29779ED935BBEC3F8.jpg

```
As far as the UUID goes, I'm still pretty unconvinced to be honest. They only work as intended if you're using a consistent algorithm. Since there's no way of extracting any meaningful data from the UUID or guarantee it's uniqueness in the journal, the timestamp functions as a conveniently readable UUID.

> If we import the DayOne2 journal, and find a match on the UUID in the existing journal we're importing into, what do we do? Display a diff and ask the user what to do?

I circumvented this issue by creating a new journal to hold the imported data.

dbxnr:
This is pretty much done at this point, although there's a lot of metadata in the exported DayOne2 that's not currently being used.
```

"location":{
            "address":"",
            "administrativeArea":"",
            "country":"",
            "latitude":,
            "localityName":"",
            "longitude":,
            "placeName":"",
            "region":{
               "center":{
                  "latitude":,
                  "longitude":
               
},

         "weather":{
            "conditionsDescription":"Light Rain and Windy",
            "pressureMB":989.7,
            "relativeHumidity":0.82,
            "temperatureCelsius":3.0,
            "visibilityKM":10.399,
            "weatherCode":"rain",
            "weatherServiceName":"Forecast.io",
            "windBearing":247.0,
            "windChillCelsius":0,
            "windSpeedKPH":9.9
         
}
```

It might be good to include the useful points, like placeName and conditionsDescription?  I'm really hesitant to start junking up the journal with data by default, so the options appear to be:

- Cherry pick a few data points to include by default on import
- Ignore all the metadata
- Add a flag to specify what metadata to import, something like `jrnl --import dayone2 file.json -m [weather || location || photos || audio || timezone || uuid ^ all]`

wren:
I appreciate the work here, but this has some big incompatibilities with the current user flow for jrnl. I think we'll have to pass for now. Please feel free to discuss implementation details in an issue and resubmit when ready.

PR #795, opened 2020-01-10
Fix for tests not handling successive input prompts

dbxnr:
Refactor enabling behave to properly handle multiple successive inputs.

Previously a test with successive prompts wouldn't correctly iterate over the mock inputs, treating them as a single input.  Single letter inputs are correctly parsed and iterated over. See tests in #796

Refactored some existing tests to be correctly handled as text.

### Checklist
- [x] The code change is tested and works locally.
- [x] Tests pass. Your PR cannot be merged unless tests pass
- [x] There is no commented out code in this PR.
- [x] Have you followed the guidelines in our Contributing document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](../pulls) for the same update/change?
- [x] Have you added an explanation of what your changes do and why you'd like us to include them?
- [x] Have you written new tests for your core changes, as applicable?

Issue #690, opened 2019-10-25
jrnl upgrade does not fail gracefully when a journal file is missing from the configuration


alichtman:
This is an edge case, but should still be handled gracefully.

```bash
$ jrnl
Welcome to jrnl v2.0.1.

It looks like you've been using an older version of jrnl until now. That's
okay - jrnl will now upgrade your configuration and journal files. Afterwards
you can enjoy all of the great new features that come with jrnl 2:

- Support for storing your journal in multiple files
- Faster reading and writing for large journals
- New encryption back-end that makes installing jrnl much easier
- Tons of bug fixes

Please note that jrnl 1.x is NOT forward compatible with this version of jrnl.
If you choose to proceed, you will not be able to use your journals with
older versions of jrnl anymore.

Following plain text journals will upgraded to jrnl v2.0.1:
    default -> /Users/alichtman/journals/journal.txt
    tech    -> /Users/alichtman/journals/tech.txt
Continue upgrading jrnl? [y/N]
y

Upgrading plain text 'default' journal stored in /Users/alichtman/journals/journal.txt...
  Created a backup at /Users/alichtman/journals/journal.txt.backup

Upgrading plain text 'tech' journal stored in /Users/alichtman/journals/tech.txt...
  Created a backup at /Users/alichtman/journals/tech.txt.backup
Traceback (most recent call last):
  File "/usr/local/bin/jrnl", line 10, in <module>
    sys.exit(run())
    │        └ <function run at 0x107fba5f0>
    └ <module 'sys' (built-in)>
  File "/usr/local/lib/python3.7/site-packages/jrnl/cli.py", line 148, in run
    config = install.load_or_install_jrnl()
             └ <module 'jrnl.install' from '/usr/local/lib/python3.7/site-packages/jrnl/install.py'>
  File "/usr/local/lib/python3.7/site-packages/jrnl/install.py", line 92, in load_or_install_jrnl
    upgrade.upgrade_jrnl_if_necessary(config_path)
    │                                 └ '/Users/alichtman/.jrnl_config'
    └ <module 'jrnl.upgrade' from '/usr/local/lib/python3.7/site-packages/jrnl/upgrade.py'>
  File "/usr/local/lib/python3.7/site-packages/jrnl/upgrade.py", line 95, in upgrade_jrnl_if_necessary
    backup(path)
    │      └ '/Users/alichtman/journals/tech.txt'
    └ <function backup at 0x107d7bc20>
[...]

PR #796, opened 2020-01-10
Fix for upgrade with missing journal

dbxnr:
Fix for #690 - if a journal is found to be missing during upgrade the user will be prompted to create it or cancel the upgrade.

Tests depend on #795 

### Checklist
- [ ] Tests pass. Your PR cannot be merged unless tests pass


jrnl/upgrade.py
```diff
@@ -11,10 +11,22 @@
...
+    try:
+        with open(filename, "rb" if binary else "r") as original:
+            contents = original.read()
+
+        with open(filename + ".backup", "wb" if binary else "w") as backup:
+            backup.write(contents)
+    except FileNotFoundError:
+        print(f"\nError: {filename} does not exist.")
+        try:
+            cont = util.yesno(f"\nCreate {filename}?", default=False)
```
wren: I think we want to skip adding the journal to the updated config, rather than creating the file. Creating the file could lead to some awkward experiences with encrypted journals, and possibly break synced files by accidentally overwriting, so it's best to skip.

jrnl/upgrade.py
```diff
@@ -144,6 +156,7 @@ def upgrade_jrnl_if_necessary(config_path):
         j.write()
 
     print("\nUpgrading config...", file=sys.stderr)
+
```
wren: Please be careful committing random changing to different parts of files that what you're working in.

features/upgrade.feature
```diff
@@ -35,3 +35,16 @@ Feature: Upgrading Journals from 1.x.x to 2.x.x
             10.06.2013 15:40 He said "[this] is the best time to be alive".
             """
         Then the journal should have 2 entries
+
+    @skip_win
```
wren: Do these fail on Windows? We don't want to add new tests that already don't pass on Windows, while we're trying to fix the Windows tests.

features/data/configs/upgrade_from_195_with_missing_journal.json
```diff
@@ -0,0 +1,11 @@
+{
+"default_hour": 9,
+"timeformat": "%Y-%m-%d %H:%M",
+"linewrap": 80,
+"encrypt": false,
```
wren: Can we add the same tests but for an encrypted journal? Just be sure it works.

features/upgrade.feature
```diff
@@ -35,3 +35,16 @@ Feature: Upgrading Journals from 1.x.x to 2.x.x
...
+    Scenario: Successful upgrade with missing journal
+        Given we use the config "upgrade_from_195_with_missing_journal.json"
+        When we run "jrnl -ls" and enter "Y Y"
```
wren: Successive input prompts should be on new lines, so it's clear what is happening.

```suggestion
        When we run "jrnl -ls" and enter
        """
        Y
        Y
        """
```

features/data/configs/upgrade_from_195_with_missing_journal.json
dbxnr: I've added a test for this which is failing, although the actual code works.  I'm not certain, but it could be that the output it being piped to stderr which doesn't have an equivalent "the output should contain".  Can you take a look?  I'll add a step in core.py if so.

Issue #793, opened 2020-01-10
Python 3.9 build is failing with import error ('Mapping' from 'collections')


wren:
The tests that we run on the nightly version of Python have started failing with this error:

```
[ImportError]                                                    
cannot import name 'Mapping' from 'collections' (/home/travis/virtualenv/pyt  
hon3.9-dev/lib/python3.9/collections/__init__.py)
```

I haven't had a chance to look into this, yet, but 3.8 just came out, and 3.9 is still in alpha, so we have some time. But we should fix this soon. 

Example: https://travis-ci.com/jrnl-org/jrnl/jobs/274130952

clbarnes:
Abstract base classes are now in `collections.abc` rather than `collections`; pretty sure is the case for all versions jrnl supports. This should have raised deprecation warnings for a few versions (`collections.abc` was new in 3.3) so you may have some warning-silencing code you can get rid of once the change is made.

wren:
@clbarnes Thanks for looking into this! That seems like it would be an easy fix if it were in jrnl. I guess we'll have to do a dependency audit to deal with this.

clbarnes:
My strategy was just to install the environment and grep in site-packages for `Mapping`, but I only got as far as using the version from pip as poetry disagrees with my dev setup. Does that sound like it would be a sufficient strategy? Wheels get unzipped when they're installed, right?

dbxnr:
passlib is definitely importing `collections.Mapping` - (pwd.py:13) 
pytz is importing it as well, although it's checking for an ImportError (lazy.py:8)

There are a few others, but they're checking the python version before importing (Py2 compatibility)

clbarnes:
Hm, it looks like passlib also has it in a `try`/`except` (in master, anyway). Are the dependency versions in pyproject up to date?

wren:
@clbarnes The dependencies in the lock file should be up to date as of about 2 weeks ago.

MinchinWeb:
It appears the actual issue here (or at least the first time Travis hits it) is with `virtualenv`. Upstream ticket --> https://github.com/pypa/virtualenv/issues/1449

Per the upstream ticket, this should be fixed in newer versions of `virtualenv`...

In any case, this is why the dev version of Python is on the testing matrix...

wren:
Good find, @MinchinWeb!

micahellison:
[...]

PR #826, opened 2020-01-29
[WIP] [Travis] manually upgrade virtualenv

MinchinWeb:
This is a test...

to fix issues with dev version of Python 3.9, try to manually upgrade virturalenv from the get-go.

See #793


MinchinWeb:
So this doesn't fix it yet...

PR #840, opened 2020-02-08
Get rid of dumb sex joke

wren:
These kinds of jokes make us look bad, and have gotten this project bad
PR on social media. It's best to get rid of them. We can still be super
glib and funny without crass jokes.


ivanguy:
yo, i liked the example with girlfriend was really meaningful. With texas it's not really obvious what u can do with that command

wren:
@ivanguy Hi! Thanks for speaking up. Can you please explain how using a different tag obfuscates the documentation?

PR #842, opened 2020-02-08
Pretty print journal entries (add color)

wren:
This was stuck in a temp branch due to our old merging strategy. We need to get everything into develop and master asap or we'll just end up with ever-increasing merge conflicts.

### Checklist
- [ ] Tests pass. Your PR cannot be merged unless tests pass
- [ ] There is no commented out code in this PR.
- [ ] Have you followed the guidelines in our Contributing document?
- [ ] Have you checked to ensure there aren't other open [Pull Requests](../pulls) for the same update/change?
- [ ] Have you added an explanation of what your changes do and why you'd like us to include them?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ before_install:
   - date
 
 install:
-  - pip install poetry
+  - pip install --upgrade poetry
   - poetry install
   - poetry run python --version
 
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -35,8 +35,9 @@ and can be edited with a plain text editor.
   - `highlight`
     if `true`, tags will be highlighted in cyan.
   - `linewrap`
-    controls the width of the output. Set to `false` if you don't
-    want to wrap long lines.
+    controls the width of the output. Set to `false` if you don't want to wrap long lines.
+  - `colors`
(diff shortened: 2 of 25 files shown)


alichtman:
Any update on this?

wren:
@alichtman Been focusing on the Homebrew release, but I have some time set out this weekend to go through the open PRs.

wren:
The failing test on this is caused by `process_ANSI_escapes` on windows (since windows paths have baskslashes.

alichtman:
Are you sure that's the right root cause? I'm having some trouble tracing the issue back to that function.

wren:
I thought so, but I'll take another look at it tomorrow. I had a version of the test that was passing on Windows, and the only difference was that function wasn't being invoked for the test.

wren:
@alichtman I got rid of the function and everything seems to be working properly now. Do you remember why this function was there? Is it required?

alichtman:
I added that function because some tests were failing, relating to Unicode and ANSI escapes, and that fixed the problem. If they’re no longer failing, it’s no longer needed.

alichtman:
Thanks for helping get this merged in! Know when develop will be merged into master?

PR #850, opened 2020-02-12
Interactive delete

dbxnr:
Fix the broken test for #698


jrnl/Journal.py
```diff
@@ -1,13 +1,15 @@
...
 import os
 import sys
 import re
+
 from datetime import datetime
-import logging
+from itertools import filterfalse
+from jrnl import Entry
+from jrnl import util
+from jrnl import time
```
alichtman:
```suggestion
from jrnl import Entry, util, time
```

PR #872, opened 2020-03-11
Improve wording

hkdobrev:
I  was confused a bit when skimming through the docs and went back to re-read this sentence. I think it's more accurate word order if you keep the phrase "any * you want" closer together and it's easier to understand.

This is just a documentation change submitted to the `gh-pages` branch.

### DIFF
--- a/usage/index.html
+++ b/usage/index.html
@@ -256,7 +256,7 @@ to have an editor configured in your journal configuration file (see
 </code></pre>
 
 <p>Will open your editor with all entries tagged with <code>@texas</code> and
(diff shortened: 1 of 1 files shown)


wren:
Hi, I think this is a good update. The `gh-pages` branch is auto-generated from the markdown files in  the [docs directory](https://github.com/jrnl-org/jrnl/tree/develop/docs), though. Can you please make the change in there instead?

wren:
Hi, again! I'm closing this for now, but please feel free to resubmit with the requested changes. Thank you!

PR #887, opened 2020-03-22
Test Version match

MinchinWeb:
Adds a test to ensure that `jrnl --version` outputs the version as defined in the package metadata.

Turns out it wasn't, so I updated it.

I needed a TOLM library to read `pypackage.toml`, so I picked the library `poetry` was using: `tomlkit`. It's added as a dev dependency (since it's only used for the test).

### Checklist

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,0 +1,12 @@
+# These are supported funding model platforms
+
+github: # Replace with up to 4 GitHub Sponsors-enabled usernames e.g., [user1, user2]
+patreon: # Replace with a single Patreon username
+open_collective: jrnl
+ko_fi: # Replace with a single Ko-fi username
+tidelift: # Replace with a single Tidelift platform-name/package-name e.g., npm/babel
+community_bridge: # Replace with a single Community Bridge project-name e.g., cloud-foundry
+liberapay: # Replace with a single Liberapay username
+issuehunt: # Replace with a single IssueHunt username
+otechie: # Replace with a single Otechie username
+custom: # Replace with up to 4 custom sponsorship URLs e.g., ['link1', 'link2']
--- a/.travis.yml
+++ b/.travis.yml
@@ -115,6 +115,8 @@ jobs:
 
     # ... and beyond!
     - name: Python nightly on Linux
+      before_install:
+        - sed -i 's/^python = ">=3\.6\.0.*"$/python = "*"/' pyproject.toml
       python: nightly
 
     # Specialty tests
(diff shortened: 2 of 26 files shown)


MinchinWeb:
> I checked the release itself, and it looks like everything is fine there. The version was updated properly, but wasn't checked back in to our repo.

So I guess that is a bug with the deploy script. Do I need to file a bug report for that elsewhere? Or does the existing #888 work?

Also, I don't know where to start tackling that myself, so I'll leave it to someone else.

> I think we can add a tag to the new test to make sure it runs during deployment, and only then.

I agree with the first part (that the test should run on the to-be-deployed artifact), but I'm think it's worthwhile keeping the test running the rest of the time too. The test itself is very fast, so time isn't a concern. And keeping the test in the rest of the time will keep someone from accidentally bumping the version number the "wrong" way.

Also, only running the test during deployments makes it hard to test the test.

That said, if we have to tag the test to run during deployment and that keeps it from running the rest of the time, I could live with that too.

> Using `@only` with Behave's active tag schema seems like maybe a great candidate for this.

This would require the testing environment to be different during deployment compared to "normal" testing runs. Is there an environmental variable we can test for?

Perhaps it would better to just use ["regular" tags](https://behave.readthedocs.io/en/latest/tag_expressions.html)?

> I'd rather our build process know as little about Poetry's internals as possible.

All we need to know is the TOML key that stores the version number. This is more than nothing of Peotry's internals, but very limited.

> Poetry's `version` command already outputs the version from the toml file

But that's not the only it does. It also pulls the program name from somewhere, which actually means we have to re-extract the version number from the Poetry output, to then test against the `jrnl` output.  The current test is `asset pyproject_version in jrnl_cli_version`.

* `jrnl --version` --> jrnl version v.2.3-beta
* `poetry version` --> jrnl v2.3-beta
* `pyproject.toml` key ["tool"]["poetry"]["version"} --> v2.3-beta

Let me know what (if anything) you directly want changed.

MinchinWeb:
I've update this PR to use `toml` (rather than `tomlkit`) as `toml` was already being installed by `black`.

I've also added a tag to this test, so it can be specifically called before deployment (when it is most needed). To do so:

```sh
behave --tags=deployment_tests
```

Note that it (at least for me locally) will output something for each of the other tests, but skips them.

This doesn't actually add this testing to the deployment script.

---

There was some discussion about pulling the version number from the *poetry* CLI output, rather than the *poetry* configuration file. I could never get that to work. See #888 for discussion, and https://github.com/MinchinWeb/jrnl/tree/version-match-direct for my WIP (non-working) code.

MinchinWeb:
@wren I think I got the conflicts resolved. Should be good to merge!

PR #910, opened 2020-04-07
[Travis] update Windows Python versions

MinchinWeb:
Updates Python 3.7.5 to 3.7.7 and 3.8.0 to 3.8.2.
~Also add the alpha 5 build of Python 3.9 as an allowable failure.~ Poetry doesn't support Python 3.9 yet. See https://github.com/python-poetry/poetry/issues/2154

### Checklist
- [ n/a ] The code change is tested and works locally.
- [ ? ] Tests pass. Your PR cannot be merged unless tests pass

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+version: ~> 1.0
 dist: xenial   # required for Python >= 3.7
 os: linux
 language: python
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,3 @@
-version: ~> 1.0
 dist: xenial   # required for Python >= 3.7
 os: linux
 language: python
@@ -49,12 +48,11 @@ aliases:
         - /c/Python36
         - /c/Python37
         - /c/Python38
-        - /c/Python39
         - /c/ProgramData/chocolatey/lib
         - /c/ProgramData/chocolatey/bin
         - /c/Users/travis/AppData/Local/pypoetry/Cache
     before_install:
-      - choco install python --version $JRNL_PYTHON_VERSION --pre
+      - choco install python --version $JRNL_PYTHON_VERSION
       - python -m pip install --upgrade pip
       - pip --version
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,13 +41,14 @@ aliases:
     os: windows
     language: shell
     env: &env_windows
-      PATH: /c/Python36:/c/Python36/Scripts:/c/Python37:/c/Python37/Scripts:/c/Python38:/c/Python38/Scripts:$PATH
+      PATH: /c/Python36:/c/Python36/Scripts:/c/Python37:/c/Python37/Scripts:/c/Python38:/c/Python38/Scripts:/c/Python39:/c/Python39/Scripts:$PATH
       PYTHONIOENCODING: UTF-8
     cache:
       directories:
         - /c/Python36
         - /c/Python37
         - /c/Python38
+        - /c/Python39
         - /c/ProgramData/chocolatey/lib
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,3 +1,4 @@
+version: ~> 1.0
```
micahellison, line 1: I looked this up -- https://docs.travis-ci.com/user/build-config-validation -- and it looks cool, but I think this belongs in a separate PR, unless it has something to do with Python versions that I don't see.

.travis.yml
```diff
@@ -48,11 +49,12 @@ aliases:
         - /c/Python36
         - /c/Python37
         - /c/Python38
+        - /c/Python39
```
micahellison, line 52: Remove this?

.travis.yml
```diff
@@ -96,7 +98,7 @@ jobs:
       python: 3.7
       env:
         <<: *env_windows
-        JRNL_PYTHON_VERSION: 3.7.5
+        JRNL_PYTHON_VERSION: 3.7.7
```
micahellison, line 101: Can this PR update Python versions for all OSes instead of just Windows? Best to keep them all updated at once.

.travis.yml
```diff
@@ -48,11 +49,12 @@ aliases:
         - /c/Python36
         - /c/Python37
         - /c/Python38
+        - /c/Python39
         - /c/ProgramData/chocolatey/lib
         - /c/ProgramData/chocolatey/bin
         - /c/Users/travis/AppData/Local/pypoetry/Cache
     before_install:
-      - choco install python --version $JRNL_PYTHON_VERSION
+      - choco install python --version $JRNL_PYTHON_VERSION --pre
```
wren, line 57: Is `--pre` still useful if 3.9 isn't being added?

.travis.yml
MinchinWeb: I needed it to figure out what was happening under the covers with Travis. Otherwise, I was having some silent failures. So it's connected in the sense that it gets the rest of the code here.

.travis.yml
MinchinWeb: done

.travis.yml
MinchinWeb: Do you know where to find a list of the MasOS versions available? Windows versions are coming from [Chocolatey](https://chocolatey.org/packages/python/#versionhistory), but I'd assumed that the MacOS versions were tied to the base image...

.travis.yml
MinchinWeb: done

.travis.yml
MinchinWeb: It looks like you can use Homebrew to install custom versions of Python, but that is beyond my experience... see http://www.pyslvs.com/blog/yuan_2018-09-22.html

The [Travis documentation](https://docs.travis-ci.com/user/reference/osx/) doesn't actually say what Python versions are available on what OSX images, just that at least one Python version will always be available.

MinchinWeb:
@wren @micahellison updated, as requested (other than the MacOS versions).

Changing the MacOS Python versions requires switching to Homebrew for installing Python, which is beyond my expertise.

This should be good to merge!

.travis.yml
MinchinWeb: removed for now

.travis.yml
micahellison: On the Mac builds, you can set `JRNL_PYTHON_VERSION` without having to modify anything else. Python comes pre-installed on macOS, so the build uses `pyenv install` to install whatever version the `JRNL_PYTHON_VERSION` is set to.

Issue #889, opened 2020-03-22
Syntax problems aren't caught by current testing


wren:
## Bug report

A bug came in recently (#878) that was a simple syntax issue. We should have caught this, but it slipped into a release. We have a patch already in the works, but I'm starting this issue to come up with an approach for preventing these types of problems from getting into releases in the future.

Maybe it's as simple as adding a linting step to our tests. If this were the case, I'd want it to be strict enough to catch syntax problems, but not so strict as to frustrate contributors with nit-picky suggestions.

mlcdf:
Hello Jonathan,

I stumble upon this issue by accident, but since I've been investing linters for Python at work, I thought I might as well share my thoughts with you.
I think something like `pyflakes` would do the job just fine.

I retrieved the code to before the fix for #878 and I run `pyflakes` :
```
[maxime@nepture jrnl]$ time pyflakes jrnl/
jrnl/install.py:4: 'getpass' imported but unused
jrnl/install.py:10: '.Journal.PlainJournal' imported but unused
jrnl/install.py:11: '.EncryptedJournal.EncryptedJournal' imported but unused
jrnl/FolderJournal.py:5: '.Entry' imported but unused
jrnl/__init__.py:3: 'os' imported but unused
jrnl/EncryptedJournal.py:13: 'typing.Optional' imported but unused
jrnl/Journal.py:4: '.util' imported but unused
jrnl/time.py:60: use ==/!= to compare str, bytes, and int literals
jrnl/plugins/yaml_exporter.py:5: 'os' imported but unused
jrnl/plugins/jrnl_importer.py:5: '..util' imported but unused
jrnl/plugins/jrnl_importer.py:18: local variable 'old_entries' is assigned to but never used
jrnl/plugins/markdown_exporter.py:5: 'os' imported but unused

real	0m0,215s
user	0m0,192s
sys	0m0,022s
```
As you can see, it's crazy fast, and I would have caught the error in `time.py` (plus others things).
One thing to note thought, it only parses the source files, it doesn't note import them (that why it's so fast).

While investigating, I noticed that the Makefile already as a "lint" entry using `flake8`, but it's not used in the CI.
Since `flake8` use `pyflakes` amongst other libs, the following is similar to the command I run above :
```
[maxime@nepture jrnl]$ flake8 jrnl/ --select F 
jrnl/install.py:4:1: F401 'getpass' imported but unused
jrnl/install.py:10:1: F401 '.Journal.PlainJournal' imported but unused
[...]

Issue #770, opened 2019-12-10
Linting Requirements on Travis-CI


MinchinWeb:
So I was mucking around the depths of Travis-CI and realized that it supports multi-stage builds/testing. As it turns out, we're already doing this, as pushing to PyPI has been added as a "deploy" stage.

The other common use for multi-stages is to have a "linting" stage, to confirm code style before running the main body of tests. I wrote a Pull Request (#769) to add `black` as such a linting test, but maybe it would be good to hear others' thoughts on this.

1. Should we have a mandatory linting stage, required to pass before a Pull Request is generally accepted? On the plus side, this keeps the code clean, and hopefully makes it easier to people to jump right into the code base. On the minus side, this can sometimes serve as an impediment to (new) people submitting pull requests.

2. If we have this linting stage, is there any other linters that we should include? `black` is focused on coding style. Thoughts that come to mind include:

    - isort -- sorts imports -- I like this a lot, but it does take a little configuration to get black and isort to play nice together.
    - mypy -- I don't think we currently use type hints, and it may not make sense on a codebase of this size.
    - flake8 -- code issues and "smells" -- under the hood it seems to call `pycodestyle`, `pyflakes`, and `mccabe`
    - is the poetry lockfile up to date -- see #747 -- Personally, I think this is a helpful check, but it's tied to a particular date and time in a way that other linters aren't, meaning it could pass now, but not in an hour (if an update to a dependency is released)
    - some others for consideration --> https://pre-commit.com/hooks.html

As a side, it would be nice if the fix were simple for a bot to come along after a failing test and propose the code that would fix it as a Pull Request to the Pull Request.

micahellison:
The mandatory linting sounded a little draconian to me at first, but after some consideration, I think it's preferable to just having a warning. The alternative would be that an accepted PR that doesn't pass linting could keep subsequent PRs in a warning state, through no fault of their contributors.

MinchinWeb:
I feel like black isn't too bad in this case either, because you can run `black` over your code and it will fix it for you.

MinchinWeb:
Do we want to add `isort` to this list?

PR #911, opened 2020-04-11
[Travis] add linting via pyflakes

MinchinWeb:
See #889, See #770

Note that this only adds the linting, and doesn't add actually fix the issues, so the tests on this are expected to fail. (When I added the black formatter, it proved to be very difficult to both add the formatter and provide the fixes at the same time). Another pull request will follow with the needed fixes.

It also updates the GitHub PR template to remind people that this linting is needed, and includes the commands needed to run the linting and the tests.

### Checklist
- [ ] Tests pass. Your PR cannot be merged unless tests pass -- *see above*
- [ n/a ] Have you written new tests for your core changes, as applicable?

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,4 +1,5 @@
 
+<!--
 # **TEMPLATE PLEASE EDIT**
 *Thank you for wanting to contribute! Please fill out this description as well as look at the checklist!*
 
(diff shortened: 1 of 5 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -9,9 +10,16 @@
 - Example of usage (if applicable)
 - Example of changes to config files (if applicable)
 *
+-->
+
 ### Checklist
 - [ ] The code change is tested and works locally.
-- [ ] Tests pass. Your PR cannot be merged unless tests pass
+- [ ] Tests pass. Your PR cannot be merged unless tests pass. -- `behave`
```
wren, line 17: I think adding the commands here is a good idea, but we should use the poetry or make commands here, since that's what's recommended in our documentation.

wren:
I took care of the pyflakes part already (I want to have it in place for the upcoming beta). But please feel free to open new PR with the PR template upgrades.

PR #933, opened 2020-05-02
Fixes #932

### LINKED ISSUE #932: "jrnl --delete something" deletes all journal entries
## Bug report

  - Jrnl version: (run `jrnl -v`)
  jrnl version v2.4

  - How you installed Jrnl
  pip install --user -U jrnl
  (previous version: 2.2, installed with pip)

  - Operating system [MacOS, Linux, Windows?]
  Linux openSUSE Tumbleweed

  Add some entries to a journal, try to delete an entry setting a filter, say a string. Even if you don't delete any entry when prompted, only the entries that match the filter will be preserved in the journal. **ALL the other entries will be deleted**.

  `$ jrnl Just a test`

  `$ jrnl Another test`

  `$ jrnl "Don't delete me"`

  `$ jrnl -n 5`
  2020-05-02 17:12 Just a test

  2020-05-02 17:13 Another test

  2020-05-02 17:13 Don't delete me

  `$ jrnl --delete nothing`

  `$ jrnl -n 5`

  `$ `

   The "--delete" option should only delete the selected entries.

  Every time I execute `jrnl --delete something` the only content of the journal file become a single line like this:
  `gAAAAABerY4t0dY_yIR4XMot62Aw7cFDIbNExvnCL9TXbLbUpjNA0yQo1nVBJ6yPl0CnN3OAs4cLVB4ePiYN-V2EnGDzkZ6_ng==`
  It changes every time I execute the command.

justdoit176:
# **Fixes critical bug #932 ** 

*Short block of text containing:
- Relevant changes in text form
  Fixes deletion of ALL entries when deleting entries with some filter set
- related issues
  #932

### Checklist
- [X] Tests pass. Your PR cannot be merged unless tests pass
- [X] Have you checked to ensure there aren't other open [Pull Requests](../pulls) for the same update/change?

### DIFF
--- a/jrnl/cli.py
+++ b/jrnl/cli.py
@@ -478,5 +478,8 @@ def run(manual_args=None):
         journal.write()
 
     elif args.delete:
+        other_entries = [e for e in old_entries if e not in journal.entries]
         journal.prompt_delete_entries()
+        journal.entries += other_entries
+        journal.sort()
         journal.write()


jrnl/cli.py
```diff
@@ -478,5 +478,8 @@ def run(manual_args=None):
         journal.write()
 
     elif args.delete:
+        other_entries = [e for e in old_entries if e not in journal.entries]
```
wren, line 481: We try to avoid looping through entire journals as much as possible. This can be slow for very large journals (you'd be surprised how large some user's journals are).

jrnl/cli.py
```diff
@@ -478,5 +478,8 @@ def run(manual_args=None):
...
         journal.prompt_delete_entries()
+        journal.entries += other_entries
```
wren, line 483: Adding the old entries back at this point means that the journal is still being deleted on disk (by the `write` in `prompt_delete_entries`). This then adds them back. So, the problem here is that the journal is now being written out twice, which is slower, and has the possibility to still delete the journal entries (in case something goes wrong in between the two writes).

jrnl/cli.py
```diff
@@ -478,5 +478,8 @@ def run(manual_args=None):
...
+        journal.sort()
```
wren, line 484: All `sort`s should happen in the open method for a journal. By keeping them there, we avoid resorting (again, this helps very large journals).

justdoit176:
Hi @wren !I agree with you, there are performance issues in my PR, it was just a quick&dirty fix to a very critical bug. It was just a copy/past from 

I tried to add this scenario to the tests but it didn't work, for some reason it doesn't catch the bug:
```
    Scenario: --delete -n 1 allows deletion of single entry
        Given we use the config "deletion.yaml"
        When we run "jrnl -n 1"
        Then the output should contain
        """
        2019-10-29 11:13 Third entry.
        """
        When we run "jrnl --delete -n 2" and enter
        """
        N
        Y
        """
        When we run "jrnl -n 1"
        Then the output should contain
        """
        2019-10-29 11:11 Second entry.
        """
```

wren:
Ok, that's fair. I'm in the process of releasing `v2.4.1` which disables the delete flag. That'll give us some time to give this a proper fix without affecting users.

justdoit176:
Better solution has been implemented in #935

Issue #284, opened 2014-09-25
Editor-created entries are lost if you enter the wrong encryption password three times


StevenMaude:
To reproduce:
1. Use an external editor.
2. Write an entry, save and quit.
3. Enter the wrong password three times.
4. Bemoan the loss of your well-crafted entry.

The journal entry is not saved (fine), but `jrnl` [immediately exits](https://github.com/maebert/jrnl/blob/master/jrnl/util.py#L50) and the temporary file containing the entry has already been [deleted](https://github.com/maebert/jrnl/blob/master/jrnl/util.py#L123) (not so fine).

What might be useful would be an option to print the entry text, so that it's possible for the user to recover their work and try again.

Issue #270, opened 2014-08-16
Ask password before writing entry


ayedo:
Is there a reason why jrnl asks for the password after an entry was written? 

I was using jrnl with an external editor and wrote an elaborate entry. I did not notice I accidentally pressed caps lock, and after three wrong attempts it seems as if my entry is simply lost?

Issue #256, opened 2014-07-28
2.0-rc1: encrypted entry fails


yarko:
With 2.0-rc1:

Actual:
- encrypted jrnl:  password (key) not asked for until after journaling
  - fails ungracefully (crash), resulting in lost entry;
- listing existing encrypted jrnl seems to work
  - `jrnl -3` for an encrypted jrnl works;

Wanted:
- get key before invoking external editor, etc.;
- if key (encryption) fails, allow graceful retry;
  - at very least, do it before entry, so that re-running `jrnl` is all that's needed, and that it's before typing an entry occurs;

PR #951, opened 2020-05-12
Ask for password before adding entry instead of after

### LINKED ISSUE #799: Wrong password 3 times after closing editor results in loss of all changes
** Current behavior **
After closing your editor (either a new entry or editing an old one), if you have an encrypted journal, you are prompted for your password. If you enter the correct password, your entry is saved and everything is fine. If you get the wrong password, you are re-prompted up to a total of 3 times. If you get the password wrong all 3 times,  you see `Extremely wrong password.` message, and jrnl quits.

The problem with this is that the temp file has already been deleted by the time you are prompted for your password, so you lose all changes.

**Expected behavior**
Since there's no risk of leaking encrypted data at this point in the flow, we should reopen the editor for the user and give them a chance to do something with the data they were just working on.

ollybritton:
This hopefully fixes #799 because prviously it would allow you to create an entry using your editor and then attempt to open the journal. This behaviour is fine for unecrypted journals but with encrypted journals it means that if you get the password wrong three times then the entry you wrote is lost.

Swapping the code around so that it attempts to open the journal before it opens an editor means that this can't happen. This should solve #270, #256 and #284 too.

It appears that this was the [behaviour in earlier versions of `jrnl`](https://github.com/jrnl-org/jrnl/blob/9618b74f8db84e7847bf15fa46838f100663db0a/jrnl/cli.py#L162) but I'm assuming that this got lost after a rewrite.

This is my first time contributing to open source so I apologise if I've formatted anything wrong or made a big mistake but I'm not really sure what the process is yet and would be happy to fix anything that I did incorrectly.

### Checklist

- [X]  Tests pass. Your PR cannot be merged unless tests pass. --
  `poetry run behave`
- [X] The code passes linting via
  [black](https://black.readthedocs.io/en/stable/) (consistent code styling). --
  `poetry run black --check . --verbose --diff`
- [X] The code passes linting via [pyflakes](https://launchpad.net/pyflakes)
  (logically errors and unused imports). -- `poetry run pyflakes jrnl features`
- [X] Have you checked to ensure there aren't other open
  [Pull Requests](../pulls) for the same update/change?
- [x] Have you added an explanation of what your changes do and why you'd like
  us to include them?

### DIFF
--- a/jrnl/cli.py
+++ b/jrnl/cli.py
@@ -345,7 +345,7 @@ def run(manual_args=None):
     else:
         _exit_multiline_code = "press Ctrl+D"
 
-    # This is where we fi±nally open the journal!
+    # This is where we finally open the journal!
     try:
         journal = open_journal(journal_name, config)
     except KeyboardInterrupt:
--- a/jrnl/cli.py
+++ b/jrnl/cli.py
@@ -345,6 +345,13 @@ def run(manual_args=None):
     else:
         _exit_multiline_code = "press Ctrl+D"
 
+    # This is where we fi±nally open the journal!
+    try:
+        journal = open_journal(journal_name, config)
+    except KeyboardInterrupt:
+        print("[Interrupted while opening journal]", file=sys.stderr)
+        sys.exit(1)
+
     if mode_compose and not args.text:
         if not sys.stdin.isatty():
             # Piping data into jrnl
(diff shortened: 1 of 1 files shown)


jrnl/cli.py
```diff
@@ -345,6 +345,13 @@ def run(manual_args=None):
     else:
         _exit_multiline_code = "press Ctrl+D"
 
+    # This is where we fi±nally open the journal!
```
micahellison, line 348: This is really minor, but could please remove the typo in "finally"?

ollybritton:
Thanks for making my first contribution to open source a good experience!

wren:
Thanks for contributing! ❤️

micahellison:
Thanks for picking our project for your first contribution!

PR #958, opened 2020-05-19
Fix title splitting logic to account for both newlines and periods

### LINKED ISSUE #945: No way to have "." in the middle of a title
## Bug report

  - Jrnl version: (run `jrnl -v`): `jrnl version v2.4.1`
  - How you installed Jrnl: `pipx install jrnl`
  - Operating system [MacOS, Linux, Windows?]: macOS

There is no way to prevent jrnl from breaking on 

Run `jrnl` to create a new entry:

```
$ jrnl
```

In your editor, enter:

```
TIL that `. script` is equivalent to `source script`


```

Broken behavior: a newline is inserted after the first `.`.

```
$ jrnl -n 1
2020-05-09 11:51 TIL that `.
| script` is equivalent to `source script`
|
| But `. script` is more portable
```

`.` should only be considered the end of the title if it actually ends a sentence.

This is possibly a regression of https://github.com/jrnl-org/jrnl/issues/360 .

eshrh:
This PR changes the logic for separating the title from the rest of the entry. This is intended to fix #945. 

It's important to note that we can't just remove all sentence terminal checking, because then normal command-entry breaks. When a user enters an entry at the command line, they obviously can't naturally newline, so we still have to break the sentence by terminals. However, if there is a newline, then we can infer that that was where the user truly wanted the title to end because they were writing in a full text editor.

Because of this, it is impossible to *fully* fix #945, but it is possible to fix it if there is a newline character following the sentence.

Here are some cases:

User:
`jrnl I did it! I won the competition. I am so happy`
Result: 
>2020-05-19 16:15 I did it!
>| I won the competition

**This is no different from earlier**

However, if the user in a text editor writes:
`I did it! I won the competition.`
`I am so happy`
The result is now
>2020-05-19 16:17 I did it! I won the competition.
>| I am so happy

**This is different**

This fixes #945 without messing up anything else.

### Checklist

- [X] Tests pass. Your PR cannot be merged unless tests pass. --
  `poetry run behave`
- [X] The code passes linting via
  [black](https://black.readthedocs.io/en/stable/) (consistent code styling). --
  `poetry run black --check . --verbose --diff`
- [X] The code passes linting via [pyflakes](https://launchpad.net/pyflakes)
  (logically errors and unused imports). -- `poetry run pyflakes jrnl features`
- [X] Have you checked to ensure there aren't other open
  [Pull Requests](../pulls) for the same update/change?
- [X] Have you added an explanation of what your changes do and why you'd like
  us to include them?
- [X] Have you written new tests for your core changes, as applicable?

### DIFF
--- a/features/core.feature
+++ b/features/core.feature
@@ -26,6 +26,12 @@ Feature: Basic reading and writing to a journal
             | There is a blank line above this.
             """
 
+    Scenario: Multiline entry with punctuation
+        Given we use the config "basic.yaml"
+        When we run "jrnl This is. the title\\n This is the second line"
+        and we run "jrnl -n 1"
+        Then the output should contain "This is. the title"
+
     Scenario: Writing an entry from command line
         Given we use the config "basic.yaml"
         When we run "jrnl 23 july 2013: A cold and stormy day. I ate crisps on the sofa."
@@ -38,6 +44,8 @@ Feature: Basic reading and writing to a journal
         Given we use the config "editor.yaml"
         When we open the editor and enter nothing
         Then we should see the message "[Nothing saved to file]"
+
+
 
     Scenario: Writing an empty entry from the command line
         Given we use the config "basic.yaml"
--- a/features/core.feature
+++ b/features/core.feature
@@ -32,6 +32,12 @@ Feature: Basic reading and writing to a journal
         and we run "jrnl -n 1"
         Then the output should contain "This is. the title"
 
+    Scenario: Single line entry with punctuation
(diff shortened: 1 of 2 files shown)


eshrh:
Nevermind it does have to do with this pr, I'm working on figuring out why some of those checks failed.

wren:
I think this is a good compromise for the title splitting. Thank you!

Can you please add some tests for the changed functionality?

eshrh:
I added one test for multi-line entries with punctuation and one test for single-line entries with punctuation.

PR #961, opened 2020-05-24
Testing hanging Windows tests on Travis

micahellison:
I'm trying to figure out where a couple tests are causing the Travis Windows build to hang, and a PR build is the most convenient way to do that. I'll remove this PR when I find a solution or give up.

Current strategy:
* Remove the two skip_win directives
* Modify the affected tests so that they fail locally
* Add print tracers to see where it hangs in the Travis build log

### Checklist

- [ ] I have read the [contributing doc](https://github.com/jrnl-org/jrnl/blob/develop/CONTRIBUTING.md).
  for the same issue.
- [ ] I have you written new tests for these changes, as needed.
- [ ] All tests pass.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ install:
   - poetry run python --version
 
 script:
-  - poetry run behave --no-capture
+  - poetry run behave
 
 aliases:
   test_mac: &test_mac
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ install:
   - poetry run python --version
 
 script:
-  - poetry run behave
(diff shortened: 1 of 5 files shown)


micahellison:
I've narrowed down the freezing line to core.py:358:
```
        if not sys.stdin.isatty():
            # Piping data into jrnl
            raw = sys.stdin.read()
```
It looks like that sys.stdin.isatty() call is true when running behave locally but false in Travis Windows. So in Travis, it calls raw = sys.stdin.read() which is causing it to wait until it's finished reading input from the user that doesn't exist.

I'm spinning this off into a new issue #962 to keep track of it while closing this PR.

PR #963, opened 2020-05-26
Create directory if it doesn't exist

### LINKED ISSUE #893: New entry lost if directory missing/inaccessible
## Bug report

   - jrnl version v2.2
   - brew
   - MacOS 

   - configure journals.work to ~/jrnl/work.txt without creating ~/jrnl
   - run `jrnl work`
   - write a masterpiece
   - `:wq`
   - error stack trace below
   - work in progress gone

   - ibid

   - Use a non-temporary directory for work-in-progress so that failed updates are preserved
   - Create the directory when missing

```
  File "/usr/local/bin/jrnl", line 8, in <module>
    sys.exit(run())
  File "/usr/local/Cellar/jrnl/2.2_1/libexec/lib/python3.8/site-packages/jrnl/cli.py", line 362, in run
    journal = open_journal(journal_name, config)
  File "/usr/local/Cellar/jrnl/2.2_1/libexec/lib/python3.8/site-packages/jrnl/Journal.py", line 391, in open_journal
    return PlainJournal(name, **config).open()
  File "/usr/local/Cellar/jrnl/2.2_1/libexec/lib/python3.8/site-packages/jrnl/Journal.py", line 80, in open
    self.create_file(filename)
  File "/usr/local/Cellar/jrnl/2.2_1/libexec/lib/python3.8/site-packages/jrnl/Journal.py", line 105, in create_file
    with open(filename, "w"):
FileNotFoundError: [Errno 2] No such file or directory: '/Users/cbreym/jrnl/work.txt'
```

eshrh:
This is intended to close #893.
It creates the directory if it does not exist. 

If the directory doesn't exist it now prints:
```
[Directory <directory> created]
[Journal <journal>  created at <filename>]
[Entry added to <journal> journal]
```
The same thing happens right before asking for a password for EncryptedJournals.

### Checklist
  for the same issue.
- [X] I have you written new tests for these changes, as needed.
- [X] All tests pass.

I can't really make tests for this because it requires a directory that does not exist, and creates one if it doesn't; so the test would fail once and then pass every time after that, which obviously isn't very good.

### DIFF
--- a/jrnl/EncryptedJournal.py
+++ b/jrnl/EncryptedJournal.py
@@ -44,7 +44,7 @@ class EncryptedJournal(Journal):
         if not os.path.exists(filename):
             if not os.path.isdir(dirname):
                 os.makedirs(dirname)
-                print(f"[Directory {dirname} created]",file=sys.stderr)
+                print(f"[Directory {dirname} created]", file=sys.stderr)
             self.create_file(filename)
             self.password = util.create_password(self.name)
             print(
--- a/jrnl/EncryptedJournal.py
+++ b/jrnl/EncryptedJournal.py
@@ -43,8 +43,8 @@ class EncryptedJournal(Journal):
         dirname = os.path.dirname(filename)
         if not os.path.exists(filename):
             if not os.path.isdir(dirname):
-                os.mkdir(dirname)
-                print(f"[Directory {dirname} created]")
+                os.makedirs(dirname)
+                print(f"[Directory {dirname} created]",file=sys.stderr)
             self.create_file(filename)
             self.password = util.create_password(self.name)
             print(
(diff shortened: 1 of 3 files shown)


jrnl/EncryptedJournal.py
```diff
@@ -40,8 +40,11 @@ def open(self, filename=None):
         """Opens the journal file defined in the config and parses it into a list of Entries.
         Entries have the form (date, title, body)."""
         filename = filename or self.config["journal"]
-
+        dirname = os.path.dirname(filename)
         if not os.path.exists(filename):
+            if not os.path.isdir(dirname):
+                os.mkdir(dirname)
```
wren, line 46: Instead of `os.mkdir`, please us `os.makedirs` in case there are nested directories that need to be created.

jrnl/EncryptedJournal.py
```diff
@@ -40,8 +40,11 @@ def open(self, filename=None):
...
+                print(f"[Directory {dirname} created]")
```
wren, line 47: We usually print messages like this to `stderr` ([example](https://github.com/jrnl-org/jrnl/blob/develop/jrnl/util.py#L53)). That keeps these messages from accidentally ending up in someone's journal if they are piping the output.

eshrh:
@wren done.

PR #964, opened 2020-05-26
Fix set_keychain errors

### LINKED ISSUE #851: Confusing error message when user does not have a keyring backend installed
## Bug report

  - Jrnl version: (run `jrnl -v`) 2.2
  - How you installed Jrnl pip3 install jrnl
  - Operating system [MacOS, Linux, Windows?] Ubuntu 18.04

An error is shown. Here is the trace. 

```
-> jrnl
[Compose Entry; press Ctrl+D to finish writing]

holaEnter password for new journal: 
Enter password again: 
Passwords did not match, please try again
Enter password for new journal: 
Enter password again: 
Do you want to store the password in your keychain? [Y/n] n
  File "/home/jonathan/.local/lib/python3.6/site-packages/keyring/backend.py", line 203, in _load_plugins
    init_func = ep.load()
  File "/home/jonathan/.local/lib/python3.6/site-packages/importlib_metadata/__init__.py", line 94, in load
    module = import_module(match.group('module'))
  File "/usr/lib/python3.6/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 994, in _gcd_import
  File "<frozen importlib._bootstrap>", line 971, in _find_and_load
  File "<frozen importlib._bootstrap>", line 955, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 665, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 678, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/usr/lib/python3/dist-packages/keyrings/alt/Windows.py", line 9, in <module>
    from . import file_base
  File "/usr/lib/python3/dist-packages/keyrings/alt/file_base.py", line 13, in <module>
    from keyring.util.escape import escape as escape_for_ini
ModuleNotFoundError: No module named 'keyring.util.escape'

During handling of the above exception, another exception occurred:

  File "/home/jonathan/.local/bin/jrnl", line 11, in <module>
    sys.exit(run())
  File "/home/jonathan/.local/lib/python3.6/site-packages/jrnl/cli.py", line 362, in run
    journal = open_journal(journal_name, config)
  File "/home/jonathan/.local/lib/python3.6/site-packages/jrnl/Journal.py", line 397, in open_journal
    return EncryptedJournal.EncryptedJournal(name, **config).open()
  File "/home/jonathan/.local/lib/python3.6/site-packages/jrnl/EncryptedJournal.py", line 46, in open
    self.password = util.create_password(self.name)
  File "/home/jonathan/.local/lib/python3.6/site-packages/jrnl/util.py", line 62, in create_password
    set_keychain(journal_name, None)
  File "/home/jonathan/.local/lib/python3.6/site-packages/jrnl/util.py", line 101, in set_keychain
    import keyring
  File "/home/jonathan/.local/lib/python3.6/site-packages/keyring/__init__.py", line 1, in <module>
    from .core import (
  File "/home/jonathan/.local/lib/python3.6/site-packages/keyring/core.py", line 192, in <module>
    init_backend()
  File "/home/jonathan/.local/lib/python3.6/site-packages/keyring/core.py", line 96, in init_backend
    filter(limit, backend.get_all_keyring()),
  File "/home/jonathan/.local/lib/python3.6/site-packages/keyring/util/__init__.py", line 22, in wrapper
    func.always_returns = func(*args, **kwargs)
  File "/home/jonathan/.local/lib/python3.6/site-packages/keyring/backend.py", line 216, in get_all_keyring
    _load_plugins()
  File "/home/jonathan/.local/lib/python3.6/site-packages/keyring/backend.py", line 207, in _load_plugins
    log.exception("Error initializing plugin %s." % ep)
TypeError: not all arguments converted during string formatting
```

- pip3 install jrnl
- Edit config file and set a new jrnl file location and set encryption to true. 
- Typed jrnl and was asked for entry and new password. 
- Asked to store in keychain: if yes or no still the same error appears. 

I wouold expect to store my entry. If encryption is set to false it works. 

eshrh:
This is intended to close #851.

There are two problems with the current keyring system:
1. The program tries to delete a password even if a password hasn't been set yet
2. Disgracefully crashes when there's no keyring backend.

This pull request solves both of those problems. 
Do note that that the new mechanism **does not** exit if a keyring backend is not found. This is because for whatever reason, the password seems to get corrupted/missing after the program crashes due to `keyring.errors.KeyringError`. 

The problem with not exiting is that if the editor is queued to pop up, the user never had an opportunity to see that their password didn't get saved. I would have just asked the user whether they wanted to continue or not, but this leads to the problem of what happens if they Ctrl-C out of that prompt. It also leads to some probably nasty scenario where you'd have to set a flag that we need to quit if the user chooses no, and then somehow finish the program without adding any text. I thought this was far too messy to be worth it.

This is very easy to test for me. The default poetry shell does not have a keyring. Perhaps we should include the `keyrings.cryptfile` package as a dev dependency?
### Checklist

  for the same issue.
- [X] All tests pass.

### DIFF
--- a/jrnl/util.py
+++ b/jrnl/util.py
@@ -60,8 +60,6 @@ def create_password(
 
     if yesno("Do you want to store the password in your keychain?", default=True):
         set_keychain(journal_name, pw)
-    else:
-        pass
     return pw
 
 
@@ -111,7 +109,7 @@ def set_keychain(journal_name, password):
             keyring.set_password("jrnl", journal_name, password)
         except keyring.errors.NoKeyringError:
             print(
-                "Keyring backend not found. Please install one of the supported backends by visiting: https://pypi.org/project/keyring/"
+                "Keyring backend not found. Please install one of the supported backends by visiting: https://pypi.org/project/keyring/",file=sys.stderr
             )
 
 
--- a/jrnl/util.py
+++ b/jrnl/util.py
@@ -110,7 +110,9 @@ def set_keychain(journal_name, password):
         try:
             keyring.set_password("jrnl", journal_name, password)
         except keyring.errors.NoKeyringError:
-            print("Keyring backend not found. Please install one of the supported backends by visiting: https://pypi.org/project/keyring/")
+            print(
+                "Keyring backend not found. Please install one of the supported backends by visiting: https://pypi.org/project/keyring/"
+            )
 
 
 def yesno(prompt, default=True):
--- a/jrnl/util.py
+++ b/jrnl/util.py
@@ -109,7 +109,8 @@ def set_keychain(journal_name, password):
             keyring.set_password("jrnl", journal_name, password)
         except keyring.errors.NoKeyringError:
             print(
-                "Keyring backend not found. Please install one of the supported backends by visiting: https://pypi.org/project/keyring/",file=sys.stderr
+                "Keyring backend not found. Please install one of the supported backends by visiting: https://pypi.org/project/keyring/",
+                file=sys.stderr,
             )
 
 
--- a/jrnl/util.py
+++ b/jrnl/util.py
@@ -61,8 +61,7 @@ def create_password(
     if yesno("Do you want to store the password in your keychain?", default=True):
         set_keychain(journal_name, pw)
     else:
-        set_keychain(journal_name, None)
-
+        pass
     return pw
 
 
@@ -108,7 +107,10 @@ def set_keychain(journal_name, password):
         except keyring.errors.PasswordDeleteError:
             pass
     else:
-        keyring.set_password("jrnl", journal_name, password)
+        try:
+            keyring.set_password("jrnl", journal_name, password)
+        except keyring.errors.NoKeyringError:
+            print("Keyring backend not found. Please install one of the supported backends by visiting: https://pypi.org/project/keyring/")
 
 
 def yesno(prompt, default=True):
--- a/features/encryption.feature
+++ b/features/encryption.feature
@@ -55,3 +55,27 @@
             Then the config for journal "simple" should have "encrypt" set to "bool:True"
             When we run "jrnl simple -n 1"
             Then the output should contain "2013-06-10 15:40 Life is good"
+
+        Scenario: Encrypt journal with no keyring backend and do not store in keyring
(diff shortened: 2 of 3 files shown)


jrnl/util.py
```diff
@@ -108,7 +107,12 @@ def set_keychain(journal_name, password):
         except keyring.errors.PasswordDeleteError:
             pass
     else:
-        keyring.set_password("jrnl", journal_name, password)
+        try:
+            keyring.set_password("jrnl", journal_name, password)
+        except keyring.errors.NoKeyringError:
+            print(
```
wren, line 113: Please print errors to `stderr` ([example](https://github.com/jrnl-org/jrnl/blob/develop/jrnl/util.py#L53)).

jrnl/util.py
```diff
@@ -61,8 +61,7 @@ def create_password(
     if yesno("Do you want to store the password in your keychain?", default=True):
         set_keychain(journal_name, pw)
     else:
-        set_keychain(journal_name, None)
-
+        pass
```
wren, line 64:
```
else:
  pass
```
Can be more succinctly written by removing the `else` entirely.

wren:
I'm not as familiar with the keyring functionality, so I'll have to go through this with some more detail (I'm not sure I know why we are attempting to delete a password if one hasn't been set). 

As for adding a keyring dev dependency, I'm open to that (it seems like a good improvement for testing, too). Do you have suggestions for what the best keyring would be for devs and testing?

eshrh:
@wren 
I have an idea as to why it might have been written this way. `set_keychain(keychain,None)` is also used in `decrypt_content`. That function first tries the password from the keychain, and if it doesn't work, deletes the key. I think this makes sense. However, it used twice for two different purposes.

The author must have realized this problem, and caught the exception if there was no password(namely `PasswordDeleteError`). They didn't realize that if they keychain doesn't exist at all there will be an entirely different error. 

I figured out a way to test with an environment with no keychain. I just made my own keyring backend and made is raise an exception every time the program tries to do anything with it.

PR #984, opened 2020-06-13
Add --diagnostic argument

### LINKED ISSUE #727: Create a diagnostic command line argument
## Feature Request

Rather than asking users to check their python version, jrnl version, operating system, etc., it would be nice to have a general diagnostic dump that outputs all these values. Maybe it could also output certain non-sensitive values from the configuration (e.g. time format).

* **Please provide examples of the usage**
Once finished, we can edit the issue template that was recently made for bugs, and ask users to run this command.

micahellison:
Adds a `--diagnostic` argument that outputs the jrnl version, Python version, and OS version.

After this is released, there should be a follow-up PR to change the bug issue template to mention this argument.

Fixes #727.

### Checklist

  for the same issue.
- [X] All tests pass.

### DIFF
--- a/features/core.feature
+++ b/features/core.feature
@@ -130,3 +130,8 @@ Feature: Basic reading and writing to a journal
         and we change directory to "features"
         and we run "jrnl -n 1"
         Then the output should contain "hello world"
+
+    Scenario: --diagnostic runs without exceptions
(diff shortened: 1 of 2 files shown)


wren (approved):
💯

PR #987, opened 2020-06-20
Add unit testing via pytest

### LINKED ISSUE #986: Add unit tests
We have behavior-driven tests through the behave framework, but we're running up against some limitations. It's slow and it's not very granular, which makes refactoring cumbersome and even dangerous.

A good way around this problem is to add unit tests. Looks like [pytest ](https://docs.pytest.org/en/stable/) is the best candidate for a test runner, and I'm putting up a PR to set that up.

Between poetry and our makefile, this should be a relatively painless addition for contributors. You'll just run `make test` as you normally do after changing some code.

micahellison:
Includes a couple sample tests, a note in CONTRIBUTING, and a step in the makefile under `make test`. I'm including the step before behave, since pytest is much faster and more granular.

Fixes #986.

### Checklist

  for the same issue.
- [X] All tests pass.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,7 @@ install:
   - poetry run python --version
 
 script:
+  - poetry run pytest
   - poetry run behave
 
 aliases:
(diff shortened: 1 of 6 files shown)


wren (approved):
🎩

Issue #737, opened 2019-11-12
Pull functionality out of util.py


micahellison:
## Feature Request

* **What is the motivation / use case for changing the behavior?**
Currently, util.py is a grab-bag of different functionality related to encryption, text coloring, configuration, and retrieving text from an editor. It also contains the colorama statement that makes text coloration possible in Windows terminals.

It would be nice to have each method moved to an appropriate class and ideally, remove util.py entirely so we can adhere to the single responsibility principle.

alichtman:
Agreed. `util.py` is getting a little too big to work in easily.

wren:
@alichtman The current PR goes very closely with your recommendations here. We ended up putting encryption stuff into `EncryptedJournal.py` because those functions need to be folded into that class (I think this will let us more easily support different encryption methods like pgp without reaching up into the control flow).

PR #991, opened 2020-07-04
Partial refactor of cli.py (mainly help screen and arg parsing)

### LINKED ISSUE #344: Clarify usage output between export and reading sections

### LINKED ISSUE #374: Feature request: support -not for @tag in the command line
Much like the -and switch, it would be really great if -not could be supported as well as the -and (even if you could only use one or the other) to provide some simple boolean querying of the entries.

### LINKED ISSUE #520: -and parameter seems to only work for the default journal
I have two journals setup like this:
```
{
  "journals": {
    "default": "~/log.txt",
    "work": "~/work.txt"
  }
}
```
If I want to view all entries with the tags `@x` and `@y` for the default log I write this:
```
jrnl -and @x @y
```
which nicely lists the correct entries. If I want to achieve the same for my `work` log I write this: ```jrnl work -and @x @y``` and this error gets thrown:
```
usage: jrnl [-h] [-v] [-ls] [-d] [-from DATE] [-until DATE] [-on DATE] [-and]
            [-starred] [-n [N]] [--short] [--tags] [--export TYPE] [-o OUTPUT]
            [--encrypt [FILENAME]] [--decrypt [FILENAME]] [--edit]
            [[...]]
jrnl: error: unrecognized arguments: @x @y
```

Am I misunderstanding how to use the `-and` parameter for non-default journals, or is this a bug?

### LINKED ISSUE #814: --tags vs --export tags
## Feature Request

`--tags` will print tags
`--export tags` will also print tags

`--export tags` will make use `-o` output file, if specified
`--tags` does not make use of the `-o` option

**Why must `-o` be restricted to `--export`?**

I find `jrnl --tags -o file.txt` much easier to type than `jrnl --export tags -o file.txt`.

### LINKED ISSUE #866: Implement arg parsing library
## Feature Request

There are some features we'd like to implement that would be made easier by using an arg parsing library (like Click). There are likely going to be some implementation challenges, since our interface has some magic built-in, but the goal of this interface is to have near identical

The current arg parsing has some troubles consistently parsing arguments in different order, optional journal names, international timestamps, etc. It needs to be redone, and this would be a good opportunity to use a library that gives us some features baked in (like cli completion). 

* **Please provide examples of the usage**
The interface should looks exactly the same as now (with only minor changes as needed for bugfixing.

### LINKED ISSUE #875: Listing jrnl entries by tag for non default journal seem to not work as expected.
## Bug report

  - Jrnl version: (run `jrnl -v`) v2.2
  - How you installed Jrnl pip3
  - Operating system [MacOS, Linux, Windows?] Debian

with default journal tags work fine 
 jrnl -s @sometag 

jrnl other_journal -s @sometag 

This does not work

listing entries with a journal tag for a non default journal will still return items.

wren:
Fixes #344
Fixes #374
Fixes #520
Fixes #814
Fixes #866
Fixes #875

This is a partial refactor (another PR to come soon) cli.py, specifically focused on the
arg parsing code, and the `run` function. For any args that were changed, I also added
the old flag as a hidden option, so existing aliases and scripts will still work behind
the scenes.

Most of the changes are transparent to users, but will make life easier for
contributors.

Changes that affect users:
- Fixed some bugs (see above)
- Reorganized the help screen (might be too big now, so maybe it's the new man page?)
- Changed the name of some flags (old flags still work)
- Some flags operate slightly differently. Instead of taking an optional string
  immediately after the argument (e.g. `--encrypt filename.txt`), they now take that
  optional string from an option (e.g. `--encrypt --file filename.txt`). Please note
  that these additional flags are completely optional, and not giving these options has
  the same default behavior as before. The affected flags are:
  - `--import` (`--import TYPE` is now `--import --format TYPE`)
  - `--encrypt` (`--encrypt FILENAME` is now `--encrypt --file FILENAME`)
  - `--decrypt` (`--decrypt FILENAME` is now `--decrypt --file FILENAME`)

Changes that don't affect users, but do affect contributors:
- Standalone commands (like `--list` and `--encrypt`) are now housed in an argparse
  option, and are executed at the correct time. This makes it easier to add more
  commands later (commands.py), and reduces the amount of overhead tracking all the
  various command line options that a user can specify. You can now add a standalone
  command by simply writing a function (commands.py), and registering the function
  in the arg parser (arg_parser.py). That's it. Then `jrnl --newcommand` will work as
  expected.
- The multiple modes of jrnl are disappearing, and we'll be left only with searching,
  and writing. This PR starts that work (although it's not complete). After searching,
  you have the option to do various things with the search results (e.g. put them into
  editor, delete them, display them in alternate format, etc). That's already how jrnl
  works for the users, but the code will now better follow that flow.
- In this PR, more functions have gone into util.py (I know, I'm sorry). The intent is
  not for them to stay there (we already have an issue to clean that file up #737). The
  functions here will be split into smaller files when that issue is dealt with.
- More tests! Mainly starting to flesh out the pytest suite, but also adding some more
  behave tests.
- Our behave now supports simulating piping data into jrnl. This was added for testing
  importing, but can be used elsewhere as needed.

### Checklist

  for the same issue.
- [ ] All tests pass.


micahellison (approved):
Great tests, nice refactoring, and it's a delight to see these argument bugs squashed. This will be nice to build off of. Thanks for your work on this.

PR #994, opened 2020-07-04
Display warning if Python version is less than 3.7

### LINKED ISSUE #992: Deprecate Python 3.6
We're refactoring the argument parsing code and are using some functionality that isn't supported until Python 3.7 (``ArgumentParser.parse_intermixed_args``). It will be a while before this code goes live, but in the meantime, we should add a deprecation notice if users are running Python 3.6.

I don't think this should significantly inconvenience users. 3.6 is about four years old and 3.9 should come out in a few months. Homebrew uses 3.8 with jrnl, and Windows users don't have a default version to worry about.

micahellison:
Fixes #992. Shows a warning in STDERR if the user's Python version is less than 3.7:
```
WARNING: Python versions below 3.7 are not supported by jrnl
```

### Checklist

  for the same issue.
- [ ] All tests pass.

### DIFF
--- a/jrnl/cli.py
+++ b/jrnl/cli.py
@@ -314,8 +314,13 @@ def configure_logger(debug=False):
 
 
 def run(manual_args=None):
-    if packaging.version.parse(platform.python_version()) < packaging.version.parse('3.7'):
-        print("WARNING: Python versions below 3.7 are not supported by jrnl\n", file=sys.stderr)
+    if packaging.version.parse(platform.python_version()) < packaging.version.parse(
+        "3.7"
+    ):
+        print(
+            "WARNING: Python versions below 3.7 are not supported by jrnl\n",
+            file=sys.stderr,
+        )
 
     if manual_args is None:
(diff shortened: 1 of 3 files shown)


jrnl/cli.py
```diff
@@ -313,6 +314,14 @@ def configure_logger(debug=False):
 
 
 def run(manual_args=None):
+    if packaging.version.parse(platform.python_version()) < packaging.version.parse(
+        "3.7"
+    ):
+        print(
+            "WARNING: Python versions below 3.7 are not supported by jrnl\n",
```
wren: Can we make this a bit friendlier, and add color?

What do you think of something like this?

```
f"""{WARNING_COLOR}
WARNING: Python versions below 3.7 will no longer be supported as of jrnl v2.5 (the next
release). You are currently are on Python {platform.python_version()}. Please update to
Python 3.7 (or higher) soon.
{RESET_COLOR}"""
```

wren (approved):
🎩

PR #1001, opened 2020-07-11
Allow editing of DayOne entries

### LINKED ISSUE #955: [DayOne] Editing an entry shows AttributeError
- Jrnl version: **2.4.2**
  - How you installed Jrnl: **pip3 install jrnl**
  - Operating system: **MacOS**

I’m using [DayOne Classic integration](https://github.com/jrnl-org/jrnl/blob/f0f26a18eb3304aff5ea53e669802a18a2f85e5a/advanced/index.html#L194) with iCloud synchronization.

Trying to edit an existing entry. No modifications made to the text itself. Tried with multiple editors (`vim`, `sublime`). Always ending up with error message **AttributeError: can't set attribute**.

```
fbreunig@nighttrain:~ $ jrnl --version
jrnl version v2.4.2
fbreunig@nighttrain:~ $ echo test | jrnl
fbreunig@nighttrain:~ $ jrnl -n 1 --edit
  File "/usr/local/bin/jrnl", line 8, in <module>
    sys.exit(run())
  File "/usr/local/lib/python3.7/site-packages/jrnl/cli.py", line 460, in run
    journal.parse_editable_str(edited)
  File "/usr/local/lib/python3.7/site-packages/jrnl/DayOneJournal.py", line 150, in parse_editable_str
    current_entry.title = line[len(date_blob) - 1 :]
AttributeError: can't set attribute
```

Note: If I delete the _entire_ text during editing, the entry is being deleted as expected:
```
[Nothing saved to file]
[1 entry deleted]
```

wren:
Fixes #955

DayOneJournal previously reimplemented Journal._parse inside of DayOneJournal.parse_editable_string, and in doing so caused issues between the two classes (the body was being reparsed for a title, overwriting the title, and essentially dropping the first line of each edited entry). This commit fixes the issue by moving the UUID to be in the body of the entry, rather than above it. Doing this allows us to use Journal._parse like normal (it still finds the correct boundaries between entries), and then DayOneJournal parses the UUID afterward.

Co-authored-by: [MinchinWeb](https://github.com/minchinweb)
Co-authored-by: [Micah Jerome Ellison](https://github.com/micahellison)

### Checklist

  for the same issue.
- [x] All tests pass.

### DIFF
--- a/jrnl/DayOneJournal.py
+++ b/jrnl/DayOneJournal.py
@@ -16,7 +16,6 @@ import pytz
 import tzlocal
 
 from . import __title__, __version__, Entry, Journal
-from . import time as jrnl_time
 
 
 class DayOne(Journal.Journal):
@@ -179,7 +178,26 @@ class DayOne(Journal.Journal):
     def editable_str(self):
         """Turns the journal into a string of entries that can be edited
         manually and later be parsed with eslf.parse_editable_str."""
-        return "\n".join([f"# {e.uuid}\n{str(e)}" for e in self.entries])
+        return "\n".join([f"{str(e)}\n# {e.uuid}\n" for e in self.entries])
+
+    def _update_old_entry(self, entry, new_entry):
+        for attr in ("title", "body", "date"):
+            old_attr = getattr(entry, attr)
+            new_attr = getattr(new_entry, attr)
+            if old_attr != new_attr:
+                entry.modified = True
+                setattr(entry, attr, new_attr)
+
+    def _get_and_remove_uuid_from_entry(self, entry):
+        uuid_regex = "^ *?# ([a-zA-Z0-9]+) *?$"
+        m = re.search(uuid_regex, entry.body, re.MULTILINE)
+        entry.uuid = m.group(1) if m else None
+
+        # remove the uuid from the body
+        entry.body = re.sub(uuid_regex, "", entry.body, flags=re.MULTILINE, count=1)
+        entry.body = entry.body.rstrip()
+
+        return entry
 
     def parse_editable_str(self, edited):
         """Parses the output of self.editable_str and updates its entries."""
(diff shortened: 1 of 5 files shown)


wren:
Well, that took longer than expected, but I think this is all working now.

@MinchinWeb @micahellison Would y'all mind taking a look over this?

minchinweb:
I haven't had the chance to test drive this yet, but it we're using a regex match:

- UUID's should all be the same length. Should the regex enforce that?
- should we be more explicit with our inserted text, perhaps by prefacing it with "UUID:"?
- "#" is used for Level 1 headings in Markdown, but this is now at the bottom of the entry. Would it make sense to to move to an HTML comment style? Like `<!-- 3547D0580346804689 -->`
    - I realize not everyone writes their entries in Markdown, but I do. If for some reason the comment got included in the output, it would automatically just not be rendered in HTML, which I think it a good default.
    - if you combine these two, you'd get `<!-- UUID: 3547D0580346804689 -->`
- should we think about extend-ability of this for other DayOne metadata? That doesn't have to be dealt with now, but (for example) currently you can't edit the tags on an entry in a straightforward, deterministic manner. If we went with the HTML comment style, it could be extended for tags, and potentially other metadata. So add another line like: `<!-- tags: test_tag, awesome project -->`

wren:
@MinchinWeb I think those are all very good questions, and we should definitely discuss if and how to handle extended metadata on entries for certain journal types, but I would consider that out of scope for issue #955.

I was also very tempted to change the `#` to something more friendly, but I decided against it in order to keep the format as close to its current look. That way, we can take some time to plan any upcoming changes in a different issue.

jrnl/DayOneJournal.py
```diff
@@ -179,87 +178,45 @@ def write(self):
...
-            ]
-            # tags in entry body
-            if matched_entries:
-                # This entry is an existing entry
-                match = matched_entries[0]
-
-                # merge existing tags with tags pulled from the entry body
-                entry.tags = list(set(entry.tags + match.tags))
-
-                # extended Dayone metadata
```
minchinweb, line 237: How is this extended metadata being preserved? or is it being dropped when the entry is edited?

jrnl/Entry.py
```diff
@@ -37,18 +39,30 @@ def title(self):
             self._parse_text()
         return self._title
 
+    @title.setter
```
minchinweb, line 42: These setters I don't think are being used by your solution. Would it make sense to drop them?

minchinweb:
Simple local testing seems to show that it works! Extended metadata stays in place. The `behave` test suite passes locally.

In regards to this PR, should we remove the setter code I added, as it's not being used?

In regards to future work:

- we should look at expanding this to support editing entry tags
- we should look at the format of the UUID within the editing output
- the "Device Agent" is currently being set to a empty string, which is less than ideal (this is not new)

I'm excited to see this merged. Thank for everyone's help!

LGTM!

jrnl/DayOneJournal.py
micahellison: It is being preserved: it is read when the journal is opened, it remains in memory as the entry is being edited, then it's saved back. And it still creates the metadata when writing a new entry.

jrnl/Entry.py
micahellison: I think it's good to leave them in. It sends a nudge to future contributors that they should use entry's accessors rather than underscore-prefixed internals.

micahellison (approved):
Looks good! Nice collaboration!

PR #1002, opened 2020-07-13
Fix readme link to submit an issue

wren:
The old link pointed at the page for a blank issue (which tells the user to stop and go to the right page). This updates the link to choose a type of issue.


micahellison (approved):
🚂 Always good to choo-choose that issue type

PR #1004, opened 2020-07-18
Fix readthedocs.io build process

### LINKED ISSUE #977: Update or remove docs on readthedocs
## Feature Request
- Update or remove the docs at https://jrnl.readthedocs.io/en/latest/advanced.html

### Use Case/Motivation
I spent a while trying to figure out why stuff I add to `~/.jrnl_config` was being ignored. These docs are marked 'latest' and still talk about this instead of the new .yaml file in v2. It's not obvious that these docs are for an older version if you land on that page directly.

If the new docs are hosted on jrnl.sh directly, the readthedocs ones should maybe redirect there? Or be updated if they are still active too.

wren:
Fixes #977

So, fun story! It turns out that readthedocs is using an old version of mkdocs for our builds. (Apparently)[  https://github.com/readthedocs/readthedocs.org/issues/5332], this is because we're an old enough project that readthedocs gives us a different default than what their documentation says we should be getting (which doesn't work anymore since we've been making upgrades). And apparently the only way to override this is by adding a requirements.txt file.

We don't use requirements.txt (because of those same upgrades), but it looks like we're stuck if we want to continue supporting readthedocs (unless we want to delete the project and start it again).

So, this PR adds a readthedocs config file (`readthedocs.yml`), and `docs/requirements.txt` to get us out of the 7th circle of build hell we were thrown into.

I've already run the build process on a test branch, and here are the results: https://jrnl.readthedocs.io/en/readthedocs-test/

I think we'll need to follow this up with a few styling fixes for the theme, but this will at least get the docs building again.

### Checklist

  for the same issue.
- [x] All tests pass.


wren:
Also, bonus! I enabled readthedocs builds on PRs, so we'll get a good view of changes to docs on PRs now.

wren:
This is close, but doesn't quite work without breaking jrnl.sh. 

Either
1. We can include an index.md for readthe docs (which doesn't seem to recognize the current index.html for some reason), or
2. We leave out index.md and read the docs gives us a generic "homepage" which basically just says "you should include an index.md." 😞

So, this needs a way to include an index for readthedocs that doesn't break our current index.html on jrnl.sh.

micahellison (approved):
👓

minchinweb:
Is `jrnl.sh` hosted on Read the Docs? Or is Read the Docs a mirror of `jrnl.sh`? I don't quite understand the relationship here...

wren:
@MinchinWeb jrnl.sh is hosted on Github Pages. Read the Docs is a documentation-gathering site that's been around for ages, and is popular among the open source community. The docs for jrnl v1.9.8 were still up there, and it looks like they were still getting a decent amount of traffic, which is why we're going through the trouble of updating it.

Honestly, if/when it's fixed up, depending on how it goes, I would consider moving jrnl.sh over to there. They have some good build options, and they get a decent amount of organic traffic anyway. They also have some built-in, privacy-respecting analytics that could be useful to bettering the docs.

PR #1011, opened 2020-07-29
Add missing dependency (packaging)

### LINKED ISSUE #1010: First run of jrnl, getting ModuleNotFoundError
### Environment
  - Jrnl version: (not sure. jrnl -v produces an error)
  - Install method: pipx
  - OS: Linux, Pop_OS 20.04

I am trying to run jrnl for the first time.

### What have you tried?
My system came with Python 2.7.18rc1.
I installed (from source) Python 3.8.5 

Even after installing the latest python, I still get the same ModuleNotFoundError

Here is the output of the jrnl command.
```
$ jrnl
  File "./.local/bin/jrnl", line 5, in <module>
    from jrnl.cli import run
  File "/home/tj/.local/pipx/venvs/jrnl/lib/python3.8/site-packages/jrnl/cli.py", line 24, in <module>
    import packaging.version
ModuleNotFoundError: No module named 'packaging'

```
So, I then tried:

```
pipx install packaging

No apps associated with package packaging. If you are attempting to install a library, pipx should not be used. Consider using pip or a similar tool instead.

```

Not sure what to do. I want to use jrnl, but I think something with Python is not working correctly.

wren:
Fixes #1010

`packaging` was missing from our latest release (only included in our dev dependencies). Our tests missed it because they all have dev deps installed. This PR adds the missing dep, and we'll deal with the test suite separately.

### Checklist

  for the same issue.
- [ ] All tests pass.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -314,7 +314,7 @@ tgrep = ["pyparsing"]
 twitter = ["twython"]
 
 [[package]]
-category = "dev"
+category = "main"
 description = "Core utilities for Python packages"
 name = "packaging"
 optional = false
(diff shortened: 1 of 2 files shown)


micahellison (approved):
📦🎉

PR #1012, opened 2020-07-29
Update dependency (cryptography)

### LINKED ISSUE #1009: Archlinux packaging issues
## Bug Report

Thanks for your time, please keep reading before closing as OS packaging problem, just trying to figure out a solution.

### Environment
  - Jrnl version: 2.4.2
  - Install method: aur / pacman (archlinux package manager
  - OS Linux

### Current Behavior
System upgrade fails due to jrnl dependency being listed as python-cryptography<3.0

### Expected Behavior
No problems upgrading, since (AFAIK) there is no (relevant) breaking changes in this library. Tested (loosely) and it worked.

### Repro Steps
* Install jrnl using arch package manager
* Try to upgrade (e.g. pacman -Syu)

### Other Information
The dependency is listed as "cryptography = "^2.7" in pyproject.toml, which seems suspiciously like something poetry would generate. Now, before anything, if there is a reason to limit `cryptography` version to <3.0 as this would, please close this issue: the solution would be to change OS packaging to use pipx, as discussed by the maintainer [here](https://aur.archlinux.org/packages/jrnl/).

Otherwise, just relaxing this constraint would help us poor arch users. I saw the discussion about loosing up dependency restrictions in some other arch issue, I agree it should be made with caution. On the other hand, when dealing with packages such as this one, the arbitrary constraint will keep bug fixes and security enhancements out of the build, and, well, one can go both ways. Poetry's behavior is understandable, and in this case there are many breaking changes in `cryptography 3.0`. Version number default semantics justifies it. So, it is a matter of opinion, but an upper limit seems more reasonable when there is a known problem and not as a prophylactic.

I am just giving some rationalizations here. The real reason is I don't like this behavior, and it has been a pain more than a solution. Feel free to ignore. Also, there is a non trivial chance I am missing something here.

### Workaround
Uninstall jrnl and reinstall using pipx. Yes, this is the recommended way by the docs. See, I like python, but its dependency hell makes me think it should be named Nachash. :grin:

wren:
Depends on #1011
Fixes #1009

Cryptography updated to 3.0 recently. This updates jrnl to use the new major version.

### Checklist

  for the same issue.
- [ ] All tests pass.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -15,7 +15,7 @@ description = "A small Python module for determining appropriate platform-specif
 name = "appdirs"
 optional = false
 python-versions = "*"
-version = "1.4.3"
+version = "1.4.4"
 
 [[package]]
 category = "main"
@@ -91,7 +91,7 @@ description = "Foreign Function Interface for Python calling C code."
 name = "cffi"
 optional = false
 python-versions = "*"
-version = "1.14.0"
+version = "1.14.1"
 
 [package.dependencies]
 pycparser = "*"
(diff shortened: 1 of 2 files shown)


wren:
Included in #1016

Issue #1017, opened 2020-08-01
Add support for Python 3.9


micahellison:
## Feature Request

It will soon be time to officially support Python 3.9. The final beta came out recently and [the release candidate phase begins this month](https://pythoninsider.blogspot.com/2020/07/python-385-released-as-security-hotfix.html). It would be nice to support the official release in a timely manner, especially since we've just deprecated 3.6.

This change should involve:
* documentation
* Travis CI builds
* poetry
* Perhaps some other things (it would be worth searching the code base for "3.8")

wren:
We should probably add 3.9 tests to the suite sooner, and mark them as `allow_failures`.

PR #1018, opened 2020-08-02
Add support for Python 3.9 build testing

micahellison:
First step of #1017 - add build steps for Python 3.9 in Travis. Allowing failures for now, until 3.9 is stable.

### Checklist

- [n/a] I have tested this code locally.
  for the same issue.
- [n/a] All tests pass.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,12 +44,13 @@ aliases:
     os: windows
     language: shell
     env: &env_windows
-      PATH: /c/Python37:/c/Python37/Scripts:/c/Python38:/c/Python38/Scripts:$PATH
+      PATH: /c/Python37:/c/Python37/Scripts:/c/Python38:/c/Python38/Scripts:/c/Python39:/c/Python39/Scripts:$PATH
       PYTHONIOENCODING: UTF-8
     cache:
       directories:
         - /c/Python37
         - /c/Python38
+        - /c/Python39
         - /c/ProgramData/chocolatey/lib
         - /c/ProgramData/chocolatey/bin
         - /c/Users/travis/AppData/Local/pypoetry/Cache
--- a/.travis.yml
+++ b/.travis.yml
@@ -131,10 +131,11 @@ jobs:
       name: Python 3.9 on Windows
       before_install:
         - choco install python --pre
-        - C:\ProgramData\chocolatey\bin\python3 -m pip install --upgrade pip
+        - ls C:/ProgramData/chocolatey/bin/
+        - C:/ProgramData/chocolatey/bin/python -m pip install --upgrade pip
         - pip --version
         - pip install toml
-        - C:\ProgramData\chocolatey\bin\python3 .build/allow_all_python_version.py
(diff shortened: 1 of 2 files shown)


micahellison:
This still needs some work. Issues to figure out:

* `before_install` step
  * Failing on 3.9 Mac. Looks like the `sed` command behaves differently on MacOS, and I'm not sure how best to replace it with find-and-replace logic.
  * Failing on 3.9 Windows. It works fine on my local machine but maybe a different version of PowerShell is running on Travis.
  * Maybe we could do this more effectively with a Python command executed from the shell, and use the same command for all OSes.
* I have somehow managed to add the 3.9 Linux build to `allow_failures` but not the others. I don't know why. I want all of them to be allowed to fail for now.
* Some new test failures on the 3.9 Linux build.
  * Is `behave` just dying in the middle of tests?
  * Tests are failing due to the version string: "ERROR: Python version 3.9.0b5+ not supported." Is it the case that any beta Python build is not allowed to run jrnl now? And is that okay?

To anyone looking at this: feel free to offer thoughts and suggestions. I'll come back to this later this week.

wren:
Yeah, the sed command was a hack thrown in there when we didn't have to consider the other OSes. We should probably use the toml library we already load in our python dev deps.

micahellison:
I've resolved the `allow_failures` problem, and I've got it working on Mac and Linux (thanks for the help @wren) but I'm still struggling with Windows. It's installing 3.9.0b5 with `choco install python --pre` but I don't know where it's installing it, since there's also a Python 2.7 installation on this Windows machine, and Python isn't behaving quite like it does in the other Windows Python 3 builds.

I'll come back to this later but, as always, any help is appreciated.

micahellison:
Okay. I think this is about as good as I'm going to get this.

Windows is correctly using Python 3.9, but it can't install cryptography because there are no wheels available for Python 3.9. The alternative is building it from source, which feels like a bit much, and isn't very representative of Windows users anyway, as they are generally not going to take this approach.. Maybe it will work as we get closer to the release in October, or if not, we'll at least know about this issue when we cross that bridge.

.travis.yml
```diff
@@ -105,10 +108,42 @@ jobs:
...
+        - python .build/allow_all_python_version.py
+      env:
+        <<: *env_windows
+        JRNL_PYTHON_VERSION: 3.9.0b5
+
     # ... and beyond!
     - name: Python nightly on Linux
       before_install:
-        - sed -i 's/^python = ">=3\.6\.0.*"$/python = "*"/' pyproject.toml
+        - sed -i 's/^python = ">=3\.7\.0.*"$/python = "*"/' pyproject.toml
```
wren: Shouldn't this use your new script (` .build/allow_all_python_version.py`)?

wren (dismissed):
Looks great with one minor comment. I'll be happy to merge if you don't want to fix it, just let me know.

wren (approved):
🐆

PR #1027, opened 2020-08-16
Clean up help screen, get rid of util.py

### LINKED ISSUE #737: Pull functionality out of util.py

wren:
Fixes #737 Gets rid of util.py! 🎉 

This mostly just moves around code with the intent to group functions into easier-to-work-with files.

There should be no change in functionality, but will make it easier to add new features and find bugs.

There are still a few things that I would like to move, but I'll make separate issues for those later as needed.

### Checklist

  for the same issue.
- [ ] All tests pass.


wren:
Welp, this breaks on Windows for some reason.

micahellison (approved):

Issue #853, opened 2020-02-17
Issue with -and and @tags order  if using jrnl profiles


mandarup:
## Bug report

* **Please tell us about your environment:**
  - Jrnl version: (run `jrnl -v`)  v2.2
  - How you installed Jrnl  - pip install jrnl   in virtualenv
  - Operating system [MacOS, Linux, Windows?] : MacOS

* **What is the current behavior?**

If custom profiles are set in jrnl.yaml
```yaml
journals:
  work:
     encrypt: false
     journal: /path/to/work.txt
```

There are two issues.

1)  with tags after `-and` as specified in the documents, it throws error
```sh
$ jrnl work -n 5 -and @pineapple @lubricant
jrnl: error: unrecognized arguments:
```

Works as expected (&)  if specify `-and` flag is after the tags
```sh
$ jrnl  work @pineapple @lubricant -and
```

2) But doesn't work if `-n` flag is used before tags
```sh
$jrnl work -n 5  @pineapple
jrnl: error: unrecognized arguments:
```

Works if `-n` flag is moved to the end after tags
```sh
$jrnl  work @pineapple @lubricant -and -n 5
```

wren:
Yup!

This looks like more issues with our arg parsing. I've made an issue (#866) to deal with this once and for all. Please move discussion there.

PR #1035, opened 2020-08-24
Update usage.md

jimmyhess:
Move the `-and` option after two or more tags instead of in-between. Using this option in-between tags does not work if the jrnl file name is not `default`.

### Checklist

  for the same issue.
- [x] All tests pass.

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -133,14 +133,14 @@ all instances of the word "dogs."
 You can filter your journal entries by tag. For example,
 
 ```sh
-jrnl @pinkie @WorldDomination
+jrnl "@pinkie" "@WorldDomination"
 ```
 
 displays all entries in which either `@pinkie` or `@WorldDomination`
 occurred. Tag filters can be combined with other filters:
 
 ```sh
-jrnl -n 5 @pinkie @WorldDomination -and
+jrnl -n 5 "@pinkie" "@WorldDomination" -and
 ```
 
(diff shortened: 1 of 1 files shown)


jimmyhess (commented):
The quotation marks are necessary in Windows

micahellison:
> The quotation marks are necessary in Windows

Hi there, I'm testing this out on Windows 10 and it's working just fine without quotes. Can you clarify this?

jimmyhess:
From PowerShell, `jrnl @tag1 @tag2` launches the editor and `jrnl "@tag1" "@tag2"` prints the entries with `@tag1` *or* `@tag2`.

wren (changes requested):
Thanks for submitting this, but the changes don't quite work for us. Please see below for specific notes on each of the changes.

For the two issues in this PR:

- `-and` can't be in between other arguments
  This was reported earlier (#853), and is fixed and merged. It's not released, yet (it's in v2.5), but we don't want to update the documentation for this since it's already fixed in the next version (which is when this documentation would be updated). Sorry for the confusion.

- Powershell requires escaping of `@` charcater
  We have a note in the docs on the [usage page](https://jrnl.sh/usage/) that tells users that most shells have reserved characters (like `#` and `*`, or in the case of powershell `@`). Maybe we could make that more prominent, or specifically call out powershell, but we don't want to add shell specific syntax into the general help (like putting quotes around each `-and` on the page).

PR #1044, opened 2020-09-08
Add punctuation more commonly used in Asian languages (ellipsis) to sentence parsing

felixonmars:
Ellipsis is used to terminate sentences in Chinese and used a lot at end of title.

### Checklist

  for the same issue.
- [ ] All tests pass.

### DIFF
--- a/jrnl/Entry.py
+++ b/jrnl/Entry.py
@@ -202,7 +202,7 @@ class Entry:
 SENTENCE_SPLITTER = re.compile(
     r"""
 (                       # A sentence ends at one of two sequences:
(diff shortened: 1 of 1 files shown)


wren:
Characters added are [u2026](https://unicode-table.com/en/2026/) and [u22EF](https://unicode-table.com/en/22EF/).

2026 is in the "General Punctuation" block, but 22EF is in "Mathematical Operators." Does it make sense for 22EF to be part of this if it's for math instead of punctuation?

felixonmars:
It's at least used by default in Mac OS's default Chinese IME, and perhaps some others too. So I think it makes sense.

wren (approved):
Thank you!

Issue #625, opened 2019-07-21
2.0.0rc3.post2 now interprets chat logs with bracket timestamps as new diary entries


sorsasampo:
After 2.0.0rc3.post2 converted my journal, adding first entry seems to cause it to read the converted journal, and reinterpret chat log lines like this:

    [26/11, 23:32] Someone: chat line.

... as new entries, and move them to the end of the journal as new entries for "2019-07-21 23:32".

I understand the timestamp format has changed, but:

1) The parsing seems to be a little bit too greedy.
2) The journal conversion should escape this, if necessary.

wren:
Hi, thanks for the report! Is this the first version of v2.0 that you've used? And to be clear, this is a journal that you converted from v1.9.8, right?

sorsasampo:
@wren, yes to first v2.0 version. Not sure about the earlier v1.9.x version, but probably the latest.

wren:
This should now be fixed with the latest in master. New release candidate coming soon.

sorsasampo:
jrnl 2.0.0-rc4 still has this issue.

wren:
@sorsasampo Can you please provide an example of a v1.9.* journal entry with square brackets that doesn't get upgraded properly to v2.0? We added tests to deal with the problem reported in this issue (using [this example journal](https://github.com/jrnl-org/jrnl/blob/master/features/data/journals/simple_jrnl-1-9-5.journal)), and every manual test I've done seems to work as expected now.

sorsasampo:
@wren

```
$ cat journal.txt
2010-06-10 15:00 A life without chocolate is like a bad analogy.

[26/11, 23:32] Someone: chat line.
$ jrnl test
[Entry added to default journal]
$ cat journal.txt
[2019-08-27 12:00] test

[2019-08-27 23:32] Someone: chat line.
```

Not sure what happens to the first line in this case.

sorsasampo:
@wren: the above was just me replacing journal.txt and running jrnl 2.0.0-rc4. Does jrnl store some indication of the journal version outside the journal file?

The actual issues I'm seeing happened when running jrnl 1.9.*, then 2.0.0rc3.post2 (which I guess did the journal upgrade), and then 2.0.0-rc4.

[...]

PR #1045, opened 2020-09-10
Use strict date matching, if using the default time format

michaellenahan:
### Description

- What is this new code intended to do?

After updating to v2, I noticed the bug described here: https://github.com/jrnl-org/jrnl/issues/625

jrnl version v2.4.5

Using this configuration:
```
(jrnl-hIZQiBQS-py3.8) ➜  jrnl git:(develop) cat ~/.config/jrnl/jrnl.yaml
colors:
  body: none
  date: none
  tags: none
  title: none
default_hour: 9
default_minute: 0
editor: 'atom'
encrypt: false
highlight: true
indent_character: '|'
journals:
  default: ~/notes/journal.txt
  home: ~/notes/home.txt
  test: ~/notes/test.txt
linewrap: 79
tagsymbols: '@'
template: false
timeformat: '%Y-%m-%d %H:%M'
version: v2.4.5
```

I'm doing my test on the "test" journal, which starts out like this:
```
[2020-09-10 19:47] Here is a chat log
[26/11, 23:32] Someone: chat line.
```

Before the patch: the second line is incorrectly parsed as a new entry.
```
(jrnl-hIZQiBQS-py3.8) ➜  jrnl git:(develop) ✗ jrnl test --short
2020-09-10 19:47 Here is a chat log
2020-09-10 23:32 Someone: chat line.
```

After the patch: as long as the default time format is unchanged, the second line is correctly interpreted as part of the first entry.
```
(jrnl-hIZQiBQS-py3.8) ➜  jrnl git:(develop) jrnl test --short
2020-09-10 19:47 Here is a chat log
```

- Are there any related issues?

https://github.com/jrnl-org/jrnl/issues/630
https://github.com/jrnl-org/jrnl/issues/625

- What is the motivation for this change?

I love to use jrnl as a notepad for when I am dealing with server downtime.
When I'm doing that, I often edit my jrnl directly and paste large amounts of server logs there.
Those log entries all start with an opening square bracket and usually have some sort of date.

My motivation:
I accept this is not a full solution to the problem, and it will only help those who do not change the default time format.
For people who do stick to the default, I think jrnl should try its best to parse the date string at the beginning of an entry strictly.
That "fixes the bug" for the majority of users, and solves my problem.
Of course, if my log file entries have the same time format as jrnl, then jrnl will parse those log file entries as new jrnl entries.
That's fine, because the default jrnl time format is '%Y-%m-%d %H:%M' and most log entries are accurate to the second or millisecond.

### Checklist

  for the same issue.
- [ ] All tests pass.
(I'm a new contributor, I'm not sure how to test, so I will submit the PR and see what happens)

### DIFF
--- a/jrnl/Journal.py
+++ b/jrnl/Journal.py
@@ -31,6 +31,7 @@ class Journal:
             "default_hour": 9,
             "default_minute": 0,
             "timeformat": "%Y-%m-%d %H:%M",
+            "timeformat_strict": False,
             "tagsymbols": "@",
             "highlight": True,
             "linewrap": 80,
@@ -129,9 +130,12 @@ class Journal:
 
         date_blob_re = re.compile("(?:^|\n)\\[([^\\]]+)\\] ")
 
-        # Use strict date matching, if using the default time format
+        # Use strict date matching
+        # (1) if using the default time format, and
+        # (2) if timeformat_strict is set to true.
         if self.config["timeformat"] == "%Y-%m-%d %H:%M":
-            date_blob_re = re.compile("(?:^|\n)\\[(\d{4}-\d{2}-\d{2} \d{2}:\d{2})\\] ")
+            if self.config["timeformat_strict"]:
+                date_blob_re = re.compile("(?:^|\n)\\[(\d{4}-\d{2}-\d{2} \d{2}:\d{2})\\] ")
 
         last_entry_pos = 0
         for match in date_blob_re.finditer(journal_txt):
--- a/jrnl/Journal.py
+++ b/jrnl/Journal.py
@@ -128,6 +128,11 @@ class Journal:
         entries = []
 
         date_blob_re = re.compile("(?:^|\n)\\[([^\\]]+)\\] ")
+
+        # Use strict date matching, if using the default time format
+        if self.config["timeformat"] == '%Y-%m-%d %H:%M':
+            date_blob_re = re.compile("(?:^|\n)\\[(\d{4}-\d{2}-\d{2} \d{2}:\d{2})\\] ")
+
         last_entry_pos = 0
         for match in date_blob_re.finditer(journal_txt):
             date_blob = match.groups()[0]
--- a/jrnl/Journal.py
+++ b/jrnl/Journal.py
@@ -130,7 +130,7 @@ class Journal:
         date_blob_re = re.compile("(?:^|\n)\\[([^\\]]+)\\] ")
 
         # Use strict date matching, if using the default time format
-        if self.config["timeformat"] == '%Y-%m-%d %H:%M':
+        if self.config["timeformat"] == "%Y-%m-%d %H:%M":
             date_blob_re = re.compile("(?:^|\n)\\[(\d{4}-\d{2}-\d{2} \d{2}:\d{2})\\] ")
 
         last_entry_pos = 0
--- a/jrnl/Journal.py
+++ b/jrnl/Journal.py
@@ -135,7 +135,9 @@ class Journal:
         # (2) if timeformat_strict is set to true.
         if self.config["timeformat"] == "%Y-%m-%d %H:%M":
             if self.config["timeformat_strict"]:
-                date_blob_re = re.compile("(?:^|\n)\\[(\d{4}-\d{2}-\d{2} \d{2}:\d{2})\\] ")
+                date_blob_re = re.compile(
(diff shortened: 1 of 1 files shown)


wren:
Have to think about this one a bit. I'm not sure if this would break any current functionality for users depending on what timestamp they are using.

michaellenahan:
> Have to think about this one a bit. I'm not sure if this would break any current functionality for users depending on what timestamp they are using.

Another idea for this would be to offer a configuration setting in jrnl.yaml for an advanced user to provide a regex for strict matching.

That way you can be non-breaking for everyone, but have strict matching for those who know for certain that they want it.

wren:
Okay, thanks for you patience while I thought this all the way through. So, I think a config option (named `timeformat_strict` so it's alphabetized next to the format) is the best approach.

The problems I have with relying on the default format are the following:
- If a user hears about new strict matching and wants to use it, but they currently have a non-default time format. They'll naturally change their format back to the default. But with this PR, that config change wouldn't recognize any entries in their journal (since now strict mode is on and no entries have the default time format).
- If someone moves a journal between machines, they'll risk the same scenario above if the config doesn't match exactly.
- This leaves out users from countries that list the day first (`dd-mm-yyyy`).

If we do a config option instead (default should be false), then users should know what they're getting into when they turn it on.

wren (changes requested):
Please do the config option instead as discussed in the comments.

michaellenahan:
Thanks! I've added the code for the timeformat_strict configuration option.
Tested locally with the following test.txt file:
```
[2020-09-10 19:47] Here is a chat log
[26/11, 23:32] Someone: chat line.
```
[1] Without any `timeformat_strict` entry in `~/.config/jrnl/jrnl.yaml`
```
(jrnl-hIZQiBQS-py3.8) ➜  jrnl git:(develop) jrnl test --short
2020-09-10 19:47 Here is a chat log
2020-09-30 23:32 Someone: chat line.
```
[2] After adding `timeformat_strict: true` to `~/.config/jrnl/jrnl.yaml`
```
(jrnl-hIZQiBQS-py3.8) ➜  jrnl git:(develop) jrnl test --short
2020-09-10 19:47 Here is a chat log
```
[3] After adding `timeformat_strict: false` to `~/.config/jrnl/jrnl.yaml`
```
(jrnl-hIZQiBQS-py3.8) ➜  jrnl git:(develop) jrnl test --short
2020-09-10 19:47 Here is a chat log
2020-09-30 23:32 Someone: chat line.
```

michaellenahan:
Another test:

Using this as the test.txt file (where I have defined "test" as one of my `jrnl`s, see configuration at the beginning of this discussion).
```
[2020-09-10 19:47] Here is a @chat log
the chat log is like this:
[26/11, 23:32] Someone: chat line.
```

BEFORE adding `timeformat_strict: true` to the config:
The output is incorrectly excluding the third line.
```
$ jrnl test @chat
2020-09-10 19:47 Here is a @chat log
| the chat log is like this:
```

AFTER adding `timeformat_strict: true` to the config:
```
$ jrnl test @chat
2020-09-10 19:47 Here is a @chat log
| the chat log is like this:
| [26/11, 23:32] Someone: chat line.
```

wren:
Hi, @michaellenahan . I haven't time to properly review this yet (I'm in the middle of moving apartments). But can you please add the tests that you ran manually to one of the test files (in the`features`directory)? That would hi s long way to making sure this works as intended for everyone. I'll have time for a review and testing later this week. Thank you!

wren:
This doesn't quite work as requested. Someone would still need to have the default timestamp in order to be able to use strict parsing. This still has the same problems I outlined in my above comment.

I'm guessing your hesitation is coming from using a regex to parse the date (there's no way to make a regex to parse arbitrary date strings generated by a user). I think using [`strptime`](https://docs.python.org/3/library/datetime.html#datetime.datetime.strptime) in `time.py` instead of a regex in `Journal.py` would be a better implementation for this functionality.

What do you think?

wren:
Hi, I'm closing this for now to clean up the queue, but please feel free to reopen with the requested changes. Thanks, again!

michaellenahan:
Hi,

I'm using a (clumsy but workable) workaround right now, whereby I put a period "." in front of those dates in square brackets, that I don't want to have interpreted as jrnl entries.

Thanks for taking the time to respond @wren !

Have not had time to look into your suggested solution and right now I don't know when I will be able to. I will leave this here for now, it can be re-opened if anyone else comes across it and is experiencing the same problem.

Issue #1014, opened 2020-07-29
Configure default display format in config


wren:
## Feature Request

This request comes out of #996. 

### Use Case/Motivation

Some users always want to display their journals in a specific format. They can do that now by adding `--format markdown` (since v2.5) or `--export markdown` to each command. This is tedious and long. Sure, users can make an alias, but the `--format` flag can get in the way of certain commands (like `--edit`).

So, the idea came up to be able to specify a default format (either global, or per journal).

### Example Usage

If the config file has, say:
```yaml
journals:
  default:
    encrypt: true
    display_format: markdown
    journal: ~/Documents/Journals/default.txt

```
`jrnl -1` would output the same way as `jrnl -1 --format markdown` would.

Also, `jrnl -1 --format text` should still work (override the config file).

### Other information

If a format doesn't exist, jrnl should output an error message, and fallback to our default jrnl format.

xeruf:
This is complete with the release of 2.5, right?

wren:
@xerus2000 Yup! Thanks for pointing it out.

PR #1050, opened 2020-10-06
Add default display format option to config file

joaquingx:
- What is this new code intended to do? Add `display_format` to config.
- Are there any related issues? https://github.com/jrnl-org/jrnl/issues/1014 
- What is the motivation for this change? Use e.g. `--format md` everytime can be tedious.
- What is an example of usage, or changes to config files? (if applicable)
Change `display_format` on `config.yml` to any of the implemented formatters.

### Checklist

  for the same issue.
- [x] All tests pass.

Ready to review @wren

### DIFF
--- a/jrnl/jrnl.py
+++ b/jrnl/jrnl.py
@@ -290,7 +290,7 @@ def _delete_search_results(journal, old_entries, **kwargs):
         journal.write()
 
 
-def _display_search_results(args, journal, config, **kwargs):
+def _display_search_results(args, journal, **kwargs):
     if args.short:
         print(journal.pprint(short=True))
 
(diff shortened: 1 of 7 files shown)


jrnl/jrnl.py
```diff
@@ -300,7 +300,11 @@ def _display_search_results(args, journal, **kwargs):
     elif args.export:
         exporter = plugins.get_exporter(args.export)
         print(exporter.export(journal, args.filename))
-
     else:
-        # Default display mode
-        print(journal.pprint())
+        # Display according display_format config option
+        config_selected = kwargs["config"].get("display_format")
```
wren: Instead of nesting this conditional in the parent, can you please move it to the top level with the other `elif`s?

wren (changes requested):
This looks really good! Tests and docs included, even. Thank you!

Just one small change, and we'll be ready to go.

jrnl/jrnl.py
joaquingx: Hey, thanks, that would be better :+1:. Done.

wren (dismissed):
Thank you! 🎖️

Issue #1061, opened 2020-10-25
Low Contrast Text on Webpage


ghost:
### Use Case/Motivation

https://contrastrebellion.com/

### Example Usage

`color: #000` in `div.flex`

wren:
Hi, @klorophatu! Thanks for reporting this. Accessibility is important.

Could you please give some more specific details about which parts of the website don't have enough contrast? Some screenshots, if you have them, would also be very useful.

wren:
Scanning [jrnl.sh](https://jrnl.sh) with [a11y](https://color.a11y.com/Contrast/) shows this:

![Screen Shot 2020-10-26 at 11 00 17 AM](https://user-images.githubusercontent.com/9453067/97210226-833a0a00-177a-11eb-8f11-4c7999ee92b9.png)

For the failed results:
1. At a quick glance, I don't see where there's any white text on light grey background. So, I'm guessing this is some quirk in the scan. Maybe some bg property we're not setting correctly?
2. This is the grey text on grey background, which is absolutely not enough contrast, and I'm guessing what the OP was about. Should be an easy fix, though.

wren:
For the other pages (this one is from overview):

![Screen Shot 2020-10-26 at 11 03 35 AM](https://user-images.githubusercontent.com/9453067/97210543-f17ecc80-177a-11eb-96f9-da8c5fbd9739.png)


It looks like our blue link color is close, but just shy of the contrast requirements. We should pick out a slightly more contrasty blue.

wren:
Ouch, I won't paste the results here (because they don't fit on my screen for a screenshot), but our syntax highlighting has a lot of problems (scan https://jrnl.readthedocs.io/en/latest/formats/ with the tool linked above).

wren:
Also, since I'm just spamming this thread apparently, are there are good ways for us to automate these tests? Maybe it's not worth automating since we don't change our colors often, but if it's simple, it could streamline site updates in the future.

wren:
cc #1005

wren:
You can view this now here: https://jrnl.sh/en/latest/

With the next release, it'll be merged into the stable channel, and will be the default for everyone.

Thanks, again for pointing this out, @klorophatu!

Issue #1060, opened 2020-10-24
Migrate to Github Actions from Travis CI


wren:
CC: #1007

## Feature Request

We need to switch away from Travis CI (it's still getting worse). As mentioned in #1007, Circle seems like a good candidate.

We should start by mirroring our current setup, and running Travis and Circle side-by-side as a test.

Steps needed to complete this migration:
- [x] linux tests
- [x] mac tests
- [x] windows tests
- [x] linting
- [x] accessibility testing for docs site ([jrnl.sh](https://jrnl.sh))
- [x] changelog generator
- if beta
    - [x] send to pypi as prerelease
    - [ ] update [jrnl prerelease tap](https://github.com/jrnl-org/homebrew-prerelease)
- if release
    - [x] send to pypi
    - [ ] send updated pr to homebrew
- [x] trigger a pipeline to do everything for a release
    - [x] update version in correct files
    - [x] tag repo

wren:
Circle has nice tools, but we've been having some troubles with their billing. They give a bunch of free credits to open source projects which are only good for linux CI. I thought that would be fine, and we could just use those free linux credits, and pay for some extra mac and windows credits, but Circle doesn't like that. The moment you pay a single dollar for any service, they take away the open source free credits, so we'd also have to pay for linux builds. At their rates, it's prohibitive to pay for the linux, mac and windows entirely, so Circle doesn't seem like it's going to work out.

Github actions luckily doesn't have these restrictions and is completely free for open source repos. So, it's looking like that will likely be the direction we go in.

wren:
I put homebrew automation in the checklist, but we already have a separate issue for that (#686), so I'll that part for that issue.

PR #1067, opened 2020-10-29
Add accessibility testing for docs site (http://jrnl.sh)

wren:
This adds more work towards #1060. It wasn't originally listed in that issue, but #1061 pointed out an accessibility problem (low contrast) on the docs site. This pointed out that we don't currently test the docs site for accessibility, so that's what this PR adds.

It runs pa11y, and outputs the results in junit format (so that circle ci can show us the reports).

### Checklist

  for the same issue.
- [x] All tests pass.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,7 +13,12 @@ executors:
 
   python39:
     docker:
-      - image: circleci/python:3.9.0
+      - image: cimg/python:3.9.0
+    resource_class: small
+
+  with_browser:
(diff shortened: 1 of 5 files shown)


wren:
Btw, we should set the accessibility tests to be required until all the issues are fixed. Which, uh, could be a while. 😬 

Screen Shot 2020-10-28 at 5 26 31 PM

micahellison (approved):
👓

Issue #1052, opened 2020-10-10
Conduct a dependency audit


micahellison:
We should review each dependency and identify:

* If the dependency is still justified
* The dependency version scheme
* How permissive the version pinning should be

We're trying to do this at least annually, if not twice a year. This would be good to complete before the end of 2020.

wren:
CC: #1009 #783

PR #1076, opened 2020-11-07
Update dependencies - pyxdg, pytest, black

micahellison:
I checked the various dependencies in `pyproject.toml` for updates and found an update for pyxdg, the first in a couple of years. I've updated the pin to the current version. Release notes [here](https://www.freedesktop.org/wiki/Software/pyxdg/).

Apologies for the noisy diff on the lockfile. That's just how poetry is running it for me. (Side note: on Windows, I had to use `dos2unix` to change the line endings)

Relates to #1052.

A few dev dependencies are updated too:
* [pytest](https://docs.pytest.org/en/stable/changelog.html)
* [black](https://github.com/psf/black/blob/master/CHANGES.md) - this PR includes the results of `make format` after updating
### Checklist

  for the same issue.
- [X] All tests pass.

### DIFF
--- a/features/steps/export_steps.py
+++ b/features/steps/export_steps.py
@@ -50,7 +50,7 @@ def check_output_field_key(context, field, key):
 @then("the json output should contain {path}")
 @then('the json output should contain {path} = "{value}"')
 def check_json_output_path(context, path, value=None):
-    """ E.g.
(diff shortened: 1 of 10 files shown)


wren (approved):
🏜

PR #1091, opened 2020-11-21
Add ability to put --edit partly through a cli entry to move it to the editor

### LINKED ISSUE #906: Allow --edit flag partway through an entry
I think it would be really helpful if the --edit flag could be invoked partway through an entry, for example:

```
$ jrnl now: I just finished homeschooling @Jane for the day. We talked about contractions, I explained that do not is often written as (oh, I\'ll need a lot of single quotes, but I really don\'t feel like escaping all of \'em) --edit
```
This would then open a file with all of the text up until the --edit already in it, so you could keep writing comfortably without copying out of your shell and pasting into your editor.

wren:
Fixes #906

Also add some functionality to test suite to see what the editor file contents were and make assertions about it.

### Checklist

  for the same issue.
- [x] All tests pass.

### DIFF
--- a/features/steps/core.py
+++ b/features/steps/core.py
@@ -230,6 +230,21 @@ def matches_editor_arg(context, regex):
     ), f"\nRegex didn't match exactly 1 time:\n{regex}\n{str(args)}"
 
 
+@then("the editor file content should {method}")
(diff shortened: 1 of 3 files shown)


micahellison (approved):
📝

PR #1093, opened 2020-11-22
Fixes for new CI pipeline (Github Actions)

wren:
More work on #1060

- Support "[ci skip]" tag on commits to avoid build dupes
- Add smarter path detection so we don't spam tons of tests
- Allow steps to cancel if previous steps were cancelled (don't always
  run)
- Separate workflows to be more modular

### Checklist

  for the same issue.
- [x] All tests pass.

### DIFF
--- a/.github/workflows/changelog.yaml
+++ b/.github/workflows/changelog.yaml
@@ -6,7 +6,7 @@ on:
 
 jobs:
   generate:
-    if: contains(toJson(github.event.commits), 'Update changelog') == false
+    if: contains(toJson(github.event.commits), '[ci skip]') == false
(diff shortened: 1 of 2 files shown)


.github/workflows/testing.yaml
```diff
@@ -35,79 +48,37 @@ jobs:
         poetry install --remove-untracked
 
     - name: Code formatting (Black)
-      if: always()
+      if: success() || failure()
```
wren: Note: Using `success() || failure()` here means that it won't run if the previous job is cancelled, which is usually what we actually want. `always()` kept running after cancellation and lead to some pretty useless results.

PR #1101, opened 2020-11-28
Add fix for changelog conditional always returning false

wren:
The changelog generator check to see if anything changed from `git diff` was always returning false. This was because the env var was being set, but not exported to the github file that subsequent steps need to see the correct value. This fixes that, and the changelog generator should run again.

Example of the workflow running successfully:
https://github.com/wren/jrnl/runs/1468224087?check_suite_focus=true

### Checklist

- [n/a] I have included a link to the relevant issue number.
  for the same issue.
- [x] All tests pass.


micahellison (approved):
✅

PR #1104, opened 2020-11-28
Check for readline module instead of Windows when initializing autocomplete in install

### LINKED ISSUE #1015: Missing readline module can be dealt with more gracefully
## Bug Report

In some Python distributions, such as ActivePython on Windows, the readline module is missing due to licensing issues. This module is only used to add some nice-to-have features (like tab completion) to a single step, so we modified jrnl to only add that functionality if the user isn't on Windows (#738).

This fixes the issue for Windows users, but it turns out there are other situations where this problem exists too. And it gets worse: the error message mentions that the readline module is missing, so users fetch it from pypi, only to end up with a [deprecated module](https://pypi.org/project/readline/) that leads to more obscure errors ("core dumped") on certain systems (#1010).

We should be able to mitigate this issue almost entirely by checking if readline exists on the system at all (`'readline' in sys.modules`), rather than checking if the user is running Windows.

micahellison:
Fixes #1015 and should now prevent crashes on install for ActivePython users that aren't on Windows. We haven't seen this `readline` issue as much lately, and unfortunately I can't reproduce every environment in which it happens, but this should hopefully mitigate the problem further, and the unit test offers some peace of mind.

### Checklist

  for the same issue.
- [ ] All tests pass.

### DIFF
--- a/jrnl/install.py
+++ b/jrnl/install.py
@@ -12,13 +12,8 @@ from . import __version__
 from .config import load_config
 from .config import verify_config_colors
 from .exception import UserAbort
-from .os_compat import on_windows
 from .prompt import yesno
 from .upgrade import is_old_version
-
-if not on_windows:
-    # readline is not included in Windows Active Python
-    import readline
 
 DEFAULT_CONFIG_NAME = "jrnl.yaml"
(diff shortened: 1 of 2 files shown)


wren (approved):
🌞

PR #1105, opened 2020-12-05
Make docs site (jrnl.sh) fully meet Web Content Accessibility Guidelines (WCAG) 2.1

### LINKED ISSUE #1061: Low Contrast Text on Webpage

wren:
Fixes #1061 

This PR fixes all of the acessibility issues that pa11y was reporting on our docs site. It also re-enables accessibility testing for PRs and branches that change anything related to the docs.

The main issues were:
1. Low contrast colors choices
    [jrnl.sh](https://jrnl.sh) will now meet the AAA standard specified in [WCAG 2.1](https://www.w3.org/TR/WCAG21/#contrast-minimum), which is to have a minimum of 4.5:1 contrast ratio for all text on the site.
2. Our search bar was missing a submit button
    No submit button makes it hard for anyone that can't use a keyboard to search our docs site. Now there's a button, so everybody wins.

Note that I had to move the mkdocs theme folder out of `docs` and into `docs_theme`. This is because readthedocs doesn't recognize `main.html` (which is necessary to add the missing search button) inside of the docs directory.  Mkdocs' site says not to put the custom theme inside the docs directory anyway, so I guess this fixes both of those issues at once.

Also, here's an example build where this all ran successfully:
- https://github.com/wren/jrnl/runs/1502555163?check_suite_focus=true

### Checklist

  for the same issue.
- [x] All tests pass.


wren:
I tried to keep the colors as similar to the original as possible, but now with the higher contrast, some of the colors look a bit garish.

Like [this page](https://jrnl--1105.org.readthedocs.build/en/1105/formats/) honestly looks a bit alarming. Maybe we should tone down the red?

wren:
Okay, all colors are now updated, and this PR should be good to go.

micahellison (approved):
📄📄📄📄

PR #1110, opened 2020-12-12
Update all dependencies and lock file

### LINKED ISSUE #1052: Conduct a dependency audit

wren:
Fixes #1052

This also unpins versions of some dependencies where the API is very
stable, or pinning wasn't getting us any additional stability.

Co-authored-by: Micah Jerome Ellison <micah.jerome.ellison@gmail.com>

### Checklist

  for the same issue.
- [x] All tests pass.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1,76 +1,73 @@
 [[package]]
-category = "main"
-description = "textwrap, but savvy to ANSI colors and styles"
 name = "ansiwrap"
+version = "0.8.4"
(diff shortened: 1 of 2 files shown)


micahellison (approved):
🔓

PR #1111, opened 2020-12-13
Add automatic deployment for homebrew releases (and prereleases)

### LINKED ISSUE #686: Automate Homebrew deployment
We have been prioritizing `pip install jrnl` as our preferred installation method. We used to support `brew isntall jrnl`, but that has fallen by the wayside. We need to support Homebrew again. It needs to be automated, though.

This was previously done in #174, but it now needs to be automated in the pipeline (alongside releases to pypi.).

wren:
Fixes #686

This new job will submit a PR to the relevant upstream repo as part of
the release workflow.

If we decide to later, we can make it directly commit to our prerelease repo instead of submitting a PR.

Here's an example of a build where this ran successfully:
- https://github.com/wren/jrnl/actions/runs/419539722

And here's an example PR that the bot opened:
- https://github.com/wren/homebrew-prerelease/pull/9

### Checklist

  for the same issue.
- [x] All tests pass.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -14,6 +14,7 @@ jobs:
     - name: Validate version
       run: |
         JRNL_VERSION="${{ github.event.inputs.version }}"
+        echo "::debug::version: $JRNL_VERSION"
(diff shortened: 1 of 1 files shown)


micahellison (approved):
☕

Issue #1006, opened 2020-07-25
Namespace-based Export Plugins


minchinweb:
I'm thinking about reworking the export plugin system to be auto-discovered based on the module name. Assuming this works well, it could be expanded to import plugins and possibly the storage backends as well.

There would be two namespaces: one for internal plugins, and another for external/contributed plugins.

The usecase would be to solve issues such as #779 where there is a difference of opinion of what the export should look like, or if you have a special format that you need. "Contributed" plugins could also be a way to test out a format before added in to the core of *jrnl*.

For myself, I have a "special" plugin that I used as a prep step to feed my jrnl entries into Pelican (a static site generator). It's always seemed like too "special" of a case to submit for the main repo, but it's something I use regularly. With the support and bug fixes to DayOne support, this is probably the only reason I have left to keep a private fork of *jrnl*.

In terms of implementation, I'm imagining that each plugin would have five methods:

- one that returns it's name (e.g. "Markdown Exporter")
- one that returns it's version (e.g. "2.4.4-beta"; for built-in plugins, this would match *jrnl*)
- one that returns a list of "format codes" that it will process (e.g. `["md", "markdown"]`; this is was will used by the `export` command of *jrnl*)
- one that is passed a single entry and returns the exported entry
- one that is passed a journal and returns the exported journal

I expect that contributed plugins would override the build-in plugins if they defined the same format code.

By using namespaced plugins, the plugin would be install-able via *pip* and would require no more configuration or activation beyond being importable by Python.

I haven't written any code yet (but I've been trying to figure out how to do this for probably 3 years...), but I wanted to get some thoughts on if there was support for this before I spend a bunch of time on it.

Thanks,

wren:
Thanks for bringing this up; I've also been wanting to get our plugins re-done.

I'm still thinking through this, but here are my thoughts so far.

- Our current system doesn't mesh well with distribution methods. This is due to the fact that plugins are expected to be loaded from a directory the jrnl core. This works just fine if you're running jrnl from source, but anywhere else (pipx, brew, pip, etc) there's a decent chance your plugin will be uninstalled on upgrade. The best way to deal with this at the moment is to either reinstall your plugin (copy the file into the correct folder in jrnl) after each upgrade, or maintain a private fork (as you mentioned) where you basically have a patch for your plugin on top of the jrnl source. Both are kind pretty annoying, and users really shouldn't have to deal with either of them.

[...]

PR #1115, opened 2020-12-14
[WIP] External Plugins

minchinweb:
@wren This one's for you! This has been rattling around in my brain for soooooo long, but sometimes a little push at the right time is all that is needed to make the magic happen. So here we are. :)

This is a first pass at allowing external, namespace plugins.

Existing plugins are now in `jrnl.plugins.importer` and `jrnl.plugins.exporter`. "Private" plugins are to be in the `jrnl.contrib.importer` and `jrnl.contrib.exporter` namespaces (so I might have my exporter at `jrnl.contrib.exporter.minchinwebexporter`).

Some other code was moved around to avoid circular imports and other weird import issues. The code hasn't been extensively tested, but `jrnl` appears to still be working locally, including listing all the export formats.

No documentation has been added.
I also haven't run the tests (but the CI pipeline will soon tell me if I broken anything).

In the future, backends could probably also be added as/converted to plugins, but this doesn't try to do that.

c.f. #1006


wren:
Awesome, thank you! I'll take a look at it by this weekend.

jrnl/plugins/util.py
```diff
@@ -1,11 +1,10 @@
 #!/usr/bin/env python
 # encoding: utf-8
 
-
 def get_tags_count(journal):
     """Returns a set of tuples (count, tag) for all tags present in the journal."""
     # Astute reader: should the following line leave you as puzzled as me the first time
-    # I came across this construction, worry not and embrace the ensuing moment of enlightment.
+    # I came across this construction, worry not and embrace the ensuing moment of enlightenment.
```
wren, line 7: haha, nice

jrnl/plugins/meta.py
```diff
@@ -0,0 +1,84 @@
...
+    name: importlib.import_module(plugin.name)
+    for plugin in __importers_contrib
+    for name in importlib.import_module(plugin.name).Importer.names
+}
+
+__exporter_types = {
+    **__exporter_types_builtin,
+    **__exporter_types_contrib,
+    **__exporter_types_template,
+}
```
wren, line 66: So, this would override builtin plugins with contrib plugins of the same type, right? Does that behavior get communicated to the user in any way? If somebody runs `jrnl --format markdown`, do they have any way to know if that's a builtin, or some random plugin they forgot they installed?

This would also be relevant for users reporting bugs. We need a way for us to know which version of which plugin they are using, I think.

wren:
@MinchinWeb From a user standpoint, what does this look like? If I want to install a new plugin, what are the steps to do that? It seems like putting a file in the new contrib directory would work. Is that right? Are there other steps?

jrnl/plugins/meta.py
```diff
@@ -0,0 +1,84 @@
...
+__importers_builtin = list(
+    pkgutil.iter_modules(
+        jrnl.plugins.importer.__path__, jrnl.plugins.importer.__name__ + "."
+    )
+)
+__importers_contrib = list(
+    pkgutil.iter_modules(
+        jrnl.contrib.importer.__path__, jrnl.contrib.importer.__name__ + "."
+    )
+)
```
wren, line 34: Instead of duplicating each of these (one for builtin, and one for contrib), can we have a function that will do all of this work given a path? Then we could run that function once for each path (builtin, and contrib).

wren: A standalone function would also be easier to add tests for.

jrnl/plugins/meta.py
minchinweb: My first thought was to add the list of plugins (with versions) to the information displayed by `jrnl --version`. But nothing has been implemented quite yet.

minchinweb: (Trying to do this from my phone wasn't working... so now I'm at the computer)

> this would override builtin plugins with contrib plugins of the same type, right?

Yes, that is what is supposed to happen. I haven't tested it yet....

> Does that behavior get communicated to the user in any way?

No, not currently. My thought was to extend `jrnl --version` to list the installed plugins, but to note the plugin used at the start of the export would be good too. In any case, the version of the plugin should be printed as well.

jrnl/plugins/meta.py
minchinweb: Can be done!

minchinweb:
> From a user standpoint, what does this look like? If I want to install a new plugin, what are the steps to do that? It seems like putting a file in the new contrib directory would work. Is that right? Are there other steps?

I've done this using namespace plugins. So to develop a new plugin, you would create a new Python project, with a layout something like:

```
jrnl
  `- contrib
       `- exporter
            `- myfancyexporter.py
```

(Note, no `__init__.py` files.)

`myfancyexporter.py` would contain an `Exporter` class, with the properties `names` (a list, and what you use as a "name" to call the exporter from `jrnl`), `extension` (a string, which is what file extension the exported files will have), and class methods `export_entry` (to export a single entry) and `export_journal` (to export the entire journal). (There should also be something to return the version of the plugin; interface TBD.) 

You would then package it like any other Python library. The user would then install it like any other Python library (via `pip` or `poetry`, etc). Assuming everything works, a new exporter would be available to `jrnl` simply by virtual of the plugin package being installed.

In theory, putting a file into `jrnl`'s *contrib* folder would work, but with namespace plugins (like I'm aiming for here), the plugins available to Python at `jnrl.contrib` **do not** have to be in the same ("physical") folder. So ideally, you should never have to directly touch your `jrnl` installation to install (or use) your "personal" plugins, nor will they be wiped on upgrades. Your personal plugin can also be versioned like any other project dependency.

wren:
> In theory, putting a file into `jrnl`'s _contrib_ folder would work, but with namespace plugins (like I'm aiming for here), the plugins available to Python at `jnrl.contrib` **do not** have to be in the same ("physical") folder. So ideally, you should never have to directly touch your `jrnl` installation to install (or use) your "personal" plugins, nor will they be wiped on upgrades. Your personal plugin can also be versioned like any other project dependency.

In that case, why are we including an empty contrib directory in jrnl? If we don't want users touching it, then I think we should not include it to discourage its use, right?

minchinweb:
> In that case, why are we including an empty contrib directory in jrnl?

To make sure the namespace exists, even if it's empty. If it doesn't exist, we can't import it, and we can't iterate over the contained packages.

wren:
I think this is looking okay. What this still needs:

- [ ] tests
- [ ] handling for name conflicts among plugins (to make it clear to the user, and us, when a plugin is installed)
- [ ] some sort of output for versions when a plugin is installed (maybe `--diagnostic`?)

minchinweb:
Also:

- [ ] documentation on making your own plugin

minchinweb:
@wren: I updated this with an output that shows the active plugins and their versions, and added documentation. Testing is more complicated, as I need a way to create a new "testing environment" to install the sample plugins as python packages. Maybe that's simple, but I haven't figured it out yet and I wanted to share what I have done.

I created a new branch because I gave up on trying to figure out how to cleanly push here, so the code is now at #1216.

Thank you for all you do to keep the program moving forward!

PR #1116, opened 2020-12-14
Change approach for docs workflow to use pa11y-ci

wren:
The previous plugin looked nice with the comments and annotations, but apparently doesn't work for PRs. It can only run on the pr target (our branch), which means that the tests are never actually run against the new code.

As you can guess, this is pretty useless for PRs. So, this PR fixes that, and uses the very nice pa11y-ci (instead of the normal pa11y). While it doesn't upload annotations, the output is very readable and seems like the best we're going to get.

### Checklist

  for the same issue.
- [x] All tests pass.

Note: The failing tests docs test will be fixed in a separate PR. This is just to get the CI working to where that PR will get real test results.

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -9,7 +9,7 @@ on:
     - 'mkdocs.yml'
     - 'readthedocs.yml'
     - '.github/workflows/docs.yaml'
-  pull_request_target:
+  pull_request:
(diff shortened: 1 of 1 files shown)


wren:
See #1113 for how the above issue affects PRs.

micahellison (approved):
📃

PR #1126, opened 2020-12-20
Fix failing contrast test in accessibility tools on docs site

wren:
There were just a few remaining accessibility fixes needed for [jrnl.sh](https://jrnl.sh). These don't really affect newer browsers, but the older browsers that are built in to some accessibility tools get confused by the way we were doing gradients. Our test suite detected everything properly, but some other tools (like color.a11y.com) use older browsers, so we were still failing tests there.

So, other than one small color change, this PR shouldn't affect what the site looks like at all, and is mostly to make sure public accessibility tools pass for our site (and to assure users we care).

### Checklist

  for the same issue.
- [x] All tests pass.

### DIFF
--- a/docs_theme/assets/index.css
+++ b/docs_theme/assets/index.css
@@ -66,6 +66,7 @@ h3 {
 }
 
 header {
+  background-color: --mid-purple;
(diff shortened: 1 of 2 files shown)


micahellison (approved):
🟪

PR #1127, opened 2020-12-20
Fix changelog generator

wren:
This provides a few small fixes to the changelog generator tool:
- In github actions, booleans always end up as strings. So, it's not enough to test `env.FULL_RELEASE` because both 'true' and 'false' will evaluate as true. This fixes the conditional to check for the string value 'true' instead of relying on a boolean.
- This updates the regex to find the version header in our changelog.  There was a bug that affected patch, alpha, and beta versions. For example, if we were looking for "v2.5" we might find "v2.5.1" or "v2.5.1-alpha" or other versions like that. This led to the changelog not being cleared properly on update, and old versions sticking around longer than they should.
- Re-add option to include author link for merged PRs in changelog.  This option was accidentally left out when migrating the changelog generator tool to github actions.

### Checklist

- [x] I have checked to ensure there aren't other open [pull requests](../pulls) for the same issue.
- [ ] All tests pass.

### DIFF
--- a/.github/workflows/changelog.yaml
+++ b/.github/workflows/changelog.yaml
@@ -112,6 +112,7 @@ jobs:
           releaseUrl: https://pypi.org/project/jrnl/%s/
           releaseBranch: develop
           verbose: false
+          author: true
 
       - name: Small fixes
         run: |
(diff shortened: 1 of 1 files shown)


micahellison (approved):
📄📄📄

PR #1134, opened 2020-12-27
Notify user when config directory can't be created because there is already a file with the same name

### LINKED ISSUE #659: Commands fail if a file conflicts with the config directory
If there's a file (in my case `~/.config/jrnl` in place of where config directory should be, all commands fail (even `jrnl --version`) with the below stack trace.

This should be easily fixable by checking the config path for conflicts, and outputting a nice message to the user.

```
  File "/usr/local/bin/jrnl", line 6, in <module>
    from jrnl.cli import run
  File "/usr/local/lib/python3.7/site-packages/jrnl/cli.py", line 14, in <module>
    from . import install
  File "/usr/local/lib/python3.7/site-packages/jrnl/install.py", line 25, in <module>
    CONFIG_PATH = xdg.BaseDirectory.save_config_path(XDG_RESOURCE) or USER_HOME
  File "/usr/local/lib/python3.7/site-packages/xdg/BaseDirectory.py", line 58, in save_config_path
    os.makedirs(path, 0o700)
  File "/usr/local/Cellar/python/3.7.4_1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/os.py", line 221, in makedirs
    mkdir(name, mode)
FileExistsError: [Errno 17] File exists: '/Users/jonathanwren/.config/jrnl'
```

micahellison:
Fixes #659. It catches an exception in the XDG method and lets the user know exactly where the problem is.

While working on this, I noticed that a ton of configuration code was in the install.py file, so I moved it to config.py as necessary and updated the tests to use patches instead of a global. The configuration code still needs a lot of work but I hope this is at least a step in the right direction.

I couldn't come up with a good strategy to automatically test this issue, since XDG is essentially global to the program. Here's my manual test:
```
C:\dev\jrnl>copy con C:\Users\micah\.config\jrnl
this is a text file
^Z
        1 file(s) copied.

C:\dev\jrnl>jrnl --ls
The path to your jrnl configuration directory is a file, not a directory:
C:\Users\micah\.config\jrnl
Removing this file will allow jrnl to save its configuration.

C:\dev\jrnl>del C:\Users\micah\.config\jrnl

C:\dev\jrnl>jrnl --ls
Path to your journal file (leave blank for C:\Users\micah\.local\share\jrnl\journal.txt):
Do you want to encrypt your journal? You can always change this later [y/N]
Journals defined in C:\Users\micah\.config\jrnl\jrnl.yaml
 * default -> C:\Users\micah\.local\share\jrnl\journal.txt
```

### Checklist

  for the same issue.
- [X] All tests pass.


jrnl/cli.py
```diff
@@ -33,5 +34,9 @@ def cli(manual_args=None):
 
         return run(args)
 
+    except ConfigDirectoryPathIsFileException as e:
```
wren: Should we name each exception with a unique name? Or should we have a generic exception (like `JrnlError` or whatever) that we catch here?

As we add more exceptions, we'll have more import statements in cli.py, and more excepts on the try block in the cli function, and a pile of blank exceptions in exception.py, but they'll all do the same thing (output a message and return). When the message already tells us what went wrong, does a unique exception name get us anything more for this eventual bloat?

features/core.feature
```diff
@@ -13,6 +13,7 @@ Feature: Functionality of jrnl outside of actually handling journals
         Then the output should match "^jrnl version v\d+\.\d+(\.\d+)?(-(alpha|beta)\d*)?"
 
     Scenario: Running the diagnostic command
+        Given we use the config "simple.yaml"
```
wren: Why do we need to specify a config here? Doesn't the diagnostic command run and exit before a config is ever loaded?

micahellison: It sure does! I think I left this in to deal with a problem in the test code that I resolved later. Removing now.

jrnl/cli.py
micahellison: I like `JrnlError` - went ahead and implemented it.

wren (approved):
🤠

PR #1137, opened 2020-12-31
Clarify installation documentation (#1097)

### LINKED ISSUE #1097: Clarify installation docs
### Affected Page(s)

  - https://jrnl.sh/installation/#installation

### What Could Be Better?

The documentation isn't super clear about what to do with pipx. There are sometimes multiple ways to install pipx, and not all of them might work. We had an issue come up recently where a user installed pipx through apt, and had lots of troubles (#1087). We should clarify this a bit in our docs instead of just linking to other docs. At very least, there should be a note about following the link to see installation instructions there.

### Other Information

This may or may not also affect the homebrew installation instructions.

Seopril:
This is intended to fix https://github.com/jrnl-org/jrnl/issues/1097. The documentation was updated to include two notes regarding issues that could be encountered during installation. This should reduce the confusion of issues that occur when pipx is installed via apt. It also should help clarify path issues that arise when the user uses the sudo command with pipx. I've attached a screenshot of what the installation documentation looks like:

image

### Checklist

  for the same issue.
- No test changes needed
- [X] All tests pass.
- All tests passed. 5 feature scenarios were skipped and 27 feature steps were skipped when running "make test".

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -18,12 +18,12 @@ pipx install jrnl
 ```
 
 !!! note
-     `pipx` should be installed through `pip`. Missing dependencies and other issues
-     may occur when installing `pipx` through `apt` or another package manager.
+     `pipx` should be installed through either `brew` or `pip`. Missing dependencies and other issues
+     may occur when installing `pipx` through `apt` or another package manager. Further installation
+     instructions can be found in [pipx's documentation](https://pipxproject.github.io/pipx/installation/).
      
-!!! note
-     If you experience path issues after installing `jrnl`, ensure that the previous
-     command was run as the intended user, without `sudo`.
+!!! tip
+     Do not use `sudo` while installing `jrnl`. This may lead to path issues.
 
 The first time you run `jrnl` you will be asked where your journal file
 should be created and whether you wish to encrypt it.
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -17,6 +17,14 @@ On other platforms, install `jrnl` using [Python](https://www.python.org/) 3.6+ 
 pipx install jrnl
 ```
 
+!!! note
+     `pipx` should be installed through `pip`. Missing dependencies and other issues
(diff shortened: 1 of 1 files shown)


docs/installation.md
```diff
@@ -17,6 +17,14 @@ On other platforms, install `jrnl` using [Python](https://www.python.org/) 3.6+
 pipx install jrnl
 ```
 
+!!! note
+     `pipx` should be installed through `pip`. Missing dependencies and other issues
```
wren: Can we note here that pipx's favored install methods are brew and pip (in that order)? Maybe adding another link to pipx's installation page would be a good idea (in case their installation methods change).

docs/installation.md
```diff
@@ -17,6 +17,14 @@ On other platforms, install `jrnl` using [Python](https://www.python.org/) 3.6+
...
+     may occur when installing `pipx` through `apt` or another package manager.
+     
+!!! note
```
wren: Could this be a tip instead of a note?

docs/installation.md
```diff
@@ -17,6 +17,14 @@ On other platforms, install `jrnl` using [Python](https://www.python.org/) 3.6+
...
+     If you experience path issues after installing `jrnl`, ensure that the previous
```
wren: Can we say this in a more direct way? Maybe something like "Don't use sudo to install jrnl" or something along those lines.

wren (changes requested):
Thank you for taking this on!

I have a few notes, but it should be good to merge with minor changes.

Seopril:
No problem! I should have time to add these changes later tonight.

docs/installation.md
Seopril: Just for clarification, are you referring to the first note being changed to a tip, the second note being changed to a tip, or both?

wren (approved):
Looks good. Thank you!

PR #1145, opened 2021-01-03
Add new `-today-in-history`, `-month`, `-day`, and `-year` search filters

### LINKED ISSUE #1143: Filter for entries from the same date in previous years
## Feature Request
Would love to be able to see what I was doing today in previous years.

### Use Case/Motivation
Facilitates reminiscing, which is a major reason many people keep a journal in the first place.

### Example Usage
``` 
jrnl -reminisce
```
Result:
```
01-07-2015 21:00 Monty Python and the Holy Grail 
@movie

01-07-2016 20:00 Family @board-games night

01-07-2019 10:00 Graduated from uni
```

Would be nice to print `(_ years ago)` for each entry as well. Maybe like this?
```
01-07-2015 21:00 (5 years ago) Monty Python and the Holy Grail 
@movie

01-07-2016 20:00 (4 years ago) Family @board-games night

01-07-2019 10:00 (1 year ago) Graduated from uni
```

KarimPwnz:
I have implemented `-month`, `-day`, and `-year` for more general searching (values are parsed through `time.parse`). Additionally, this PR includes `-reminisce` which shows entries from today over the years, by modifying `-month` and `-day` (this resolves #1143).

However, I did not implement the `(_ years ago)` output suggested in #1143: it would add too much complexity. ~~Also, the included tests check `-reminisce` implicitly through `-month` and `-day`: I did not find a way to elegantly create journals exactly a year ago.~~

**Edit:** I have implemented `-reminisce` tests—see comments below.

### Checklist

  for the same issue.
- [x] All tests pass.


KarimPwnz:
There is some accidental functionality overlap with `-on` (you can do `-on January`, etc.); however, `-month`, `-day`, and `-year` are explicit, more general, and can be used in combination. Furthermore, `-on` is only meant for a single date.

KarimPwnz:
~~For testing `-reminisce` explicitly, we could use `unittest.mock.patch` on `datetime.datetime` in a step that creates journals exactly n year(s) ago.~~

**Edit:** I ended up using the built-in `time.parse` in `-reminisce` and mocking it in tests.

features/search.feature
```diff
@@ -214,6 +214,47 @@ Feature: Searching in a journal
         | But I'm better.
         """
 
+    Scenario Outline: Searching by month, day, or year
```
wren: This is a great start to the tests. For tests, we want to be sure to include the other journal types (encrypted, directory, day one) and not just the default type.

We've recently found some features that don't work across every journal type, and we didn't even know about it until we expanded testing in this way.

jrnl/jrnl.py
```diff
@@ -77,6 +78,10 @@ def _is_write_mode(args, config, **kwargs):
             args.edit,
             args.export,
             args.end_date,
+            args.reminisce,
```
wren: Working on a new name for this arg (see the issue).

features/search.feature
KarimPwnz: Ah! Fair enough. Do you recommend I create new journals for each type based on `dates_similar.journal`, or should I try to adopt the existent journals?

jrnl/jrnl.py
KarimPwnz: 👍

jrnl/Journal.py
```diff
@@ -220,11 +223,19 @@ def filter(
         if contains:
             contains_lower = contains.casefold()
 
+        # Create datetime object for comparison below
+        # this approach allows various formats
+        if month or day or year:
+            compare_d = time.parse(f"{month or 1}/{day or 1}/{year or 1}")
```
wren: Careful! We have some users that use day-first formats (e.g. 31/12/2021). Could any settings like potentially invert the month and day here??

features/steps/core.py
```diff
@@ -159,6 +160,11 @@ def disable_keyring(context):
     keyring.core.set_keyring(NoKeyring())
 
 
+@given('we set current date and time to "{dt}"')
```
wren, line 163: 🎉

jrnl/Journal.py
KarimPwnz: `compare_d` is independent of the journal type and the configured dateformat—it's only used for comparison.

KarimPwnz: Note that `time.parse` returns a `datetime` object.

wren: Also, what happens if a user enters something like `jrnl -month feb`?

wren (changes requested):
This is a promising start.

I put a few comments inline. Once those, and the questions in the issue are resolved, I think we can move this one forward. 

Thank you!

jrnl/Journal.py
KarimPwnz: It should work perfectly! I accounted for it in the tests too.

KarimPwnz: Hmm... perhaps I can change `f"{month or 1}/{day or 1}/{year or 1}"` to something clearer. In either case, it works!

features/search.feature
KarimPwnz: I could also modify the existent journals. Thoughts?

KarimPwnz: Done 👍

jrnl/Journal.py
KarimPwnz: It seems like `month/day/year` is the only one `parsedatetime` (which is used in `time.parse`) parses like we need.

wren: Ah, okay. I'm glad you've looked at this already!

I thought this might be affected by system settings (like locale or timezone, maybe). Are you sure those don't do anything here?

wren (dismissed):
This looks great. Thank you!

We're just waiting on nailing down the name for the reminisce arg, and then we can get this merged in.

jrnl/Journal.py
KarimPwnz: Turns out that `dateutil` is responsible for parsing this, and in debugging their code, I found a more standard format: `month.day.year`.

Regarding system settings, I doubt there is any issue: I am using jrnl's `time.parse`.

I am marking this as resolved.

jrnl/jrnl.py
KarimPwnz: Done.

KarimPwnz:
I've updated the name. I think this is complete now.

wren:
@KarimPwnz Thank you! I started a new dayjob last week and it's taking up all my time during the week right now. I'll get through every review and issue reply on Saturday. Thanks for your patience!

KarimPwnz:
Oh, congrats on the job! Do take your time.

wren (approved):
Looks great! Thank you!

PR #1146, opened 2021-01-04
Add new date format (`--format date`) for heatmapping

### LINKED ISSUE #743: Add daily streak report with calendar heatmap
It'd be cool to be able to see how often you're journaling. Extracted from #591.

I'm thinking that the calendar heatmap option in [termgraph.py](https://github.com/mkaz/termgraph) is a good choice.

KarimPwnz:
This PR adds a date count export plugin, allowing termgraph heatmaps : 
```bash
jrnl --export dates | termgraph --calendar
```

Closes #743.

### Checklist

  for the same issue.
- [x] All tests pass.


jrnl/plugins/util.py
```diff
@@ -3,6 +3,18 @@
 # Copyright (C) 2012-2021 jrnl contributors
 # License: https://www.gnu.org/licenses/gpl-3.0.html
 
+from collections import Counter
+
+
+def get_date_counts(journal):
```
wren: We've been trying to move everything out of this util file. Since this function is only used in the new plugin, can you move it into there?

KarimPwnz: I figured. On it!

jrnl/plugins/util.py
```diff
@@ -3,6 +3,18 @@
...
+    """Returns a collections.Counter object containing date counts"""
+    date_counts = Counter()
+    for entry in journal.entries:
+        # entry.date.date() gets date without time
+        date = str(entry.date.date())
```
wren: Maybe this should use the date format from the config file?

jrnl/plugins/datecount_exporter.py
```diff
@@ -0,0 +1,25 @@
...
+# encoding: utf-8
+# Copyright (C) 2012-2021 jrnl contributors
+# License: https://www.gnu.org/licenses/gpl-3.0.html
+
+from .text_exporter import TextExporter
+from .util import get_date_counts
+
+
+class DatecountExporter(TextExporter):
+    """This Exporter can lists the tags for entries and journals, exported as a plain text file."""
```
wren: Copy/paste error (still refers to tags plugin).

jrnl/plugins/util.py
KarimPwnz: Isn't the point of a datecount plugin to output the dates universally (so it can be piped elsewhere)? I think you discuss that in #743.

jrnl/plugins/datecount_exporter.py
KarimPwnz: Woops!

jrnl/plugins/datecount_exporter.py
```diff
@@ -0,0 +1,25 @@
...
+
+    names = ["datecount"]
```
wren: Is this the best name for this? The plugin that counts tags is just `tags` so maybe just `dates`? Or `count-by-date` as an more descriptive alternate?

features/format.feature
```diff
@@ -554,3 +554,13 @@ Feature: Custom formats
         | basic_encrypted |
         | basic_folder    |
         | basic_dayone    |
+
+    Scenario: Export date counts
+        Given we use the config "datecount.yaml"
+        When we run "jrnl --export datecount"
```
wren: The `--export` flag has been replaced by `--format` since v2.5. The export one still works, but might be removed in the distant future. Can you change this?

features/format.feature
```diff
@@ -554,3 +554,13 @@ Feature: Custom formats
         | basic_encrypted |
         | basic_folder    |
         | basic_dayone    |
+
+    Scenario: Export date counts
```
wren: This test should run for all journal types. See the test just above this one in the same file for an example.

features/format.feature
KarimPwnz: Ah! I was going off the example in #743. Let me do that.

KarimPwnz: Fixed.

features/format.feature
KarimPwnz: 👍

jrnl/plugins/datecount_exporter.py
KarimPwnz: I used the name suggested in #743, but `dates` would make more sense. What do you think?

jrnl/plugins/datecount_exporter.py
KarimPwnz: Fixed!

jrnl/plugins/util.py
KarimPwnz: Fixed!

features/format.feature
KarimPwnz: Done!

jrnl/plugins/datecount_exporter.py
wren: Yup, that's my fault for not updating an old issue. I think I like `dates` better.

KarimPwnz: Fair enough. I'll update the name shortly.

jrnl/plugins/datecount_exporter.py
```diff
@@ -0,0 +1,29 @@
...
+
+    @classmethod
+    def export_journal(cls, journal):
+        """Returns dates and their frequencies for an entire journal."""
+        date_counts = Counter()
+        for entry in journal.entries:
+            # entry.date.date() gets date without time
+            date = str(entry.date.date())
+            date_counts[date] += 1
+        result = "\n".join(f"{date}, {count}" for date, count in date_counts.items())
```
wren: Can we make this format match the output of `--tags` (using colons)? Termgraph support custom delimiters anyway, so it won't break anything with integration, and gives our output a more uniform look.

wren: Also, maybe we should respect the date format in the config file? It would give users more flexibility with how they use the plugin.

wren (changes requested):

jrnl/plugins/datecount_exporter.py
KarimPwnz: I can make it match the output of `--tags`, but I am not sure about respecting the config's date format. In #743, you call for this plugin to maintain output consistency. Have I misunderstood something?

wren: That a good point. I was thinking that since this command doesn't need to be piped through intermediate commands anymore that the added configurability could be useful. But I think you're right that it's best to stick to one format for now.

KarimPwnz: Thank you for clarifying 👍

KarimPwnz: > Can we make this format match the output of --tags (using colons)?

Hmm... I am not sure how standard colons are. But I can definitely make the change if consistency's necessary.

jrnl/plugins/datecount_exporter.py
KarimPwnz: Fixed.

jrnl/plugins/datecount_exporter.py
KarimPwnz: I think `--tags` uses colons because commas are already present to separate tags. Wouldn't that make it a special case? Again, I feel like using commas here is more standard (at-least `Termgraph` accepts that by default).

wren (approved):
Good points. Thank you!

PR #1148, opened 2021-01-04
Emphasize installing dependencies before testing

gustiando:
As a beginner in Python, it was unclear `poetry install` needed to be run before `make test` for instance., which resulted in a subjectively obscure error until I figured it out. This is an extremely minor proposal that from a newbie's perspective, ensuring one of the most appropriate command for installing dependencies is the very first instruction could help reduce some friction when considering contributing.

### Checklist

  for the same issue.
- [x] All tests pass.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -85,9 +85,9 @@ You can find an inventory of commands in the `makefile`. \*nix users can run the
 
 A typical development workflow includes:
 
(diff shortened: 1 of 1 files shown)


wren (approved):
It's very helpful to get these kinds of clarifying PRs for our documentation.

It can be difficult for us to remember what it was like starting out sometimes, so thank you for keeping us honest!

PR #1164, opened 2021-01-23
Make journal selection behavior more consistent when there's a colon with no date

### LINKED ISSUE #512: jrnl writes to default journal if a colon is used after a journal's name

wren:
Fixes #512

### Checklist

  for the same issue.
- [x] All tests pass.

### DIFF
--- a/features/multiple_journals.feature
+++ b/features/multiple_journals.feature
@@ -29,6 +29,20 @@ Feature: Multiple journals
         And journal "work" should have 1 entry
         And journal "work" should contain "2012-07-23"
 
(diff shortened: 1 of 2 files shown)


micahellison (approved):
💯

PR #1169, opened 2021-01-29
Add `--config-override` feature

### LINKED ISSUE #1157: Override (select) configuration values during invocation
### Use Case/Motivation

One that I can think of, which prompted the idea, was to be able to quickly launch a floating terminal with the built-in input prompt so I can log an entry and then Ctrl-D to exit. In my normal writing workflow I prefer to have a highly configured editor instance for writing with all the markdown bells and whistles. 

To accomplish this I propose that `jrnl` can be launched with an override option like `--override '{"editor": " "}'` that increases flexibility for specific workflows, while still preserving a configured setup for "main" writing workflows. 

### Example Usage

`jrnl --override '{"editor": " ", "linewrap": 40}'`
### Other information

I suppose this might be an orthogonal way to implement #1026 or #1068. 
I tried out a unittested prototype for this specific use-case in [my fork](https://github.com/sriniv27/jrnl), but I don't want to duplicate effort or step on anyone's toes; what are your thoughts?

sriniv27:
- What is this new code intended to do?
  - This submission is intended to provide a mechanism to override the stored configuration for the lifetime of the current invocation of `jrnl` by way of a  delimited list of key-value pairs. The keys can be provided in dot-notation form e.g. `colors.body` 
- Are there any related issues?
  #1157 
- What is the motivation for this change?
  I would like to run jrnl with different configurations for some specific workflows but still maintain my current configuration for "normal" writing workflows.
  For example:  
  - My default writing configuration is set so that I use `code -w` as my editor. For rapid logging, I want to be able open a terminal and be dropped into the stdin prompt, record my thought, and Ctrl+D out. 
  - In tmux have a separate, narrow, vertical pane that has the contents of my day's log, for which I might want a smaller `linewrap` value than the default
  - See discussion in #1157 for other examples/ideas
- What is an example of usage, or changes to config files? (if applicable)
   See [motivating comment](https://github.com/jrnl-org/jrnl/issues/1157#issuecomment-770296603)
  `jrnl -on today --config-override editor '' --config-override journals.log "~/logfile.txt"`
  `jrnl -on today --config-override linewrap 30 --config-override highlight false --config-override display_format fancy`
### Checklist

  for the same issue.
- [x] All tests pass.


wren (changes requested):
This is looking great!

I'd like to change the approach, so please see the ticket for my notes.

Thank you for putting so much work into this!

docs/advanced.md
```diff
@@ -61,6 +61,28 @@ and can be edited with a plain text editor.
   Or use the built-in prompt or an external editor to compose your
   entries.
 
+### Modifying Configurations from the Command line 
+
+You can override a configuration field for the current instance of `jrnl` using `--config-override CONFIG_KEY:CONFIG_VALUE` where `CONFIG_KEY` is a valid configuration field, specified in dot-notation and `CONFIG_VALUE` is the (valid) desired override value.
```
wren: This still uses the colon format from before. Can you update for the new syntax?

features/overrides.feature
```diff
@@ -0,0 +1,53 @@
+Feature: Implementing Runtime Overrides for Select Configuration Keys
+
+        Scenario: Override configured editor with built-in input === editor:''
+        Given we use the config "tiny.yaml"
```
micahellison: Could you use the basic_ configs for each journal type, like in tags.feature? Then remove tiny.yaml? It'd be nice to need keep the number of configs down.

features/overrides.feature
```diff
@@ -0,0 +1,53 @@
...
+        ┖─────────────────────┘
+        ┎─────╮2013-06-10 15:40
+        ┃ Lif ╘═══════════════╕
+        ┃ e is  goo d.        │
+        ┠╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
+        ┃ But I'm better.     │
+        ┖─────────────────────┘
+        """
+
+        @skip_win
```
micahellison: I removed this @skip_win on my Windows machine and this test is passing

features/overrides.feature
```diff
@@ -0,0 +1,53 @@
...
+        Scenario: Override color selections with runtime overrides 
+        Given we use the config "tiny.yaml"
+        When we run jrnl with -1 --config-override colors.body blue
+        Then the runtime config should have colors.body set to blue
+
+        @skip_win
```
micahellison: I do see that this test and the test below are failing in Windows, but I think it would be good to ensure they are working. Let me know if you'd like some help troubleshooting this.

features/steps/override.py
```diff
@@ -0,0 +1,123 @@
...
+                expected_config['journal'] ='features/journals/journal.jrnl'
+
+                assert mock_write_in_editor.call_count == 1
+                assert mock_write_in_editor.call_args[0][0]['editor']==editor
+        except SystemExit as e:
+            context.exit_status = e.code
+    # fmt: on
+
+
+@then("the stdin prompt must be launched")
```
wren: Can we change this wording to match the editor one? So, something like "the stdin prompt should have been called".

features/steps/override.py
```diff
@@ -0,0 +1,123 @@
...
+        return original_parse
+
+    def wrapper(input, *args, **kwargs):
+        input = context.now if input == "now" else input
+        return original_parse(input, *args, **kwargs)
+
+    return wrapper
+
+
+@given("we use the config {config_file}")
```
wren: Please try to re-use the functions in `core.py` when possible. We shouldn't copy them into a new file if we can avoid it.

features/steps/override.py
```diff
@@ -0,0 +1,123 @@
...
+    with open(context.config_path, "r") as f:
+        cfg = yaml.load(f, Loader=FullLoader)
+    context.cfg = cfg
+
+
+@then("the runtime config should have {key_as_dots} set to {override_value}")
+def config_override(context, key_as_dots: str, override_value: str):
+    key_as_vec = key_as_dots.split(".")
+
+    def _mock_callback(**args):
```
wren: Is this used anywhere?

features/steps/override.py
```diff
@@ -0,0 +1,123 @@
...
+        print("callback executed")
+
+    # fmt: off
+    try: 
+        with \
```
wren: Can we reduce the amount of mocks we put into this test? If we mock too many things, we lose the value of these tests (since we could be mocking out bugs).

features/steps/override.py
```diff
@@ -0,0 +1,123 @@
...
+@when("we run jrnl with {args}")
+def run_command(context, args):
+    context.args = split_args("%s" % args)
+    context.parser = parse_args(context.args)
+    with open(context.config_path, "r") as f:
+        cfg = yaml.load(f, Loader=FullLoader)
+    context.cfg = cfg
+
+
+@then("the runtime config should have {key_as_dots} set to {override_value}")
```
wren: This is a great idea for a test step!

Can it be merged with the other config step in `core.py`?

jrnl/override.py
```diff
@@ -0,0 +1,47 @@
+# import logging
+def apply_overrides(overrides: list, base_config: dict) -> dict:
```
micahellison, line 4: There is a similar `update_config` method in config.py that merges configuration of individual journals into the global config. Is there a reason to keep it separate?

jrnl/jrnl.py
```diff
@@ -49,6 +50,11 @@ def run(args):
             args=args, config=config, original_config=original_config
         )
 
+    # Apply config overrides
+    overrides = args.config_override
+    if overrides:
+        config = apply_overrides(overrides, config)
```
wren: This should be farther up in the file (right after line 40 `config = scope_config(config, args.journal_name)`) since there are commands that require the config. If the overrides stay here, then none of those commands (import, encrypt, decrypt, etc) will work with overrides.

pyproject.toml
```diff
@@ -38,15 +38,16 @@ pyyaml = ">=5.1"
...
 
+
 [tool.poetry.dev-dependencies]
 behave = "^1.2"
 mkdocs = "^1.0"
 black = {version = "^20.8b1",allow-prereleases = true}
 toml = ">=0.10"
 pyflakes = ">=2.2.0"
 pytest = ">=6.2"
+mock = "^4.0.3"
```
micahellison: Is there a reason for this? `mock` is in the standard library.

tests/test_config.py
```diff
@@ -0,0 +1,97 @@
...
+    mock_load_or_install,
+    mock_isatty,
+    minimal_config,
+    expected_override,
+    capsys,
+):
+    mock_load_or_install.return_value = minimal_config
+    mock_isatty.return_value = True
+
+    cli_args = shlex.split('--config-override editor "nano"')
```
micahellison: I think we should we be using os_compat split_args instead of shlex.split to be cross-platform

tests/test_config.py
```diff
@@ -0,0 +1,97 @@
...
+@mock.patch("sys.stdin.isatty")
+@mock.patch.object(install, "load_or_install_jrnl")
+@mock.patch("subprocess.call")
+def test_override_configured_editor(
+    mock_subprocess_call,
+    mock_load_or_install,
+    mock_isatty,
+    minimal_config,
+    expected_override,
+    capsys,
```
micahellison: I think this `capsys` is an extra argument here and in the next function

tests/test_config.py
```diff
@@ -0,0 +1,97 @@
...
+    exp_out_cfg = minimal_config.copy()
+    exp_out_cfg["editor"] = "nano"
+    exp_out_cfg["journal"] = "/tmp/journals/journal.jrnl"
+    yield exp_out_cfg
+
+
+@mock.patch("sys.stdin.isatty")
+@mock.patch.object(install, "load_or_install_jrnl")
+@mock.patch("subprocess.call")
+def test_override_configured_editor(
```
micahellison: This looks more like a behavorial test than a unit test - I believe it's already covered in the feature file. Ideally, the pytest unit test should be much smaller and testing individual functions or "units" of code, rather than large chunks of the program.

tests/test_config.py
```diff
@@ -0,0 +1,97 @@
...
+    exp_out_cfg["journal"] = "/tmp/journals/journal.jrnl"
+    yield exp_out_cfg
+
+
+@mock.patch("sys.stdin.isatty")
+@mock.patch(
+    "jrnl.install.load_or_install_jrnl", wraps=jrnl.install.load_or_install_jrnl
+)
+@mock.patch("subprocess.call")
+def test_override_configured_colors(
```
micahellison: This test has similar issues to the test above.

jrnl/override.py
```diff
@@ -0,0 +1,47 @@
...
+    :param base_config: Configuration Loaded from the saved YAML
+    :type base_config: dict
+    :return: Configuration to be used during runtime with the overrides applied
+    :rtype: dict
+    """
+    config = base_config.copy()
+    for pairs in overrides:
+
+        key_as_dots, override_value = list(pairs.items())[0]
+        keys = key_as_dots.split(".")
```
wren: Can the dot notation resolve logic be a standalone function? This will make it easier to test.

jrnl/args.py
```diff
@@ -17,6 +17,48 @@
...
+    """
+
+    assert len(input) == 2
+    runtime_modifications = {}
+
+    cfg_key = input[0]
+    cfg_value = input[1]
+    cfg_value = cfg_value.strip()
+
+    # Convert numbers and booleans
```
wren: Instead of doing our own type coercion here, we should rely on the yaml library we already use. What if we don't choose a type until we know where the value is going? Then we can match the type to the destination (which has already been parsed by the yaml library).

jrnl/args.py
```diff
@@ -17,6 +17,48 @@
...
+    if cfg_value.isdigit():
+        cfg_value = int(cfg_value)
+    elif cfg_value.lower() == "true":
+        cfg_value = True
+    elif cfg_value.lower() == "false":
+        cfg_value = False
+
+    runtime_modifications[cfg_key] = cfg_value
```
wren: Does this handle keys with dots?

jrnl/args.py
```diff
@@ -17,6 +17,48 @@
...
+
+    return runtime_modifications
+
+
+class ConfigurationAction(argparse.Action):
```
wren: Is this still needed?

wren (changes requested):
Getting there! Just a few more notes. Thank you!

micahellison (changes requested):
Thanks for all this work! I have a few questions and recommendations in my review.

features/steps/override.py
sriniv27: good catch. my bad

jrnl/jrnl.py
sriniv27: good to know, and thanks for catching that. Sounds like behavior I should have tested, so I can add some

features/steps/override.py
sriniv27: oh sure, thanks. I'll be more careful

jrnl/args.py
sriniv27: Yes, it stores the keys in dot notation. Those are unpacked into the config tree in apply_overrides

jrnl/args.py
sriniv27: It is. This is the action for --config-pverride argument. I intended to have an action for config modification operations, i.e. set, unset, and overrides.

features/overrides.feature
sriniv27: So I removed the tags and the tests fail in the CI runner with a the following UnicodeEncodeError: 

`UnicodeEncodeError: 'charmap' codec can't encode character '\U0001f4af'` which is the 100 emoji. That's about as far as I've been able to get so far. 

Here's the stack trace I'm examining: 
```
 File "c:\hostedtoolcache\windows\python\3.7.9\x64\lib\runpy.py", line 193, in _run_module_as_main
FAILURE in step 'the runtime config should have editor set to nano':
  Feature:  Implementing Runtime Overrides for Select Configuration Keys
  Scenario: Apply multiple config overrides
Exception UnicodeEncodeError: 'charmap' codec can't encode character '\U0001f4af' in position 533: character maps to <undefined>
    "__main__", mod_spec)
  File "c:\hostedtoolcache\windows\python\3.7.9\x64\lib\runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "D:\a\jrnl\jrnl\.venv\Scripts\behave.exe\__main__.py", line 7, in <module>
  File "d:\a\jrnl\jrnl\.venv\lib\site-packages\behave\__main__.py", line 183, in main
    return run_behave(config)
  File "d:\a\jrnl\jrnl\.venv\lib\site-packages\behave\__main__.py", line 127, in run_behave
    failed = runner.run()
  File "d:\a\jrnl\jrnl\.venv\lib\site-packages\behave\runner.py", line 804, in run
    return self.run_with_paths()
  File "d:\a\jrnl\jrnl\.venv\lib\site-packages\behave\runner.py", line 824, in run_with_paths
    return self.run_model()
  File "d:\a\jrnl\jrnl\.venv\lib\site-packages\behave\runner.py", line 626, in run_model
    failed = feature.run(self)
  File "d:\a\jrnl\jrnl\.venv\lib\site-packages\behave\model.py", line 350, in run
    formatter.eof()
  File "d:\a\jrnl\jrnl\.venv\lib\site-packages\behave\formatter\progress.py", line 86, in eof
    self.report_failures()
  File "d:\a\jrnl\jrnl\.venv\lib\site-packages\behave\formatter\progress.py", line 133, in report_failures
    self.stream.write(u"%s\n" % step.error_message)
  File "c:\hostedtoolcache\windows\python\3.7.9\x64\lib\encodings\cp1252.py", line 19, in encode
    return codecs.charmap_encode(input,self.errors,encoding_table)[0]
UnicodeEncodeError: 'charmap' codec can't encode character '\U0001f4af' in position 533: character maps to <undefined>
Error: Process completed with exit code 1.
```

jrnl/override.py
sriniv27: I'd only ever seen it used in the context of applying the --encrypt and --decrypt postconfig commands, but I don't see any argument against just calling update_config after unpacking . Lemme work on updating the implementation.

jrnl/override.py
sriniv27: can do

pyproject.toml
sriniv27: removed and adjusted import declaration

tests/test_config.py
sriniv27: I just ended up deleting the unittest file. The behavior is covered in the feature files

tests/test_config.py
sriniv27: good to know! For the purpose of this test, however, this is moot since I ended up deleting test_config.py altogether

jrnl/args.py
sriniv27: I'm forwarding the input to `yaml.load` and returning that output. Let me know if that is not the direction you'd like to take it and we can have a discussion on implementing it

features/steps/override.py
sriniv27: done!

wren:
@sriniv27 Hi! I see that you requested review, but there are still commits happening, and activity in the related issue. Is this ready for review?

It's fine either way, I just didn't want to comment on things if you're still working on them.

sriniv27:
I've only been making commits here to directly address review comments and smaller housekeeping things. I'd call this ready for review.

features/overrides.feature
```diff
@@ -0,0 +1,101 @@
...
+        Scenario: Override configured linewrap with a value of 23
+        Given we use the config "simple.yaml"
+        And we use the password "test" if prompted
+        When we run "jrnl  -2 --config-override linewrap 23 --format fancy"
+        Then the output should be
+
+        """
+        ┎─────╮2013-06-09 15:39
+        ┃ My  ╘═══════════════╕
+        ┃ fir st  ent ry.     │
```
wren, line 25: The `fancy` format looks like it has some issues. We need to fix this (not in this PR; just putting a note so we don't forget).

features/overrides.feature
micahellison: I created #1200 to troubleshoot, but it looks like all these passed in the CI this time. Could you try removing them and confirming?

features/overrides.feature
```diff
@@ -0,0 +1,101 @@
...
+        2013-06-10 15:40 Life is good.
+        | But I'm better.
+        """
+        
+
+        Scenario: Make an entry into an overridden journal 
+        Given we use the config "basic_dayone.yaml"
+        And we use the password "test" if prompted
+        When we run "jrnl --config-override journals.temp features/journals/simple.journal temp Sep 06 1969: @say Ni"
+       Then we should get no error
```
wren: Spacing issue.

jrnl/args.py
wren: I've been going back and forth on this, but I don't this deserialization should happen inside of arg parsing. At this point in the control flow, we don't have the config loaded yet, so we can't validate anything against it.

Can you move the function `deserialize_config_args` into config, and take out the arg parsing action?

jrnl/args.py
```diff
@@ -314,8 +349,32 @@ def parse_args(args=[]):
...
+        """,
+    )
+
     # Handle '-123' as a shortcut for '-n 123'
     num = re.compile(r"^-(\d+)$")
     args = [num.sub(r"-n \1", arg) for arg in args]
 
-    return parser.parse_intermixed_args(args)
+    parsed_args = parser.parse_intermixed_args(args)
+    return parsed_args
```
wren: Is this change doing anything?

pyproject.toml
```diff
@@ -38,6 +38,7 @@ pyyaml = ">=5.1"
 pytz = ">=2020"          # https://pythonhosted.org/pytz/#issues-limitations
 tzlocal = ">2.0, <3.0"   # https://github.com/regebro/tzlocal/blob/master/CHANGES.txt
 
+
```
wren: Can you remove this?

pyproject.toml
```diff
@@ -46,7 +47,6 @@ toml = ">=0.10"
 pyflakes = ">=2.2.0"
 pytest = ">=6.2"
 yq = ">=2.11"
-
```
wren: Can you put this back?

tests/test_override.py
```diff
@@ -0,0 +1,79 @@
...
+
+@pytest.fixture()
+def minimal_config():
+    cfg = {
+        "colors": {"body": "red", "date": "green"},
+        "default": "/tmp/journal.jrnl",
+        "editor": "vim",
+        "journals": {"default": "/tmp/journals/journal.jrnl"},
+    }
+    yield cfg
```
micahellison: Could you return instead of yield here? I don't see any iterating on this.

tests/test_override.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+def test_apply_override(minimal_config):
+    overrides = [{"editor": "nano"}]
+    apply_overrides(overrides, minimal_config)
+    assert minimal_config["editor"] == "nano"
+
+
+def test_override_dot_notation(minimal_config):
+    cfg = minimal_config.copy()
```
micahellison: No need to .copy() this -- the fixture is refreshed every time a test runs.

tests/test_override.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+def test_multiple_overrides(minimal_config):
+    overrides = [
+        {"colors.title": "magenta"},
+        {"editor": "nano"},
+        {"journals.burner": "/tmp/journals/burner.jrnl"},
+    ]  # as returned by parse_args, saved in parser.config_override
+
+    cfg = apply_overrides(overrides, minimal_config.copy())
```
micahellison: Also no need to .copy() here

tests/test_override.py
```diff
@@ -0,0 +1,79 @@
...
+    assert cfg["colors"]["body"] == "blue"
+
+
+def test_get_config_node(minimal_config):
+    assert len(minimal_config.keys()) == 4
+    assert _get_config_node(minimal_config, "editor") == "vim"
+    assert _get_config_node(minimal_config, "display_format") == None
+
+
+from jrnl.override import _get_key_and_value_from_pair
```
wren: Can you move this to the top of the file?

tests/test_override.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+def test_get_kv_from_pair():
+    pair = {"ab.cde": "fgh"}
+    k, v = _get_key_and_value_from_pair(pair)
+    assert k == "ab.cde"
+    assert v == "fgh"
+
+
+from jrnl.override import _convert_dots_to_list
```
wren: Can yo move this to the top of the file?

tests/test_override.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+class TestDotNotationToList:
```
wren, line 69: 🚀

micahellison (changes requested):
Looks really close to finished! Thanks again for your work on this.

wren (changes requested):
This is looking really close to ready!

Please add me to "reqeust review" when it's ready again. Thanks again for all the hard work.

tests/test_override.py
sriniv27: The intent of `yield` here is to return the result of the fixture setup. In pytest fixtures, teardown code follows the `yield` call. At least, that's how I interpreted [the doc page](https://docs.pytest.org/en/reorganize-docs/yieldfixture.html). 

That said, there isn't any teardown code here. So a `return` should be good enough.

features/overrides.feature
sriniv27: created #1207 so there's something to mark this resolved

sriniv27 (commented):
I think that's all the comments addressed

wren:
Awesome! I'll take a look again tonight.

micahellison (approved):
💯

wren (approved):
Looks great! Thank you! 🚀

PR #1199, opened 2021-02-23
update per-journal config documentation

### LINKED ISSUE #1196: Update documentation about journal-level config values
### Affected Page(s)

  - https://jrnl.sh/en/stable/advanced/

### What Could Be Better?

The section on multiple journal files makes a passing mention that a specific journal can have its own configured options, but the examples are only limited to preferences for `encrypt`. It is left to the reader to extrapolate that things like `timeformat`, `editor`, etc. can also be configured on a per-journal basis.  

### Other Information

I think that a couple of code snippets or even an extension of the existing snippet should suffice. 

Another (not necessarily exclusive) option is to update recipes with workflows that allude to it.

sriniv27:
Fixes #1196 
### Checklist

  for the same issue.
- [] I have written new tests for these changes, as needed.
- [x] All tests pass.

### DIFF
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -123,7 +123,8 @@ journals:
     journal: ~/recipes.txt 
 ```
 
-The `work` journal is encrypted and prints to `json` by default. It's also edited using an existing instance of VSCode. Similarly, the `food` journal prints to markdown by default.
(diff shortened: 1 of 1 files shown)


micahellison (approved):
Thanks for the PR! Looks great.

PR #1211, opened 2021-03-06
Fix broken brew release process

micahellison:
Our brew release process broke a while back due to us using an unsupported feature in `brew bump-formula-pr` which stopped working. This PR works around it.

See background in [Homebrew issue 10733](https://github.com/Homebrew/brew/issues/10733). Big thanks to @Rylan12 for support on this issue. I ended up working around this by tapping the formula, then working in the tap directory.

### Checklist

- [n/a] I have included a link to the relevant issue number.
- [n/a] I have tested this code locally.
  for the same issue.
- [ ] All tests pass.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -206,7 +206,7 @@ jobs:
         max_attempts: 6
         retry_wait_seconds: 30
         command: >
-          brew bump-formula-pr "Formula/${FORMULA_NAME}.rb"
+          brew bump-formula-pr "Formula/${FORMULA_NAME}"
           --url $(jq ".releases[\"${PYPI_VERSION}\"][1].url" -r api_response.json)
           --sha256 $(jq ".releases[\"${PYPI_VERSION}\"][1].digests.sha256" -r api_response.json)
           --version=$PYPI_VERSION
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -152,16 +152,10 @@ jobs:
           } >> $GITHUB_ENV
         fi
 
-    - name: Checkout homebrew repo
-      uses: actions/checkout@v2
-      with:
-        token: ${{ secrets.JRNL_BOT_TOKEN }}
-        repository: ${{ env.FORMULA_REPO }}
-
-    - name: Symlink repo into Homebrew's tap directory
+    - name: Tap formula
       run: |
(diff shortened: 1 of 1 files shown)


wren (approved):
🏭

.github/workflows/release.yaml
```diff
@@ -153,23 +152,26 @@ jobs:
...
-    - name: Checkout homebrew repo
-      uses: actions/checkout@v2
-      with:
-        token: ${{ secrets.JRNL_BOT_TOKEN }}
-        repository: ${{ env.FORMULA_REPO }}
+    - name: Tap formula
+      run: |
+        brew tap ${FORMULA_REPO}
+        echo '::debug::Set tap directory'
+        echo "BREW_TAP_DIRECTORY='$(brew --repository)/Library/Taps/${FORMULA_REPO}'" >> $GITHUB_ENV
```
Rylan12, line 159:
```suggestion
        echo "BREW_TAP_DIRECTORY='$(brew --repository ${FORMULA_REPO})'" >> $GITHUB_ENV
```

`brew --repository OWNER/TAP` or `brew --repository OWNER/homebrew-TAP ` will return the whole `$(brew --repository)/Library/Taps/OWNER/homebrew-TAP` path for you.

.github/workflows/release.yaml
```diff
@@ -202,7 +204,7 @@ jobs:
         max_attempts: 6
         retry_wait_seconds: 30
         command: >
-          brew bump-formula-pr "Formula/${FORMULA_NAME}.rb"
+          brew bump-formula-pr "${FORMULA_NAME}"
```
Rylan12, line 207: This should work fine. You could also use `"${FORMULA_REPO}/${FORMULA_NAME}"` to always ensure that you specify the right formula. On the off chance that a formula called `jrnl-beta` is added to Homebrew/core, this might no longer work as expected. I doubt that will happen, though, so probably not worth worrying about.

Rylan12 (approved):
This looks fine to me. The only thing I'd say is that the [Homebrew/actions/setup-homebrew](https://github.com/Homebrew/actions/tree/master/setup-homebrew) action will handle a lot of this for you. It may be worth taking a look to see if that's easier. If not, though, I think this will work just fine so I wouldn't worry about it.

Just a few comments:

PR #1219, opened 2021-03-21
More graceful handling of low linewrap values 

### LINKED ISSUE #1201: Fancy format crashes on low linewrap values
### Environment
  - `jrnl --diagnostic` output: 
```
jrnl: v2.7
OS: Windows 10
```
  - Install method:  source

### Current Behavior

When the `linewrap` config value is particularly low, the fancy formatting crashes with a stack trace that leads to the following error:
```
ValueError: invalid width -1 (must be > 0)
```

### Expected Behavior

jrnl shouldn't crash. It should be able to gracefully work around these restrictions with wrapping. Maybe part of the solution would be distancing the title and dates from each other so that they don't have to fight each other over the same horizontal length.

### Repro Steps

I created a default journal with the following config properties:
```
    timeformat: '%F %r'
    linewrap: 23
```
Then I ran:
```
jrnl -2 --format fancy
```

### Other Information

Full stack trace:
```
  File "C:\cache\pypoetry\virtualenvs\jrnl-g_ywnB0W-py3.8\Scripts\\jrnl", line 5, in <module>
    cli()
  File "C:\dev\jrnl\jrnl\cli.py", line 35, in cli
    return run(args)
  File "C:\dev\jrnl\jrnl\jrnl.py", line 66, in run
    search_mode(**kwargs)
  File "C:\dev\jrnl\jrnl\jrnl.py", line 172, in search_mode
  File "C:\dev\jrnl\jrnl\jrnl.py", line 327, in _display_search_results
  File "C:\dev\jrnl\jrnl\plugins\text_exporter.py", line 79, in export
    return cls.export_journal(journal)
  File "C:\dev\jrnl\jrnl\plugins\fancy_exporter.py", line 76, in export_journal
    return "\n".join(cls.export_entry(entry) for entry in journal)
  File "C:\dev\jrnl\jrnl\plugins\fancy_exporter.py", line 76, in <genexpr>
    return "\n".join(cls.export_entry(entry) for entry in journal)
  File "C:\dev\jrnl\jrnl\plugins\fancy_exporter.py", line 46, in export_entry
    title_lines = w.wrap(entry.title)
  File "c:\users\micah\.pyenv\pyenv-win\versions\3.8.2\lib\textwrap.py", line 354, in wrap
    return self._wrap_chunks(chunks)
  File "c:\users\micah\.pyenv\pyenv-win\versions\3.8.2\lib\textwrap.py", line 248, in _wrap_chunks
    raise ValueError("invalid width %r (must be > 0)" % self.width)
ValueError: invalid width -1 (must be > 0)

```

sriniv27:
This is intended to fix #1201 in the following manner:
1. Enforce a positive initial linewrap
2. If the the total linewrap is not big enough to accommodate the date string in its entirety, throw an error explaining so. 

I tried another direction where the date string is wrapped as well and the top left corner is extended downwards until all of the date can be chopped up and wrapped. It's pretty tricky to do gracefully so I'm not confident of checking that in yet. You can see the code [at the branch](https://github.com/sriniv27/jrnl/tree/format-fancy-low-linewrap) if you like.

### Checklist

  for the same issue.
- [x] All tests pass.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,3 +61,6 @@ coverage.xml
 .coverage
 .vscode/tasks.json
 todo.txt
+
+tags
+.vimrc
--- a/.gitignore
+++ b/.gitignore
@@ -61,6 +61,3 @@ coverage.xml
 .coverage
 .vscode/tasks.json
 todo.txt
-
-tags
-.vimrc
--- a/jrnl/exception.py
+++ b/jrnl/exception.py
@@ -33,8 +33,10 @@ class JrnlError(Exception):
             ),
             "LineWrapTooSmallForDateFormat": textwrap.dedent(
                 """
-                The provided linewrap value of {config_linewrap} is too small
-                to display the timestamps in the configured time format.
+                The provided linewrap value of {config_linewrap} is too small by {columns} columns
+                to display the timestamps in the configured time format for journal {journal}.
+
+                You can avoid this error by specifying a linewrap value that is larger by at least {columns} in the configuration file or by using --config-override at the command line 
                 """
             ),
         }
--- a/jrnl/exception.py
+++ b/jrnl/exception.py
@@ -30,7 +30,21 @@ class JrnlError(Exception):
 
                 Removing this file will allow jrnl to save its configuration.
             """
-            )
+            ),
+            "LineWrapTooSmallForDateFormat": textwrap.dedent(
+                """
+                The provided linewrap value of {config_linewrap} is too small
+                to display the timestamps in the configured time format.
+                """
+            ),
+            "LineWrapTooSmallForFancy": textwrap.dedent(
+                """
+                The provided linewrap value of {config_linewrap} is too small 
+                to properly format journal contents in fancy/boxed format. 
+                Either provide a larger value for the linewrap or display the
+                journal in another format.
+                """
+            ),
         }
 
         return error_messages[self.error_type].format(**kwargs)
--- a/jrnl/exception.py
+++ b/jrnl/exception.py
@@ -37,14 +37,6 @@ class JrnlError(Exception):
                 to display the timestamps in the configured time format.
                 """
             ),
-            "LineWrapTooSmallForFancy": textwrap.dedent(
-                """
-                The provided linewrap value of {config_linewrap} is too small 
-                to properly format journal contents in fancy/boxed format. 
-                Either provide a larger value for the linewrap or display the
-                journal in another format.
-                """
-            ),
         }
 
         return error_messages[self.error_type].format(**kwargs)
--- a/jrnl/plugins/fancy_exporter.py
+++ b/jrnl/plugins/fancy_exporter.py
@@ -91,5 +91,3 @@ def check_provided_linewrap_viability(linewrap, card, journal):
             columns=width_violation,
             journal=journal,
         )
-    else:
-        pass
--- a/jrnl/plugins/fancy_exporter.py
+++ b/jrnl/plugins/fancy_exporter.py
@@ -41,7 +41,8 @@ class FancyExporter(TextExporter):
         card = [
             cls.border_a + cls.border_b * (initial_linewrap) + cls.border_c + date_str
         ]
-        cls.check_linewrap(linewrap, card)
+        check_provided_linewrap_viability(linewrap, card, entry.journal)
+
         w = TextWrapper(
             width=initial_linewrap,
             initial_indent=cls.border_g + " ",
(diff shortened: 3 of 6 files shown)


wren:
Hi! Thanks for this. We were looking at this today, and it's looking good, but I want to do some more testing on it before doing a proper review.

.gitignore
```diff
@@ -61,3 +61,6 @@ coverage.xml
 .coverage
 .vscode/tasks.json
 todo.txt
+
+tags
+.vimrc
```
micahellison: Could you remove this? I think this might be better suited for your [global .gitignore](https://docs.github.com/en/github/getting-started-with-github/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer)

jrnl/exception.py
```diff
@@ -30,7 +30,13 @@ def _get_error_message(self, **kwargs):
...
                 Removing this file will allow jrnl to save its configuration.
             """
-            )
+            ),
+            "LineWrapTooSmallForDateFormat": textwrap.dedent(
+                """
+                The provided linewrap value of {config_linewrap} is too small
+                to display the timestamps in the configured time format.
+                """
+            ),
```
micahellison: I really like the use of this custom exception. A couple thoughts:

* Can we tell the user what the minimum allowable value is in this format?
* As with the other exception messages, we should also tell the user what they can do to fix it. Something like: "Modify your linewrap configuration or use --config-override to increase it."

jrnl/plugins/fancy_exporter.py
```diff
@@ -70,6 +75,13 @@ def export_entry(cls, entry):
         card.append(cls.border_l + cls.border_b * body_linewrap + cls.border_m)
         return "\n".join(card)
 
+    @classmethod
+    def check_linewrap(cls, linewrap, card):
+        if len(card[0]) > linewrap:
+            raise JrnlError("LineWrapTooSmallForDateFormat", config_linewrap=linewrap)
+        else:
+            pass
+
```
micahellison: I don't think this method needs to exist. Normally I like breaking things down into individual methods as much as possible, but it seems easier to read the conditional and the error raise (without needing the else: pass at all) in the calling method. Or, if the method does need to exist, maybe specify what is being checked.

jrnl/plugins/fancy_exporter.py
```diff
@@ -33,16 +36,18 @@ def export_entry(cls, entry):
         """Returns a fancy unicode representation of a single entry."""
         date_str = entry.date.strftime(entry.journal.config["timeformat"])
         linewrap = entry.journal.config["linewrap"] or 78
-        initial_linewrap = linewrap - len(date_str) - 2
+        initial_linewrap = max((1, linewrap - len(date_str) - 2))
```
micahellison, line 39: Would it make more sense here to check if `initial_linewrap` becomes negative and throw the error?

micahellison (changes requested):
Thanks for submitting this. Just have a few changes/clarifications.

.gitignore
sriniv27: Sure! Thanks for the link; learnt something new today

jrnl/plugins/fancy_exporter.py
sriniv27: hmmm... my only reason to break it up this way was to unittest it. That said, I agree that it doesn't need to be a *method*. I think it would make sense as a free function. 

I would prefer it stay as its own piece of logic for testability/reuse/maintainability.

I have updated the signature to better express intent.

jrnl/plugins/fancy_exporter.py
sriniv27: It'd make sense. I thought about it and decided to go the route I did thinking that we could still squeeze out some narrow linewrap values (I don't have a practical reason for it). 

That said, Simply enforcing that `initial_linewrap` is not negative simplifies the code flow, though I'll need to rework the exception since there is no `card` at that point in the execution. 

I have no strong preference, so give me a direction to go in after looking at the rest of the PR updates.

jrnl/plugins/fancy_exporter.py
micahellison: Those are all great points! Thanks for these changes.

jrnl/plugins/fancy_exporter.py
```diff
@@ -74,3 +80,16 @@ def export_entry(cls, entry):
...
+def check_provided_linewrap_viability(linewrap, card, journal):
+    if len(card[0]) > linewrap:
+        width_violation = len(card[0]) - linewrap
+        raise JrnlError(
+            "LineWrapTooSmallForDateFormat",
+            config_linewrap=linewrap,
+            columns=width_violation,
+            journal=journal,
+        )
+    else:
```
wren: Can we get rid of `else: pass`? It doesn't seem to do anything here.

jrnl/plugins/fancy_exporter.py
micahellison: Got it. I don't feel very strongly about this either, and I think it's pretty clear as is right now. Thanks for the explanation.

tests/test_export.py
```diff
@@ -0,0 +1,39 @@
...
+import pytest
+
+
+@pytest.fixture()
+def datestr():
+
+    yield "2020-10-20 16:59"
+
+
+from textwrap import TextWrapper
```
wren: Can you move this import statement to the top of the file with the rest of them?

sriniv27: I ended up deleting the import with the unused function `provide_date_wrapper`

wren (approved):
Looks great. Thank you!

micahellison (approved):
Thanks for your work on this. Looks great. 💯

PR #1229, opened 2021-04-14
Update absolute URLs to preview images in metatags

maebert:
When sharing on social media or most messengers, the preview image doesn't show because it now gets published to `en/stable/img/...`:

image

This fixes it so we can brag about jrnl in all its glory again:

image


wren (approved):
Good catch, and thank you!

PR #1233, opened 2021-04-24
Remove `--version` from brew release workflow

wren:
Homebrew doesn't like explicitly declaring a version for a formula, prefers to infer it from the url (as mentioned in homebrew/homebrew-core#67264). This will throw an error on PRs, so this PR removes that part of the workflow.


micahellison (approved):
☕

PR #1250, opened 2021-05-29
Ensure that line endings in all py files are Linux style instead of Windows

micahellison:
It looks like I left a few .py files with Windows line endings a while back. I'm just putting in this PR to make sure they are all using Linux line endings instead.

I know the diffs are hard to read, but they should be exactly the same except for the line endings. I used the `dos2unix` tool on all .py files to make this change.

### Checklist

  for the same issue.
- [x] All tests pass.


wren (approved):
lgtm 👾

Issue #1180, opened 2021-02-06
Move and expand contributing.md content into docs site


micahellison:
## Documentation Change

There is a lot going in our contributing.md file and we're thinking of adding more. Moreover, we have several different audiences for this file (docs writers, testers, and developers), and we might be able to reach more of them through the jrnl.sh site.

It seems like it would be a good move to create a new section in the documentation for contributors, with subsections for each of these audiences.

As part of this, the developer documentation could be expanded to include some information on the project's architecture and our process for pinning and resolving GitHub issues.

After this work is done, the contributing.md shouldn't be removed, but it should mainly just link to this new documentation.

PR #1265, opened 2021-06-19
[WIP] Edit feature plugin documentation

micahellison:
This builds on the namespace-based export plugins #1006. I want to do what I can to make sure that this documentation is relatively easy and approachable for new plugin developers.

Done so far:
- use same style as the rest of the documentation (informal "you", `jrnl` instead of *jrnl*, minimize passive voice)
- remove justification for change, since this is a how-to guide
- add reference to Python namespace package docs
- use tip/warning tooltip styles
- order documentation in order from general to specific

### Checklist

  for the same issue.
- [ ] All tests pass.

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -18,9 +18,12 @@ same Python interpreter that was used to install them.
 
 ## Getting Started
 
-Before you start working on a plugin, it's good to have a basic understanding of [Python packaging](https://packaging.python.org/guides/) and namespace packages in particular.
+Before you start working on a plugin, it's good to have a basic understanding
+of [Python packaging](https://packaging.python.org/guides/) and namespace
+packages in particular.
 
-You can find a sample plugin in the in the `/tests/external_plugins_src/` directory of the `jrnl` [source](https://github.com/jrnl-org/jrnl).
+You can find sample plugins in the `/tests/external_plugins_src/` directory of
+the [jrnl source](https://github.com/jrnl-org/jrnl).
 
 ## Understanding the Entry Class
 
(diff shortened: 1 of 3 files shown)


micahellison:
When I started working on this, I thought it would be a good sub-section of the new Contributing section in #1180 but after actually seeing it, I realized it makes a lot more sense to keep it separate. Writing your own plugin is a lot different than contributing directly to the jrnl project, and a lot of the contributing documentation isn't even related to coding.

In the long run, I think we'll want an additional user doc on installing plugins, so we end up with two main plugin docs. I made issue #1267 for that.

minchinweb:
Hey @micahellison ! Further to f2281497b309192292547eedafe87170e67cb793, I really think we should keep the sample plugins as part of the displayed documentation.

1. Keeping them in means that the (generated) documentation is self-contained, which means you can turn it in to a PDF and read it offline.
2. Keeping the plugin inline reduces context changes. My hope is that the plugin documentation is expansive and comprehensive enough that you could write your plugins based on the documentation alone, without having to context-swtich to find something elsewhere on the internet.

I specifically wrote as simple a complete plugin as I could with an eye of putting these in the documentation, and they come in at 40 and 50 lines. If these are two long, maybe we should put some work on slimming them down.

Hope my couple thoughts helps.

micahellison:
HI @MinchinWeb, funnily enough, I removed the included code because I found the context change of a page full of code really jarring. It might be appropriate for a tutorial, but I don't believe it's appropriate for a reference, except as an addendum rather than than something that interrupts the flow of an article. Splitting up the docs could help with this.

Also, I am still not feeling great about the example code. I understand your motivation for wanting to have a short-as-possible example of a plugin, but I'm far more concerned that providing an incomplete example as the bedrock for a whole plugin ecosystem may end up causing lots of difficulty for both us and the users. That's why I filed #1268, with the hopes that creating a sample plugin repo with some testing infrastructure would lead to well-tested plugins.

micahellison:
Fixed some merge issues but tests are failing now -- looks like it's due to linting issues on the feature-plugins branch.

micahellison:
Closing due to staleness. I'll get back to this if/when the feature-plugin branch stabilizes.

PR #1275, opened 2021-06-26
Remove `__init__.py` files

wren:
Since we're implementing a plugin system that relies on implicit namespace packages, we should remove these files so that they don't confuse and muddle our virtual envs (like the ones we use to run tests). Also, they're no longer needed as of Python 3.3.

PEP 420 says "Allowing implicit namespace packages means that the requirement to provide an __init__.py file can be dropped completely..."

### Checklist

  for the same issue.
- [x] All tests pass.


micahellison (approved):
🥳

Issue #1239, opened 2021-05-01
External formatter plugins


jayrod:
## Feature Request
Allow for external formatter plugins

### Use Case/Motivation
I would like to create formatting plugins for jrnl entries that include other services. Like an API rest backend or a database backend. The idea would be to export or "format" my data and send it off to another store. I see the plugins for the different types of formatters but I feel that this idea might not fit. It would be great if we could create external plugins that could be loaded as new formatters. 

My bigger use case is that I would like to use jrnl with many pen testers during an engagement and then aggregate it at the end  and then create a big report.

### Example Usage
jrnl --format database

'' Add sql uri in config ''

jrnl --format database
Data saved to database

### Other information

Other backends could be API.. splunk, mongo.. etc

jayrod:
I would also love to put in a PR for this if someone could point me in the right direction. My guess is that this enhancement would cause a large change to how formatters are loaded.

I guess... that they could be loaded at the beginning and checked when the --formatter flag is evaluated. 

not sure how to elegantly add such functionality.  I've done work like this using pluginlib however.

micahellison:
Hi there! Thanks for filing this. We've been considering adding format plugins as well, and we've been eyeing the [pluggy](https://pypi.org/project/pluggy/) library because we really like it how it's working for us in pytest, and it can be set up to couple really loosely with jrnl. This pluglinlib library looks nice too, but it takes a very class-based approach, and we intend to rework much of the existing jrnl classes in future versions, which could really wreck a potential plugin ecosystem using that approach.

If you're interested in adding a touch point just for formats, we'd be happy to work with you and review your PR when you have questions. Or, if you think there's a strong case to use a different framework, we'd love to hear your thoughts.

And you're absolutely right that implementing this would cause a large change to how formatters are loaded. It would probably need to happen every time the application is loaded, even, since the list of formats is provided in the `--help` output. But I think that's acceptable, given how we've seen this work while using pytest, specifically with the [icdiff](https://github.com/jeffkaufman/icdiff) plugin which is very similar to the kind of format plugin we're talking about.

minchinweb:
@jdavis @micahellison 

The framework for this is implemented by Pull Request #1216. Rather than a library for plugins, it only relies only on namespaces: so create you plugin in the `jrnl.contrib.exporter` namespace and *jrnl* will automatically load it once installed. See [details](https://github.com/jrnl-org/jrnl/pull/1115#issuecomment-745571224).

[...]

PR #1277, opened 2021-06-26
Fix testing for plugins

wren:
More work on #1006 and #1239.

Apparently, we don't run the test suite on PRs to any branch but develop, so tests won't run on this. I ran them here: https://github.com/wren/jrnl/actions/runs/975213282

This PR also makes tests run on all PRs so we don't have to do that anymore.

### Checklist

  for the same issue.
- [x] All tests pass.


micahellison (approved):
✨🔌🧪✨

PR #1278, opened 2021-06-27
Restore `__main__.py` functionality

minchinweb:
Allows you to access jrnl by running `python -m jrnl`

This was removed by #1275, but there is nothing in the notes about that pull request to suggest the was done purposefully.


wren (approved):
It looks like I got a bit delete-happy on that pr. Thanks for the fix!

PR #1279, opened 2021-06-27
Remove old GitHub issue template

minchinweb:
This seems to be superseeded by the `.github/ISSUE_TEMPLATE` folder and should no longer be needed.

### DIFF
--- a/issue_template.md
+++ b/issue_template.md
@@ -1,8 +1,0 @@
-# Stop
-
-Please don't file a blank issue.
-
-Fill out one of the templates from the link below and we'll be better able to
-help you.
-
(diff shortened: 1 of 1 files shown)


wren:
Oh, this file is fun times. It turns out it's still accessible if you go directly to https://github.com/jrnl-org/jrnl/issues/new. I'm not sure how people even get there, honestly (bookmarks, maybe?), but we kept getting blank issues filed a while back.

That's actually why this file is even here (if you look at the contents, it's just a note to please use our templates). So, I unfortunately think we need to keep this around for legacy reasons (at least until that url is no longer accessible).

PR #1280, opened 2021-06-27
Use relative version imports for plugins

minchinweb:
Further to #1275, which changed the import location of `__version__`.

The plugins are using the realitive import here on purpose. On the off chance that jrnl is installed twice (somehow) on a system, I want the plugin to pull it's version from it's jrnl, not from the global one.


wren:
@MinchinWeb I'm not sure I understand the scenario here. Can you come up with repro steps for where the current code might cause a problem?

PR #1193, opened 2021-02-14
Move test suite to Pytest (replace Behave)

wren:
Fixes #1192

Moves test suite entirely into pytest from behave.

PR #1281, opened 2021-06-27
Custom importer Improvements

minchinweb:
Further to #1006 

- explicitly write the journal to disk (after importing entries), rather than assuming the plugin will do this
- be more consistent in how data fed to the custom JSON importer is dealt with (i.e. pharse both input files and stdin into JSON)
- first pass at a test for the custom JSON importer. The test isn't running for me locally, but also isn't displaying any error message, so I'm not sure what's going on


jrnl/plugins/exporter/dates.py
```diff
@@ -7,7 +7,7 @@
 
 from jrnl.plugins.base import BaseExporter
 
-from jrnl.__version__ import __version__
+from ...__version__ import __version__
```
wren: Can you separate the relative imports here since there's already another PR for that change, please?

micahellison:
Hey @MinchinWeb, it looks like the test isn't running because it's being excluded by the `@skip_only_with_external_plugins` tag. The more I think about it, the more I'm hesitant to ask for a change on this, though, since it would need to be re-written anyway after merging #1193 in, and we're very close to finishing that.

features/plugins.feature
```diff
@@ -31,7 +31,8 @@ Feature: Functionality of Importer and Exporter Plugins
         Given We use the config "basic_onefile.yaml"
         When We run "jrnl --version"
         Then the output should contain pyproject.toml version
-        And The output should contain "<plugin_name> : <version> from jrnl.<source>.<type>.<filename>" 
+        And the output should contain "<plugin_name> : <version> from jrnl.<source>.<type>.<filename>"
```
wren, line 34: It's good that we're fixing up the tests, so this is fine. For future test fixes/updates, though, can we focus on the `/tests/bdd/features` directory instead?

features/plugins.feature
```diff
@@ -65,9 +66,15 @@ Feature: Functionality of Importer and Exporter Plugins
             | json        | <pyproject.toml version> | plugins | exporter | json     |
             | txt         | <pyproject.toml version> | plugins | exporter | text     |
 
+    @skip_only_with_external_plugins
+    Scenario Outline: Custom JSON Import
+        Given we use the config "empty_folder.yaml"
+        When we run "jrnl --import ./features/data/simple_import.json"
```
wren: No need to fix this here, but when we move this test to Pytest BDD, we should pipe this json in so it sits in the feature file (instead of having a separate file to manage).

tests/external_plugins_src/jrnl/contrib/importer/simple_json.py
```diff
@@ -27,13 +27,14 @@ def import_(journal, input=None):
...
             try:
-                data = sys.stdin.read()
+                f = sys.stdin.read()
             except KeyboardInterrupt:
                 print(
                     "[Entries NOT imported into journal.]",
                     file=sys.stderr,
                 )
                 sys.exit(0)
+            data = json.loads(f)
```
wren, line 37: 👍

wren (changes requested):
I added some more inline comments.

jrnl/plugins/exporter/dates.py
minchinweb: Rebased to drop this commit.

minchinweb:
> it looks like the test isn't running because it's being excluded by the `@skip_only_with_external_plugins` tag.

Except most of the other tests in the file are similarly tagged, and they do run....

As for #1193, it looks like that was merged to *develop*, so I'm not sure how best to move forward on that.

Otherwise, I think all the review comments have been resolved.

wren:
@MinchinWeb I updated the `feature-plugins` branch with the updates to develop, and then resolved the conflicts from this branch (since the merge conflict was pretty big, I didn't want you to have to deal with that). Anyway, thank you for addressing the feedback!

minchinweb:
@wren So what's left? Just to move the tests to the new pytest framework? (P.S. thanks for dealing with the merge conflict!)

micahellison:
Hey @MinchinWeb -- I just fixed the test, though it is failing now. There were a couple problems with it -- it was using the scenario outline, but had no examples, so behave iterated through the empty list and ran 0 tests from it. It also wasn't using the `--file` and `--format` keywords or checking for errors.

Also, could you modify the import test to take in new data? Currently, it imports data that's already in the journal defined by simple.yaml, so even if the import silently fails, it will still pass when it checks the contents at the end of the test:
```
        Then the journal should contain "My first entry."
        And the journal should contain "Life is good."
```

wren:
@MinchinWeb Yup! We just need to move the tests.

And to clarify what @micahellison is saying, the sample plugin doesn't quite work right now (you can test by installing it locally). So, fixing the test or modifying the test, it should hopefully go into pytest.

minchinweb:
P.S. Note the the Window scenarios in the CI aren't actually running tests...

I get about a dozen failing tests locally (on Windows) and the trackbacks aren't making a lot of sense....

micahellison:
Closing due to staleness. Feel free to re-open with new contributions.

PR #1282, opened 2021-06-27
Plugin list on Diagnostic

minchinweb:
Further to #1006, displays active plugins on `--diagnostic` rather than `--version`

### DIFF
--- a/jrnl/args.py
+++ b/jrnl/args.py
@@ -78,7 +78,7 @@ def parse_args(args=[]):
         action="store_const",
         const=preconfig_diagnostic,
         dest="preconfig_cmd",
-        help=argparse.SUPPRESS,
+        help="Print active plugins",
     )
     standalone.add_argument(
         "--list",
--- a/jrnl/commands.py
+++ b/jrnl/commands.py
@@ -17,16 +17,6 @@ import sys
 
 def preconfig_diagnostic(_):
     from jrnl import __version__
-
-    print(
-        f"jrnl: {__version__}\n"
-        f"Python: {sys.version}\n"
-        f"OS: {platform.system()} {platform.release()}"
-    )
-
-
-def preconfig_version(_):
-    from jrnl import __version__
     from jrnl.plugins.collector import (
         IMPORT_FORMATS,
         EXPORT_FORMATS,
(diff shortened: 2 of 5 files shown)


wren:
I forgot to comment on this after our conversation in #1006, sorry!

This PR is dependent on what we decide to do with versions numbers for internal plugins, so I don't think we should merge this yet. I'm going to mark it as a draft for now until that conversation is finished.

minchinweb:
Not stale.

Suggested format (from [here](https://github.com/jrnl-org/jrnl/discussions/1310#discussioncomment-1352192)):

```
Formats:
    Import:
        jrnl    : jrnl (internal)
    Export:
        boxed   : fancy (internal)
        default : custom-default v1.1.0 from foo.bar.somewhere_else
        foo     : bar v1.3
```

1. "Formats", "Import, and "Export in the titles so it's more consistent with the docs
2. Include the name of the plugin even if it's internal
3. Only include the `from foo.bar.whatever` if the plugin is not at the default location (so there's less noise for the defaults)
4. Line up the colons so it's more consistent with other general output

minchinweb:
Output looks like:

```
λ jrnl --diagnostic
jrnl: v2.8.1
OS: Windows 10

Active Plugins:
    Importers:
        jrnl     : jrnl (internal)
        json     : simple_json v1.0.0
    Exporters:
        boxed    : fancy (internal)
        dates    : dates (internal)
        default  : pretty (internal)
        fancy    : fancy (internal)
        json     : custom_json v1.0.0
        markdown : markdown (internal)
        md       : markdown (internal)
        pretty   : pretty (internal)
        rot13    : rot13 v1.0.0
        short    : short (internal)
        tags     : tag (internal)
        test     : flag v0.0.1
        testing  : flag v0.0.1
        text     : text (internal)
        txt      : rot13 v1.0.0
        xml      : xml (internal)
        yaml     : yaml (internal)
```

Tests appear to be failing because the `skip` tags aren't working, but that is beyond the scope of this PR (adjusting plugin info display). Also, Windows tests don't appear to run at all.

---

I think it's ready to merge.

wren (approved):
Looks good. Thank you for making the updates!

PR #1285, opened 2021-07-03
Change all YAML FullLoader calls to SafeLoader

micahellison:
The pyyaml project has some security issues with its FullLoader. It's not a huge deal for jrnl users, given that it's from the config file rather than untrusted input (such as user supplied data to a web server), but since all the tests are passing with SafeLoader, there's no harm in switching to this.

### Checklist

  for the same issue.
- [X] All tests pass.

### DIFF
--- a/features/steps/core.py
+++ b/features/steps/core.py
@@ -17,7 +17,7 @@ import keyring
 
 import toml
 import yaml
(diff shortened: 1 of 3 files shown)


wren (approved):
🥔

PR #1296, opened 2021-07-09
`jrnl.__version__` magic works!

minchinweb:
Further to #1006, this allows `from jrnl import __version__` to work even without an `__init__.py` file!

Turns out pretty much everything in Python is mutable, you just have to know what to mutate! :)

Also, apply *isort* to the adjusted imports.

This also tries to address the [release workflow](https://github.com/jrnl-org/jrnl/blob/develop/.github/workflows/release.yaml#L83) which will override the version file. However, bash is **not** my strong suit, so confirm my function will work as expected.

### Checklist

  for the same issue.
- [ n/a ] I have written new tests for these changes, as needed.
- [x] All tests pass.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -82,15 +82,14 @@ jobs:
         poetry version "$JRNL_VERSION"
         {
           echo "# This file is managed automatically by the GitHub release flow"
-          echo ""
+          echo
           echo "import sys"
-          echo ""
+          echo
           echo "__version__ = \"$JRNL_VERSION\""
-          echo ""
-          echo "# this makes the version available at `jrnl.__version__` without requiring a"
-          echo "# `__init__.py` file in the *jrnl* root directory!"
-          echo "sys.modules[\"jrnl.__version__\"] = __version__"
-          echo ""
+          echo
+          echo '# this makes the version available at `jrnl.__version__` without requiring a'
+          echo '# `__init__.py` file in the *jrnl* root directory!'
+          echo 'sys.modules["jrnl.__version__"] = __version__'
         } > jrnl/__version__.py
 
     - name: Commit updated files
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -80,7 +80,18 @@ jobs:
       if: ${{ github.event.inputs.include_repo_version == 'true' }}
       run: |
         poetry version "$JRNL_VERSION"
-        echo __version__ = \"$JRNL_VERSION\" > jrnl/__version__.py
+        {
+          echo "# This file is managed automatically by the GitHub release flow"
+          echo ""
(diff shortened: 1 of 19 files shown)


wren:
Converting to draft until it's ready for review (the tests pass). I'll also add a few quick comments inline shortly.

jrnl/__version__.py
```diff
@@ -1 +1,9 @@
+# This file is managed automatically by the GitHub release flow
+ 
+import sys
+
 __version__ = "v2.8.1"
+
+# this makes the version available at `jrnl.__version__` without requiring a
+# `__init__.py` file in the *jrnl* root directory
+sys.modules["jrnl.__version__"] = __version__
```
wren, line 9: Nice! 🚀

.github/workflows/release.yaml
```diff
@@ -80,7 +80,7 @@ jobs:
       if: ${{ github.event.inputs.include_repo_version == 'true' }}
       run: |
         poetry version "$JRNL_VERSION"
-        echo __version__ = \"$JRNL_VERSION\" > jrnl/__version__.py
+        echo -e "# This file is managed automatically by the GitHub release flow\n\nimport sys\n\n__version__ = \"$JRNL_VERSION\"\n\n# this makes the version available at `jrnl.__version__` without requiring a\n# `__init__.py` file in the *jrnl* root directory\n\nsys.modules[\"jrnl.__version__\"] = __version__" > > jrnl/__version__.py
```
wren: Instead of putting this all on one line, can you make this a bit more readable with something like this?

```yaml
{
echo "one"
echo "two"
echo "three"
} > jrnl/__version__.py
```

wren: Also, watch out for the double angle brackets (`>>`). This concatenates to a file instead of overwriting it, so this currently would keep bloating up the file forever instead of updating it.

wren:
Would you mind leaving isort out of this PR? It's making it a bit difficult for me to sort through the noise to the actual changes.

wren:
Thanks for the PR and please mark this as "Ready for review" when you've made updates.

.github/workflows/release.yaml
minchinweb: I've adjusted this to the multiline format you recommend. I really don't have a way to test this bash part,so let me know if it needs updating...

minchinweb:
I've made the "line" in the GitHub flow a multiline "echo", but I really don't have a way to test it on my side.

I've also pulled any isort changes not directly related to this change in how to import `__version__`. I actually surprised that the CI linting tests pass...

wren (approved):
Looks good. Thank you!

Note: Tests are failing because I updated the base branch with the new test suite (pytest-bdd), so we can ignore for now, since this isn't going into `develop`, yet.

PR #1301, opened 2021-07-10
Add documentation about saved passwords in Windows

### LINKED ISSUE #1142: Document keyring security concerns on Windows, Mac
## Documentation Change

I've been playing around with the keyring functionality on Windows and I thought it might be worth documenting a few of its behaviors. At first, I thought Windows didn't have a keyring at all, but it turns out that the keyring module we use interfaces with Windows Credential Manager and is "working" just fine with jrnl.

jrnl was developed with the Mac's keyring in mind. Mac users have a keyring that, by default, prompts for credentials before a program can write or read stored keys. Windows, however, simply allows any program you're running to access Windows Credential Manager, with no locking mechanism, no ability to allow the user to refuse access, and no Windows prompt whatsoever. Worse, it even allows access to data saved by other programs.

So, for Windows users, saving a jrnl password to the keyring is not very secure, and that password could be retrieved by a relatively simple script running as the user.

### Affected Page(s)

* Maybe this page, which could some notes on cross-platform behavior: https://jrnl.sh/en/stable/encryption/
* Definitely this page: https://jrnl.sh/en/stable/privacy-and-security/

### What Could Be Better?

Windows users should be warned, at the very least, about this issue, and told where to find the credentials in the Credential Manager. Maybe we should even modify the program itself to warn users (if so, let's spin that off to a separate issue).

### Other Information

* [StackExchange thread lambasting Windows Credential Manager](https://security.stackexchange.com/questions/119765/how-secure-is-the-windows-credential-manager)

micahellison:
Fixes #1142 from a Windows perspective. I can't speak much to the Mac environment, though, and would appreciate any suggestions from someone who can.

### Checklist

  for the same issue.
- [x] All tests pass.

### DIFF
--- a/docs/privacy-and-security.md
+++ b/docs/privacy-and-security.md
@@ -86,6 +86,18 @@ you have a journal, where that journal file is, and when you last edited it.
 With a sufficient power imbalance, someone may be able to force you to unencrypt
 it through non-technical means.
 
+## Saved Passwords
+
+When creating an encrypted journal, you'll be prompted as to whether or not you
+want to "store the password in your keychain." This keychain is accessed using
+the [Python keyring library](https://pypi.org/project/keyring/), which has different
+behavior depending on your operating system.
+
+In Windows, the keychain is the Windows Credential Manager (WCM), which can't be locked
+and can be accessed by any other application running under your username. If this is
+a concern for you, you may not want to store your password.
+
+
 ## Notice any other risks?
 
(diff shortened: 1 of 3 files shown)


wren:
I'm not sure why this is failing in CI. Looks like something is up with the local server we use?

micahellison:
Fixed - added back in a missed local dep (yq), which includes xq, which is used to search the sitemap XML (for pa11y I think?).

wren (approved):
🧪

PR #1313, opened 2021-07-31
Allow emoji in config file in Windows by always opening it as unicode

### LINKED ISSUE #1162: Journals with emoji names can't be accessed on Windows
### Environment
  - `jrnl --diagnostic` output: 
```
jrnl: v2.7-beta
OS: Windows 10
```
  - Install method:  source, running through `poetry shell`
  - This is specifically using the new-ish Windows Terminal, not the old Command Prompt. The Command Prompt just crashes for me when I send emojis to jrnl through it, but it looks like an issue with Windows or my Python configuration, and not jrnl.

### Current Behavior

In my `jrnl.yaml` config file, I defined a new journal using an emoji in the name, like so:
```
  😀:
    journal: C:\Users\micah\.local\share\jrnl\😀.txt
```
However, I can't see the emoji properly when running `jrnl --list`, nor can I write to it, even though the emoji name seems to be correctly passed from the Windows command line to journal.

### Expected Behavior

* I should be able to see the emoji properly in the journal name when running `jrnl --list`
* I should be able to write a new entry in the emoji-named journal

### Repro Steps

With the configuration specified above, you can see the name of the journal has been mangled (perhaps by the config parser?).

```
C:\dev\jrnl>jrnl --list
Journals defined in C:\Users\micah\.config\jrnl\jrnl.yaml
 * default  -> C:\Users\micah\.local\share\jrnl\journal.txt
 * ðŸ˜€     -> C:\Users\micah\.local\share\jrnl\ðŸ˜€.txt
```

The actual Windows Terminal mangles the emoji when I'm entering it, but it's still valid unicode and is passed to jrnl eventually... it's just that jrnl doesn't recognize it as a journal, and instead writes it to my default journal. This first line should be opening up an editor window to write a new entry in the 😀 journal.
```
C:\dev\jrnl>jrnl ��
C:\dev\jrnl>jrnl -1
2021-01-23 14:13 😀
```

### Other Information

I asked @wren about how this works on the Mac, and it doesn't seem to be an issue there. I suspect that our config/YAML parser has some OS-specific behavior.

micahellison:
Fixes #1162. I forced utf-8 encoding on open, since apparently Python won't use that in Windows by default.

### Checklist

  for the same issue.
- [x] All tests pass.

### DIFF
--- a/jrnl/config.py
+++ b/jrnl/config.py
@@ -20,6 +20,7 @@ DEFAULT_JOURNAL_NAME = "journal.txt"
 DEFAULT_JOURNAL_KEY = "default"
 
 YAML_SEPARATOR = ": "
+YAML_FILE_ENCODING = "utf-8"
 
 
 def make_yaml_valid_dict(input: list) -> dict:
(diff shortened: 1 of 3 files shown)


wren (approved):
✨

PR #1314, opened 2021-07-31
Fix failure to import into directory journal

### LINKED ISSUE #999: Importing to directory-based journals doesn't do anything, and has incorrect message
### Environment
  - Jrnl version:  v2.4.3
  - Install method:  pipx
  - OS  macos 10.15.3

### Current Behavior

```
$ j -2 --export text | j dirtest --import
[2 imported to default journal]
```

The above should be importing to the `dirtest` journal, but it says it's importing to default (it's actually not importing anywhere).

### Expected Behavior

Users should be able to import journal entries to directory-based journals the same way they can to single-file journals.

### Repro Steps

1. Make a directory-based journal (create an empty directory, and add to config as `dirtest` or whatever)
2. Add an entry to it (e.g. `jrnl dirtext some new entry`) to make sure it works
3. Try to import any number of entries to it from another journal (`jrnl --export text | jrnl dirtest --import`, or `jrnl --import -i journal_file.txt`)
4. Observe the message saying entries were imported to default journal
5. Observe entries were not actually imported to either journal (`jrnl -1` and `jrnl dirtest -1` or however you want to try to find those entries)

micahellison:
Fixes #999.

A couple issues were at fault:
- the FolderJournal only writes entries with the "modified" property set to True. However, imported entries weren't marked as "modified"
- the FolderJournal writing method wasn't identifying new ("modified") dates appropriately

There are some larger issues with this whole object model, mainly that it's not really hiding complexity. Whether you're calling or inheriting a Journal, you've got to know how its innards work, and that makes it difficult to extend. But first and foremost, I wanted to get this bug fixed.

### Checklist

  for the same issue.
- [x] All tests pass.

### DIFF
--- a/jrnl/FolderJournal.py
+++ b/jrnl/FolderJournal.py
@@ -22,10 +22,10 @@ def get_files(journal_config):
 class Folder(Journal.Journal):
     """A Journal handling multiple files in a folder"""
 
-    def __init__(self, **kwargs):
+    def __init__(self, name="default", **kwargs):
         self.entries = []
         self._diff_entry_dates = []
-        super(Folder, self).__init__(**kwargs)
+        super().__init__(name, **kwargs)
 
     def open(self):
         filenames = []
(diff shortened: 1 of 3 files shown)


wren (approved):
📦

viegasfh:
Regarding this issue, I would to add that even though the import wasn't working in the command line, there is a workaround:

1. Do a jrnl nameofjournal --edit (or -n 1 --edit) to edit in your text editor
2. Paste the entries from the text based journal (in vim do an insert text file)
3. Save and exit

After you save, you will find a message like the following:

```
[-44 entries deleted, 44 entry modified]
```
I don't know why it says -44 entries deleted, because it had none of them before. So, it should only say 44 entries added.

Doing a jrnl nameofjournal --export pretty, shows all the entries.

So, maybe this fix could also improve the message displayed so that it says the correct thing in cases where we paste the entries directly in the editor.

PR #1328, opened 2021-08-21
Fix deletion of entries on folder journal through `--delete` flag

### LINKED ISSUE #1316: Entries from directory based journals do not get deleted
### Environment
  - `jrnl --diagnostic` output: 
  - Install method: 
jrnl: v2.8.1
Python: 3.8.10 (default, Jun  2 2021, 10:49:15)
[GCC 9.4.0]
OS: Linux 4.4.0-19041-Microsoft

Installed using pip

### Current Behavior

Today I added an entry to one of my directory journals by mistake and when I tried to delete the entry, it was just not deleting it.

### Expected Behavior

When running the command to delete entries, it is expected that entries for both file based and directory based journals should be removed uppon confirmation. Unfortunately, only file based entries are deleted.

### Repro Steps

1) Add the entry to the directory based journal
2) Try to remove the entry with jrnl two -n 1 --delete (here I am using two as directory based journal)
3) List the last entry and it should still be there

### Other Information

There isn't really much to be added here. It is just silent. You expect the entry to be removed, but it isn't.

micahellison:
Fixes #1316. The deletion code wasn't using FolderJournal's special mechanism of tracking the dates of each entry modified.

### Checklist

  for the same issue.

### DIFF
--- a/jrnl/FolderJournal.py
+++ b/jrnl/FolderJournal.py
@@ -83,8 +83,14 @@ class Folder(Journal.Journal):
                 # print("empty file: {}".format(filename))
                 os.remove(filename)
 
+    def delete_entries(self, entries_to_delete):
+        """Deletes specific entries from a journal."""
+        for entry in entries_to_delete:
+            self.entries.remove(entry)
+            self._diff_entry_dates.append(entry.date)
+
     def parse_editable_str(self, edited):
-        """Parses the output of self.editable_str and updates it's entries."""
+        """Parses the output of self.editable_str and updates its entries."""
         mod_entries = self._parse(edited)
         diff_entries = set(self.entries) - set(mod_entries)
         for e in diff_entries:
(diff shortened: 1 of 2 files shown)


jrnl/FolderJournal.py
```diff
@@ -83,8 +83,14 @@ def write(self):
...
 
+    def delete_entries(self, entries_to_delete):
+        """Deletes specific entries from a journal."""
+        for entry in entries_to_delete:
+            self.entries.remove(entry)
+            self._diff_entry_dates.append(entry.date)
+
     def parse_editable_str(self, edited):
-        """Parses the output of self.editable_str and updates it's entries."""
+        """Parses the output of self.editable_str and updates its entries."""
```
wren, line 93: 💥

wren (approved):
␡ 📝

PR #1150, opened 2021-01-06
Add delimiters in YAML format

Seopril:
This is an intended fix for issue #1065. The yaml export now functions as shown in the "Expected Behavior" described in that bug report. 

### Example Output:

```
---
title: Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod
date: 2021-01-05 11:26
starred: False
tags: 
...

Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.

Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.
```

### Checklist

- [X] I have read the [contributing doc](https://github.com/jrnl-org/jrnl/blob/develop/CONTRIBUTING.md).
- [X] I have included a link to the relevant issue number.
- [X] I have tested this code locally.
- [X] I have checked to ensure there aren't other open [pull requests](../pulls)
  for the same issue.
- [X] I have written new tests for these changes, as needed.
- [X] All tests pass.

Issue #1065, opened 2020-10-27
YAML Export lacking delimiters


marekkowalczyk:
### Environment
  - jrnl: v2.4.5 Python: 3.8.6 (default, Oct  8 2020, 14:07:53) [Clang 11.0.0 (clang-1100.0.33.17)] OS: Darwin 18.7.0

  - Install method: brew

### Current Behavior

```
title: Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod
date: 2020-03-19 09:39
starred: False
tags: 
 
Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.

Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.

```

### Expected Behavior

````
---
title: Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod
date: 2020-03-19 09:39
starred: False
tags: 
...

Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.

Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.
 
````

Alternatively:

````
---
title: Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod
[...]

Issue #835, opened 2020-02-08
YAML export does not export proper YAML or correctly spelled "starred" field


micahellison:
## Bug report

  - Jrnl version: 2.2
  - How you installed Jrnl: pip install jrnl
  - Operating system [MacOS, Linux, Windows?]: Windows 10

### Steps to repro:
```
jrnl --export yaml -o .\jrnltest
```

### Expected results:
 * Export should be in YAML format
 * Fields should be spelled correctly

### Actual results:
 * The output has YAML in it, but the entry contents are not encapsulated in a YAML field; instead there's just a space before the entry.
 * The "starred" field is spelled "stared" 
```
title: Test
date: 2020-02-08 12:05
stared: False
tags:
 So, I was testing jrnl today, and I was wondering, what happens when I export to YAML?
```
Uncovered while reviewing the tests in PR #824.

### Background/discussion
The YAML export's history is rooted in Day One compatibility (#159, #773, #349). That's fine, and it looks like we may need to preserve this behavior as a Day One export, but calling it a YAML export means that anybody expecting to extract/transform/load their jrnl anywhere else via YAML is going to have a bad time.

Maybe we rename the export type to something more appropriate to its intended function. Or, if the original target of this export can take proper YAML, then maybe we rework it. @MinchinWeb, any thoughts?

wren:
Yup! I'm in favor of renaming it to clarify that it's DayOne-related. 

@MinchinWeb Your opinion?

MinchinWeb:
The YAML export is designed to support DayOne (although I hope all exporters will do so), but it's actual roots are a modification of the Markdown exporter to include the entry metadata in a format useful as an input format for [Pelican](https://github.com/getpelican/pelican), a static site generator. This is why the files export with a `.md` file extension.

I'm fine with renaming it, but a more appropriate name would be along the lines of "Markdown with YAML front matter" or "Pelican Markdown"; i.e. nothing to do with DayOne.

---

"stared" appears to be an spelling mistake from when the exporter was [first added](https://github.com/jrnl-org/jrnl/commit/f7cadfc9dd13c5809f7049ea6e3176a3f08e67c6); no reason to keep the error.

MinchinWeb:
The (original) discussion about why the exporter is named the why it is is found in this pull request --> https://github.com/jrnl-org/jrnl/pull/258

micahellison:
This was fixed by #907

PR #258, opened 2014-07-30
Export to Pelican

MinchinWeb:
This allows you to export your JRNL to Pelican formatted Markdown. The exported files have the necessary metadata at the top, and then contain the body of the entry. Files are saved with `.md` extension.

These can then be used as source content to generate the files of a Pelican blog.

Example file:

```
Title: RAM?
Date: 2013-07-30 12:02:00
Tags: computer, RAM

8GB of RAM for my desktop??...
```

PR #1337, opened 2021-09-11
Restore (old) YAML exporter as "Pelican Markdown"

minchinweb:
As noted in the original implementation (by me, see #258, from 2014), the "Yaml" exporter was never intended as a "pure" YAML exporter but rather as an exporter with YAML front matter and a Markdown body so the exported files could be used as source files for [Pelican](https://github.com/getpelican/pelican). This was re-iterated in Issue #835.

However, the issue (of the YAML exporter not being a "pure"/"true" YAML exporter) was re-raised in #1065, specifically that the "Yaml" exporter wasn't emitting valid YAML (as pointed out above, it was never intended to). That issue was resolved (by #1150, released with v2.7) by making the export body valid YAML, but it actually created new issues:

- why is the YAML exporter writing `.md` (or Markdown) files (by extension)?
- why is the YAML exporter processing Markdown headers in the body text?
- what is the replacement for the Pelican YAML/Markdown export that the "Yaml" exporter originally supplied?

<strike>This pull request fixes these new issues:

- the "old" YAML exporter is now available as the "pelican-markdown" exporter
- the "pure" YAML exporter now emits `.yaml` files
- the "pure" YAML exporter doesn't assume the body is Markdown and so doesn't adjust Markdown headers anymore

Note that this is a **Breaking Change** becuase #1150 is/was a breaking change (i.e the output format of the YAML exporter has significantly changed, and can no longer be used as Pelican source files). It is also a breaking change as the filenames of the exported files (from the "pure" YAML exporter) has changed (see https://github.com/jrnl-org/jrnl/pull/1337/commits/04bfa7e7d841ea0362f487c5e6c0ea68ad529847)</strike>

This pull request brings back the old "YAML" exporter (which I use basically daily), renaming it as "Pelican Markdown" (to be entirely explicit about it's intended purpose).

(Fixes for the new/pure YAML exporter will be under a different Pull Request (eventually).)

### Checklist

  for the same issue.


features/format.feature
```diff
@@ -446,11 +446,11 @@ Feature: Custom formats
...
         """
-        2020-08-29_entry-the-first.md
-        2020-08-31_a-second-entry-in-what-i-hope-to-be-a-long-series.md
-        2020-09-24_the-third-entry-finally-after-weeks-without-writing.md
+        2020-08-29_entry-the-first.yaml
+        2020-08-31_a-second-entry-in-what-i-hope-to-be-a-long-series.yaml
+        2020-09-24_the-third-entry-finally-after-weeks-without-writing.yaml
         """
-        And the content of file "2020-08-29_entry-the-first.md" in the cache should be
+        And the content of file "2020-08-29_entry-the-first.yaml" in the cache should be
```
micahellison: Could you implement your tests in the pytest-bdd instead of behave? We're trying to phase out behave. See `/tests/bdd`

jrnl/plugins/__init__.py
```diff
@@ -16,6 +17,7 @@
 __exporters = [
     JSONExporter,
     MarkdownExporter,
+    PelicanMarkdownExporter,
```
micahellison, line 19: Since Markdown front matter is used by a lot more stuff than just Pelican, I think we should name it markdown-front-matter (and thus name this class MarkdownFrontMatterExporter).

jrnl/plugins/pelican_markdown_exporter.py
```diff
@@ -0,0 +1,139 @@
...
+                """Setext style H2"""
+                newbody = newbody + heading + "## " + previous_line
+                line = ""
+            elif multi_tag_regex.match(line):
+                """Tag only lines"""
+                line = ""
+            else:
+                newbody = newbody + previous_line
+            previous_line = line
+        newbody = newbody + previous_line  # add very last line
```
micahellison: I like the commenting and general readability of this section, but it looks like a good candidate to be broken out into its own function. The overall flow of the exporter would be easier to read, and the specific  function could be unit-tested. What do you think?

jrnl/plugins/pelican_markdown_exporter.py
```diff
@@ -0,0 +1,139 @@
...
+                    str(entry.creator_generation_date)
+                )
+            if hasattr(entry, "creator_host_name"):
+                dayone_attributes += f"    host name: {entry.creator_host_name}\n"
+            if hasattr(entry, "creator_os_agent"):
+                dayone_attributes += f"    os agent: {entry.creator_os_agent}\n"
+            if hasattr(entry, "creator_software_agent"):
+                dayone_attributes += (
+                    f"    software agent: {entry.creator_software_agent}\n"
+                )
```
micahellison: It would be great if this exporter didn't have to know anything about the specifics of the Day One Classic format so we can get a bit closer to the single responsibility principle. For instance, the DayOne class could keep an inventory of its metadata, and this exporter could just ask it for that metadata.

micahellison (changes requested):
Some initial thoughts here as you're working on this.

jrnl/plugins/pelican_markdown_exporter.py
```diff
@@ -0,0 +1,139 @@
+#!/usr/bin/env python
```
wren: Would you mind splitting adding a new exporter into a separate PR? I think this PR has good fixes for the YAML exporter, but I'd like to be able to talk about adding a new plugin separately from these fixes.

features/format.feature
```diff
@@ -528,6 +528,92 @@ Feature: Custom formats
         | basic_folder    |
         # | basic_dayone    | @todo
 
+@skip_win
```
wren: Indentation typo.

features/format.feature
MinchinWeb: I haven't added any tests, only adjusted the (exiting) tests so they pass. Both the behave and pytest-bdd versions.

Moving the test suite to pytest-bdd is beyond the intended scope of this PR.

jrnl/plugins/__init__.py
minchinweb: I tired naming it generically last time, and the exporter got dumped. I would like a specific name so it is clear exactly what the target use is.

Is there some YAML + Markdown standard is use outside of Pelican?

jrnl/plugins/pelican_markdown_exporter.py
minchinweb: It is likely a good section to be broken out into a function, except that it isn't used elsewhere and so hasn't already.

In any case, this is the code as it has been since 2014, and significant refactoring of the plugin is beyond the intended scope of this PR.

jrnl/plugins/pelican_markdown_exporter.py
minchinweb: The problem with that logic (asking the DayOne format for the metadata) is that it still needs to be presented in an export specfic format; i.e. there is no way (at least that I know of) to generalize this to a function that will work across the YAML, JSON, and XML exporters.

In any case, this is the code as it has been since 2014, and significant refactoring of the plugin is beyond the intended scope of this PR.

jrnl/plugins/pelican_markdown_exporter.py
minchinweb: The whole point is this isn't adding a new exporter, but rather restoring the YAML exporter that existed in version 2.6 and before. The "pure" YAML is the new plugin.

features/format.feature
minchinweb: fixed

features/format.feature
minchinweb: The added tests have been added to the pytest setup.

minchinweb:
The failing tests are due to poety.

Documentation will also need to be updated, but this should be ready to merge.

jrnl/plugins/__init__.py
micahellison: > Is there some YAML + Markdown standard is use outside of Pelican?

There's no official standard, though it is widely used, and it is consistently referred to as "Front Matter" by a variety of well-known tools, including Jekyll, Hugo, and Assemble.

micahellison:
We have some cache changes in develop. I'm merging it in with the hopes that it will fix the Github Actions failure on Python 3.9, since the cache seems to be tripping over 3.9.6 vs. 3.9.7.

jrnl/plugins/pelican_markdown_exporter.py
wren: I sense some frustration in your reply (am I reading that right?).

I'll give some more context below, but the gist is that I would really prefer to see this split up into two PRs (a bugfix one, and an enhancement one). Partially because it's easier to review and document smaller changes, but also because I want to be sure you get credit in the changelog for the work you've done.

So, for context: I don't consider changing undocumented functions a breaking change.  Since the docs say this is a YAML format, as long as the output is still valid YAML, then it's not a breaking change in my eyes.

This does raise the question about whether these formats should be considered a stable API or not, but that feels like a conversation more related to the new plugin system that's in the works (thanks again for all your work on that). For the time being, though, these plugins aren't/haven't been stable (YAML isn't the only one that's changed output, and some of them flat out don't work under certain conditions).

So, that's all to say that I think the best way to prevent these types of misunderstandings is to make sure that your work is well documented and tested. I try to review the docs when considering changes, so it's much more likely that a breaking change will be caught during code review if it's in the docs (this also helps other contributors since they wouldn't have to dig through old closed issues to understand what a feature is supposed to do).

Anyway, I understand if you are frustrated, but hopefully you can understand where I'm coming from.

jrnl/plugins/__init__.py
minchinweb: > There's no official standard, though it is widely used, and it is consistently referred to as "Front Matter" by a variety of well-known tools, including Jekyll, Hugo, and Assemble.

Perhaps the concept is being well used, but each of static site generators have their own take on it, differing in deliminator as well as what keys they use. Pelican is the first sample below; the rest are taken from the documentation of the others you mention:

```
Title: My super title
Date: 2010-12-03 10:20
Modified: 2010-12-05 19:30
Category: Python
Tags: pelican, publishing
Slug: my-super-post
Authors: Alexis Metaireau, Conan Doyle
Summary: Short version for index and feeds

This is the content of my super blog post.
```

```
---
layout: post
title: Blogging Like a Hacker
---
```

```
---
title: "My First Post"
date: 2019-03-26T08:47:11+01:00
draft: true
---
```

```
---
page:
  title: Home
content: ../path/to/content.md
---
```

One blogger even went as far write a [conversion script](https://dev.to/wayofnumbers/how-i-migrate-my-data-science-blog-from-pelican-to-hugo-g63) when he moved from Pelican to Hugo.

The tl,dr version seems to be that if there is a standard, Pelican isn't following it, and thus a generic name seems out of place.

wren: @MinchinWeb @micahellison It sounds like there's more conversation needed about this specific plugin. Can you please take it out of this PR and either make a new PR, or make an issue to iron out all these details? It would speed up getting the rest of the PR merged, and would help keep the relevant conversation in one place for posterity.

minchinweb: @micahellison Is there a standard somewhere that I'm not aware of? I haven't found anything published that also lines up with what Pelican uses...

jrnl/plugins/pelican_markdown_exporter.py
minchinweb: @wren Frustration that the fun was been drawn out of this, and to a lesser extent, frustration at the limited time I have to move any of this forward.

I have slimmed this PR down to just deal with the "old" YAML exporter (renamed "Pelican-Markdown"). The other pieces will be re-introduced as a separate pull request as time and ability permit; that said, moving this and the plugin system forward remain higher personal priorities.

As for documentation, the YAML exporter is documented, [via the class doc string](https://github.com/jrnl-org/jrnl/blob/a9c539821f127c0bda93e35617858e3b73b717fe/jrnl/plugins/yaml_exporter.py#L17) as `"""This Exporter can convert entries and journals into Markdown formatted text with YAML front matter."""`. That original line was written in May of 2015. The format documentation on the jrnl.sh site is probably a good idea, but it was written much later. I've updated that documentation site with this PR which should solve this going forward.

Hopefully this is enough to get this moving again.

minchinweb:
@micahellison tests are passing again!

micahellison:
This approach, as it stands, is just not going to work for jrnl's core code base. I understand wanting an export format that conforms to a specific site generation workflow, but it just feels too niche to include and support in the long term as a major format.

I think there's a path forward for supporting a Pelican format, but it would be as a subset of the broader front matter format, perhaps with some sort of switch in the configuration, or as a suffix in the format name. The main difference between the various front matter formats we've discussed (Hugo, Jekyll, etc.) is less about whether or not they are front matter (they all are, even if there is no official standard), but instead in the specific object model that each format using.

A PR adding in that functionality should also not remove the existing YAML format, since that's a separate concern.

jrnl/plugins/pelican_markdown_exporter.py
wren: @MinchinWeb I'm sorry that you're feeling frustrated. It's clear that you disagree with how we numbered some of our past versions. I want to thank you for bring this issue up (it's led me to think more about what semver means to jrnl, and what things I might be able to do differently in the future).

That being said, though, I don't think going down the road of comparing timestamps of different files is productive in this conversation; what's done is done (especially since PyPI doesn't allow renumbering releases).

I'm also sorry that you're not having fun anymore. I just want to take a moment to remind you that you're a volunteer, and it's absolutely fine to take a break for a while if working on this is no longer fun for you. I value and appreciate your time, so please take care of yourself. We'll be here whenever you're ready to come back and continue contributing.

PR #1344, opened 2021-09-18
Add CACHE_STRING secret to bust GitHub Actions python cache

micahellison:
Sometimes the cache gets out of date and a PR starts failing, especially if it has been open for a while. Looks like one cause for this is if a Python version is updated by GitHub actions after the PR was submitted.

This change here gives us a GitHub secret we can use to reset the cache.

This solution isn't ideal because it doesn't solve the root cause, but it does give us a quick fix for the problem when it occurs.

### Checklist

- [ ] I have read the [contributing doc](https://github.com/jrnl-org/jrnl/blob/develop/CONTRIBUTING.md).
  for the same issue.

### DIFF
--- a/.github/workflows/testing.yaml
+++ b/.github/workflows/testing.yaml
@@ -49,7 +49,7 @@ jobs:
       uses: actions/cache@v2
       with:
         path: .venv
-        key: ${{ runner.os }}-${{ hashFiles('poetry.lock') }}-${{ matrix.python-version }}-v2-${{ secrets.CACHE_STRING }}
+        key: ${{ runner.os }}-${{ hashFiles('poetry.lock') }}-${{ matrix.python-version }}-${{ secrets.CACHE_STRING }}
 
     - name: Install dependencies
       if: ${{ matrix.python-version != '3.10-dev' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/testing.yaml
```diff
@@ -45,11 +45,11 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
 
-    - name: poetry cache
+    - name: poetry cache # Change CACHE_STRING secret to bust the cache. Useful with minor Python version changes.
       uses: actions/cache@v2
       with:
         path: .venv
-        key: ${{ runner.os }}-${{ hashFiles('poetry.lock') }}-${{ matrix.python-version }}-v2
+        key: ${{ runner.os }}-${{ hashFiles('poetry.lock') }}-${{ matrix.python-version }}-v2-${{ secrets.CACHE_STRING }}
```
wren:
```suggestion
        key: ${{ runner.os }}-${{ hashFiles('poetry.lock') }}-${{ matrix.python-version }}-${{ secrets.CACHE_STRING }}
```

wren (approved):
👤

wren (approved):
🧨

Issue #1287, opened 2021-07-03
Pytest BDD follow-up tasks


micahellison:
We just merged in a big PR to migrate our tests from `behave` to `pytest-bdd`. There are a few follow-up tasks, though:

- [x] Update CI: add flake8, remove PyFlakes
- [x] Implement tests added to Behave during pytest BDD migration then remove Behave from the makefile and CI
- [ ] Improve handling of the ExitStack logic for mocking

PR #1346, opened 2021-10-02
Move remaining tests from behave to pytest

wren:
Fixes one of the items in #1287

Some key points
- add newer tests to pytest (from behave)
- add new step implementations to pytest
- remove 'lib' from gitignore (since we're using it in pytest now)
- general feature file cleanup

Note: this does not implement any tests using `--config-override`. It leaves them out
with a comment to be implemented later.

### Checklist

  for the same issue.


micahellison (approved):
🧪🧪🧪🧪🧪

Issue #1335, opened 2021-09-06
Rendering Markdown With a Formatted and Syntax Highlighted View


viegasfh:
### Affected Page(s)

  - https://jrnl.sh/en/stable/recipes/

### What Could Be Better?

Many of Jrnl users render their output as Markdown, and when using the option --format md, we don't get a formatted and syntax highlighted view of output. I am proposing this as a feature, but at the moment there is a workaround using an external tool through piping. There are a few of these tools that can render Markdown in a formatted and syntax highlighted view, but the one  I am sharing here is called **mdless**.

This is how I we can use it:

```
jrnl myjournal -1 --format md | mdless -w 70
```
I use a column width of 70, but you can always use the default or adjust as you wish.

With this recipe one can visualize the Markdown output formatted in the command line.

So that you can see the difference,  I am sharing the default output by jrnl here:

![image](https://user-images.githubusercontent.com/290857/132238330-8a25e9fb-d87a-4cab-ac7c-79b90966ef2b.png)

And the formatted output with mdless here:

![image](https://user-images.githubusercontent.com/290857/132238381-ba102566-2fe3-4570-9430-01d13c801ba6.png)

### Other Information

For Windows users I guess they would have to use PowerShell.

I use a lot of Markdown markup in my entries and always render it as Markdown, and because I found this combination useful, I decided to share it and propose it to be included in the documentation.

micahellison:
Thanks for submitting this. I think this would be welcome in the recipes docs. Would you like to make a pull request for this?

viegasfh:
> Thanks for submitting this. I think this would be welcome in the recipes docs. Would you like to make a pull request for this?

Certainly! I will add it to the recipes doc and then I'll make a pull request and wait for suggestions for improvement in case you find it needs to.

viegasfh:
@micahellison , I have already added this to the documentation, but then got a little busy and forgot to ask you about the inclusion of images. Do you have any guidelines with regards to using images (screenshots in this case). I have already used code blocks where I could, but the rendered Markdown I can't show the example without an image screenshot.

What do you suggest? This way I finish and make a pull request.

micahellison:
> @micahellison , I have already added this to the documentation, but then got a little busy and forgot to ask you about the inclusion of images. Do you have any guidelines with regards to using images (screenshots in this case). I have already used code blocks where I could, but the rendered Markdown I can't show the example without an image screenshot.
> 
[...]

PR #1354, opened 2021-10-12
Added Recipe for visualizing Markdown in the CLI

viegasfh:
I have added a section in the recipes documentation that illustrates how to visualize Markdown from the CLI with mdless.

### Checklist

  for the same issue.

### DIFF
--- a/docs/recipes.md
+++ b/docs/recipes.md
@@ -28,7 +28,7 @@ You can do things like
 jrnl @fixed -starred -n 10 -to "jan 2013" --short
 ```
 
-To get a short summary of the 10 most recent, favourited entries before
+To get a short summary of the 10 most recent, favourite entries before
 January 1, 2013 that are tagged with `@fixed`.
 
 ### Statistics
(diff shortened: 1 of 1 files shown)


micahellison (approved):
Thank you for the PR! I particular appreciate all the little housekeeping changes you made throughout this doc.

Just a few notes -- no need to take action on any of these since I just made them.
* Piping works on cmd.exe so no need to mention PowerShell
* I was a little wary of sounding like we are endorsing mdless, so I condensed the language around that
* I thought a link to less might help, since Windows users generally only know about `more`

I'll merge this all in, and it will show up with the next jrnl release.

micahellison:
Fixes #1335.

viegasfh:
Thanks for your remarks and fixes, @micahellison !

I'll take the remarks into consideration in future contributions. As for cmd.exe, I honestly didn't know it supported piping. I have always had the impression it didn't. Thanks for pointing that! I am used to piping on Unix/Linux, but always had the impression that somehow cmd.exe didn't support piping, or maybe the early incarnations of it didn't?

Anyway, thanks once again!

PR #1373, opened 2021-11-13
Use full Python version for GitHub Actions cache key

micahellison:
We keep running into stale cache issues when this happens:
* Someone opens a PR
* Python releases a new patch version (like from 3.9.7 to 3.9.8)
* The next time tests run on that PR, they crash because they're have references to the older patch version

This should fix the issue by grabbing the full `python --version`  output such as "Python 3.9.8" and using that as part of the cache key.

### Checklist

  for the same issue.


wren (approved):
lgtm 🧨

PR #1378, opened 2021-11-20
Use locale from config if specified, otherwise use user's environment locale

### LINKED ISSUE #1353: jrnl doesn't use OS's locale settings when displaying time
### Environment
```
jrnl: v2.8.3
Python: 3.9.1 (tags/v3.9.1:1e5d33e, Dec  7 2020, 17:08:21) [MSC v.1927 64 bit (AMD64)]
OS: Windows 10
```

Coming from Germany, I would like to see the locale’s full weekday name as described [here](https://docs.python.org/3/library/time.html#time.strftime). So on Fridays, the output should be "Freitag".

### What have you tried?
Setting the configuration string to `timeformat: '%A %d-%m-%Y %H:%M'` did not change the language as I expected.

micahellison:
Fixes #1353 by checking the config file for a `locale` setting. If there's no locale setting, then it uses the environment's locale instead.

🚨 Affects both the presentation _and_ storage of dates! 🚨

I'm also going to keep this as a draft until I:

- [ ] add documentation about this feature
  - [ ] include some "gotchas" around Windows vs other environments - from my testing, it seems like Windows only uses the Control Panel's language setting, instead of any environment variables
- [ ] address concerns around automatically fetching the locale and possibly messing with the dates on existing journals (maybe only an issue for people that use %c in their timeformat?) - I don't want every entry between the 1st and 12th day of the month to suddenly get its date mangled because someone is switching their environment between en_US and en_GB, for instance
- [ ] add additional tests?
- [ ] possibly force locale in test code with mocks so that "make test" works right out of the box regardless of the developer's locale
- [ ] gracefully deal with unsupported locales instead of spewing a stack trace

If anyone else would like to work any of the above, feel free to contribute. I'll mainly be working on this on Saturdays, Pacific time.

### Checklist

  for the same issue.

### DIFF
--- a/jrnl/jrnl.py
+++ b/jrnl/jrnl.py
@@ -1,6 +1,7 @@
 # Copyright (C) 2012-2021 jrnl contributors
 # License: https://www.gnu.org/licenses/gpl-3.0.html
 
+import locale
 import logging
 import sys
 
@@ -47,6 +48,9 @@ def run(args):
     except UserAbort as err:
         print(f"\n{err}", file=sys.stderr)
         sys.exit(1)
+
+    # From here on, dates and numbers will use the locale supplied by the user
+    _init_locale(config)
 
     # Run post-config command now that config is ready
     if callable(args.postconfig_cmd):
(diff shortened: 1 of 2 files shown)


micahellison:
Finally got a basic test working on Windows and Mac after learning--to my horror!--that strftime('%c') has different behavior across OSes even if the locale is the same. Even more specific format specifiers can vary, like %A (the day of the week) which has varying capitalizations in French.

However, Ubuntu is still failing because it doesn't seem to have any of these languages. I think I either need to a) find a locale on our Ubuntu CI setup that isn't en_US, or b) install those locales before running tests. Maybe I'll tackle that next weekend.

micahellison:
I'm closing this since I haven't made any progress on this for six months and I don't think I'm going to return to this. However, if you're working on this issue and have any questions, please feel free to @ me.

PR #1397, opened 2022-01-15
Add --co alias for --config-override

### LINKED ISSUE #1394: Add --co alias for --config-override
## Feature Request

It would be nice to use `--config-override` with a shorter alias.

### Use Case/Motivation

When `--config-override` was implemented, I started using `--co` for it. Our CLI parser automatically matches up abbreviated arguments like this with their longer endings... unless it's ambiguous, which it is in the current beta release because of `--config-file`:
```
jrnl --co
usage: jrnl [--debug] [--help] [--version] [--list] [--encrypt] [--decrypt] [--import] [-on DATE] [-today-in-history] [-month DATE]
            [-day DATE] [-year DATE] [-from DATE] [-to DATE] [-contains TEXT] [-and] [-starred] [-n [NUMBER]] [-not [TAG]] [--edit]
            [--delete] [--format TYPE] [--tags] [--short] [--config-override CONFIG_KV_PAIR CONFIG_KV_PAIR]
            [--config-file CONFIG_FILE_PATH]
            [[...]]
jrnl: error: ambiguous option: --co could match --config-override, --config-file
```
It would be nice to specify this clearly. The new `--config-file` argument already has a `--cf` alias defined, so `--co` is very consistent. It also doesn't interfere with `-contains` since that argument has only one hyphen.

### Example Usage

View latest 999 entries with a different timeformat than usual:
```
jrnl --co timeformat '%c' -999
```

micahellison:
Fixes #1394.

### Checklist

  for the same issue.

### DIFF
--- a/jrnl/args.py
+++ b/jrnl/args.py
@@ -336,6 +336,15 @@ def parse_args(args=[]):
            \t jrnl --config-override colors.body blue --config-override colors.title green
         """,
     )
+    config_overrides.add_argument(
+        "--co",
+        dest="config_override",
+        action="append",
+        type=str,
+        nargs=2,
+        default=[],
+        help=argparse.SUPPRESS,
+    )
 
     alternate_config = parser.add_argument_group(
         "Specifies alternate config to be used",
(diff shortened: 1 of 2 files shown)


wren (approved):
lgtm

PR #1412, opened 2022-02-12
Drop support for Python 3.7 and 3.8

### LINKED ISSUE #1406: Sync jrnl's Python version support more closely to Python release cycle
Currently we're supporting Python 3.7+, but Python has been iterating pretty quickly in recent years and we'd like to take advantage of new features.

According to [PEP 602](https://www.python.org/dev/peps/pep-0602/), each minor Python version gets bug fixes for a year and a half, and right now Python 3.9 is the earliest version in that phase. There's another, longer period for security fixes, but I don't think we need to support Python versions throughout that entire cycle.

Maybe we could support each minor Python version for around 6 months after its final bugfix release. For instance Python 3.8 stable was released 2019-10-14, but its final bugfix release was 2021-05-03, meaning we would have looked into dropping it around the end of 2021. This schedule would have us dropping 3.9 at the end of 2022.

Feel free to chime in if you have any thoughts -- I'm especially interested in package managers' thoughts on this.

micahellison:
Fixes #1406.

### Checklist

  for the same issue.
~~I have written new tests for these changes, as needed.~~

### DIFF
--- a/.github/workflows/testing_prs.yaml
+++ b/.github/workflows/testing_prs.yaml
@@ -32,7 +32,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [ 3.7, 3.8, 3.9, '3.10' ]
+        python-version: [ 3.9, '3.10' ]
         os: [ ubuntu-latest, macos-latest, windows-latest ]
     steps:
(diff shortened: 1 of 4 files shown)


wren (approved):
⚖️

Issue #1415, opened 2022-02-19
Warn user when there are duplicate keys in the config file


micahellison:
### Use Case/Motivation

If there are duplicate keys on the same level of the config file (e.g. "editor" shows up twice at the top level), our loader just picks one of them (currently the first, though I'm investigating replacing the loader with another loader that is picking the last). I think this is fine, and we should be permissive when it comes to loading the YAML file, but it would also be good to give some feedback to the user since this could be tough to troubleshoot otherwise.

### Example Usage

Maybe the warning could say something like: "Warning: the [key name here] key appears twice at the same level in your configuration file."

### Other information

It's NOT a problem if a key appears once at each level - for instance, if I have an "editor" specified at the top level, but a different "editor" specified for a particular journal. Don't warn the user in this case.

To resolve this issue:
- [ ] Add the warning when the config file loads
- [ ] Add BDD tests to ensure that the duplicate key warning:
  - [ ] appears with top level duplicates
  - [ ] appears with journal-level duplicates
  - [ ] does not appear with a value appearing once at the top level and once at the journal level
  - [ ] does not appear when using --config-override on an existing value
  - [ ] does appear when using --config-file to point to a config file with duplicate keys

jonakeys:
Please assign this issue to me. I have the code ready, but need to create the tests. After that I'll create the PR.
Edit:
PR filed

Issue #1102, opened 2020-11-28
Refactor configuration retrieval/writing


micahellison:
### Use Case/Motivation

Down the road, we'd like to allow users to modify the config from the command line (#330, #1068) but currently our configuration code is a bit finnicky.

We also currently have a process in our code that is constantly rewriting the config file, when it probably shouldn't be.

And it'd be very nice to support and preserve detailed default comments in the config file (#946).

And, as we come up with new config keys and deprecate old ones, it would be nice to have defaults for various keys that never need to be expressed in the config file, rather than constantly adding a "config stew" of keys.

To resolve this issues, we're probably going to want to:

* Pick out a different YAML parsing library to read the config file
* Create a class that represents config data
  * This class, or the code that reads the config data, should include defaults as necessary (even if the key wasn't found)
* Ensure that all code reading config data uses this new class
* Modify the existing code that writes to the config file

micahellison:
It looks [pydantic](https://pydantic-docs.helpmanual.io/) could be promising for validating the config file when jrnl first loads.

micahellison:
For #946 to work using ruamel.yaml, we'll need to deal with two different representations of the configuration:

* the actual YAML contents (including with comments), which can be treated more like a database context
* the config values as read by the rest of jrnl, which can be treated more like a Plain Old Python Object (aka data transfer object)

One of the big innovations of ruamel seems to be around modifying just part of the YAML at once. We could take that approach, or keep loading the whole thing into memory then dumping it back. I don't think the performance implications are going to be significant either way so my impulse is to focus instead on maintainability wherever possible.

micahellison:
This refactor should also at least leave open the ability to read different versions of the config file, and probably some sort of versioning applied to the default values when they're not in the config file.

Issue #946, opened 2020-05-09
Add inline documentation to configuration file


micahellison:
## Feature Request

The configuration file would be a lot easier to work with if it had:
* some comments explaining each configuration value
* a more general section explaining the structure of the config file (e.g. multiple journals)
* a link to the online documentation

Since jrnl writes back to the config file dynamically, this concern should probably be handled in the code that writes to the config file.

micahellison:
I did a little research on this, and it looks tricky to use Python write to YAML with inline comments next to each property. I'm hesitant to add that kind of complexity to operations on a file that is so important to every function of this program.

However, I think it's realistic to add a preamble of comments to the beginning of the file. We'd still want to be careful to add it only once, but also to avoid removing any commented-out lines the user has already added to the start of the file.

wren:
It looks like [ruamel](https://pypi.org/project/ruamel.yaml/) is a pyyaml fork (much more active lately than pyyaml, too) that has the ability to read and write comments from a yaml file along with the rest of the data.

I haven't looked at it closely to see if it's worth adding as a dependency, but it could be useful for this ticket.

PR #1416, opened 2022-02-19
Replace PyYAML with ruamel.yaml

micahellison:
This PR changes our YAML config parser from PyYAML to ruamel.yaml as a first step toward being able to support and preserve comments in the config file (#946) and refactor the config data retrieval (#1102).

The only behavioral change I've found so far is that when there are duplicate keys at the same level in the config file, PyYAML picks the first key, while ruamel.yaml picks the last key. I filed #1415 to address this.

### Checklist

  for the same issue.


wren:
The format for the config file changes a bit with this. It's still valid YAML, but it's a different style. Is that what we want?
Before:
```yaml
colors:
    body: none
    date: black
    tags: yellow
    title: cyan
```

After:
```yaml
colors: {body: none, date: black, tags: yellow, title: cyan}
```

micahellison:
Got the collapsing YAML issue fixed with the by setting the YAML object's default_flow_style to false.

wren (dismissed):
Looks good as soon as conflicts are resolved. 🤠

wren (approved):
🤠

Issue #1141, opened 2021-01-02
Remove sys.exit calls as necessary


micahellison:
There are nearly a dozen sys.exit calls throughout jrnl's source. Ideally, only the main program flow (cli.py) would be calling this function. It shouldn't be the responsibility of any other part of the code to decide to close the program or return a success/error status to the operating system.

As part of this, we could also potentially clean up some except blocks in features/steps/core.py that trap for SystemExit (which fires when sys.exit is called).

PR #1417, opened 2022-02-19
Reformat messages and add new centralized exception handling

wren:
This is the start of fixing #1141 . Once we have centralized exception handling in place (this PR), then we'll be able to go through each error, exception, sys.exit call and get rid of them. There will be a follow-up PR to start going through and remove all of those calls.

### Checklist

  for the same issue.


jrnl/editor.py
```diff
@@ -48,15 +51,16 @@ def get_text_from_editor(config, template=""):
 
 def get_text_from_stdin():
     _how_to_quit = "Ctrl+z and then Enter" if on_windows() else "Ctrl+d"
-    print(
-        f"[Writing Entry; on a blank line, press {_how_to_quit} to finish writing]\n",
-        file=sys.stderr,
+    print_msg(
```
micahellison, line 56: image
I think the "To finish writing" instructions should be colorized or at least formatted somehow. What do you think? See screenshot above to illustrate how input text can easily be conflated with the "To finish writing" text.

jrnl/editor.py
```diff
@@ -48,15 +51,16 @@ def get_text_from_editor(config, template=""):
...
+        f"To finish writing, press {_how_to_quit} on a blank line.",
+        msg=Message.NORMAL,
     )
     try:
         raw = sys.stdin.read()
     except KeyboardInterrupt:
         logging.error("Write mode: keyboard interrupt")
-        print("[Entry NOT saved to journal]", file=sys.stderr)
-        sys.exit(0)
+        print_msg("Entry NOT saved to journal", msg=Message.NORMAL)
```
micahellison: Should this be a "WARNING" instead of "NORMAL"?

micahellison (changes requested):
Just a couple small comments/questions

jrnl/editor.py
wren: Added some brackets around it.

micahellison (approved):
🎨📝

Issue #1422, opened 2022-03-12
DayOne journals are off by one hour during daylight savings time


micahellison:
### Environment
  - `jrnl --diagnostic` output: 
```
jrnl: v2.8.4
Python: 3.8.2 (tags/v3.8.2:7b3ab59, Feb 25 2020, 23:03:10) [MSC v.1916 64 bit (AMD64)]
OS: Windows 10
```
  - Install method:  running off develop brunch with `poetry run`

### Current Behavior

When I make an entry in a DayOne journal at 2022-03-12 at 11:59pm, it's recorded as 2022-03-13 at 12:59am. I believe it's because of a combination of daylight savings time and time zones: 2022-03-13 is "spring ahead" for daylight savings time, while I am in Pacific time (UTC-8). So my local "2022-03-12 at 11:59pm" is UTC 2022-03-13 07:59, which is after our 2:00am "spring ahead" daylight savings transition, putting our time zone at UTC-7, leading to the entry's local datetime being 2022-03-13 00:59.

### Repro Steps

```
C:\dev\jrnl\tests>poetry run jrnl --config-file c:\dev\jrnl\tests\data\configs\dayone.yaml 2022-03-12 at 11:59pm: this should be on 20
22-03-12 at 11:59pm
[Entry added to default journal]

C:\dev\jrnl\tests>poetry run jrnl --config-file c:\dev\jrnl\tests\data\configs\dayone.yaml -2
2013-07-17 11:38 This entry is starred!

2022-03-13 00:59 this should be on 2022-03-12 at 11:59pm
```

I tried the same test with different values and it lines up with -8 hours from the daylight savings time jump:
```
2022-03-12 17:59 this should be on 2022-03-12 at 5:59pm

2022-03-12 18:59 this should be on 2022-03-12 at 6:59pm

2022-03-12 20:59 this should be on 2022-03-12 at 7:59pm

2022-03-13 00:59 this should be on 2022-03-12 at 11:59pm
```

### Other Information

Relates to tzlocal work. See my note in #1338 about possibly using standard library calls instead.

Only affects DayOne journals. All other journals are not time zone aware.

PR #1423, opened 2022-03-12
Fix tests for daylight savings

wren:
This PR:
- Fixes current test that was failing on the day before daylight savings
- Adds new set of tests to catch this issue in the future
- Skips the new test for now, because the issue still exists (#1422)

### Checklist

  for the same issue.


micahellison (approved):
📅

PR #1436, opened 2022-03-19
Remove sample format and its asteval dependency

### LINKED ISSUE #1435: Remove "sample" format and its asteval dependency
## Background
The are two old template files lurking in the source under /plugins/ that are not referenced by the rest of jrnl:
* template.py
* template_exporter.py

These files don't have much documentation, and I'm not sure they are complete. Their creation isn't associated with any GitHub issue, and they could only really ever be used via a monkey patch. They are the sole reason for our dependency on `asteval`.

This is *not* to be confused with the `template` config key, which is handled elsewhere.

## Impact
The only documentation that references these files is in the `--help` page under `--format`, which mentions that `sample` is one of the available formats. `sample` routes to this sample template, hence its name. However `sample` isn't mentioned in our web documentation and I don't think it ever has been.

I think it makes sense to just remove these. It'd be great to have some solid template exporting functionality, but I don't think this is the foundation to build that on. Plus it's always nice to pare down unused dependencies. Moreover, I don't think it is a good long-term situation to have an arbitrary format named "sample".

micahellison:
Fixes #1435. 

### Checklist

  for the same issue.


wren (approved):
🖇️

PR #1449, opened 2022-04-20
Yaml export errors now don't show stack trace

### LINKED ISSUE #1227: Directory YAML export shows stack trace when it doesn't need to
### Environment
  - `jrnl --diagnostic` output: 
```
jrnl: v2.8
OS: Windows 10
```
  - Install method:  pipx

### Current Behavior

After running the following in a directory without nonexistent_dir:
```
jrnl --format yaml --file nonexistent_dir
```
I get the following output:
```
ERROR: YAML export must be to individual files. Please specify a directory to export to.
    exec(code, run_globals)
  File "c:\users\micah\.local\pipx\venvs\jrnl\lib\site-packages\jrnl\cli.py", line 35, in cli
    return run(args)
  File "c:\users\micah\.local\pipx\venvs\jrnl\lib\site-packages\jrnl\jrnl.py", line 73, in run
    search_mode(**kwargs)
  File "c:\users\micah\.local\pipx\venvs\jrnl\lib\site-packages\jrnl\jrnl.py", line 179, in search_mode
  File "c:\users\micah\.local\pipx\venvs\jrnl\lib\site-packages\jrnl\jrnl.py", line 337, in _display_search_results
  File "c:\users\micah\.local\pipx\venvs\jrnl\lib\site-packages\jrnl\plugins\text_exporter.py", line 77, in export
    return cls.write_file(journal, output)
  File "c:\users\micah\.local\pipx\venvs\jrnl\lib\site-packages\jrnl\plugins\text_exporter.py", line 35, in write_file
    f.write(cls.export_journal(journal))
TypeError: write() argument must be str, not None
```

### Expected Behavior

The first line is fine: `ERROR: YAML export must be to individual files. Please specify a directory to export to.`

But there's no need for the stack trace.

apainintheneck:
### Checklist

  for the same issue.

This is a fix for the bug mentioned in #1227 related to failed YAML exports showing the stack trace along with an error message. Now it just shows the error message.

Instead of printing an error message and returning `none`, I changed it to throw a `RuntimeError`. The advantage of this is that the calling function doesn't throw a `TypeError` when it gets back a `none` value instead of a `string`. It also gives us the opportunity to clean up the file that was open.

Closes #1227.

### DIFF
--- a/jrnl/plugins/text_exporter.py
+++ b/jrnl/plugins/text_exporter.py
@@ -35,6 +35,9 @@ class TextExporter:
                 return f"[Journal exported to {path}]"
         except IOError as e:
             return f"[{ERROR_COLOR}ERROR{RESET_COLOR}: {e.filename} {e.strerror}]"
+        except RuntimeError as e:
+            os.remove(path)
+            return e
 
     @classmethod
     def make_filename(cls, entry):
@@ -54,6 +57,9 @@ class TextExporter:
                 return "[{2}ERROR{3}: {0} {1}]".format(
                     e.filename, e.strerror, ERROR_COLOR, RESET_COLOR
                 )
+            except RuntimeError as e:
+                os.remove(full_path)
+                return e
         return "[Journal exported to {}]".format(path)
 
     def _slugify(string):
--- a/jrnl/plugins/text_exporter.py
+++ b/jrnl/plugins/text_exporter.py
@@ -36,7 +36,6 @@ class TextExporter:
         except IOError as e:
             return f"[{ERROR_COLOR}ERROR{RESET_COLOR}: {e.filename} {e.strerror}]"
         except RuntimeError as e:
-            os.remove(path)
             return e
 
     @classmethod
@@ -58,7 +57,6 @@ class TextExporter:
                     e.filename, e.strerror, ERROR_COLOR, RESET_COLOR
                 )
             except RuntimeError as e:
-                os.remove(full_path)
                 return e
         return "[Journal exported to {}]".format(path)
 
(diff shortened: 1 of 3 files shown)


jrnl/plugins/text_exporter.py
```diff
@@ -35,6 +35,9 @@ def write_file(cls, journal, path):
                 return f"[Journal exported to {path}]"
         except IOError as e:
             return f"[{ERROR_COLOR}ERROR{RESET_COLOR}: {e.filename} {e.strerror}]"
+        except RuntimeError as e:
+            os.remove(path)
```
wren: Why are we deleting the path here? This seems like new behavior.

apainintheneck: It seemed like a good idea to cleanup the empty output file after an exception was raised.

apainintheneck: Currently, if a directory doesn't exist, a file is opened instead which remains after the command terminates.

apainintheneck: If you're worried about this applying too generally to every RuntimeError, we could also create a more specific error like ExportError. That would be used to signal that the empty or partially written file needs to be deleted and the error message should be returned as a string.

micahellison: I can understand the rationale for cleaning this path up, but I'd prefer that jrnl doesn't do this. A user might be exporting to a directory that already has other data in it, and this could remove that data. Better to leave the data and let the user deal with it appropriately.

apainintheneck: Okay, sounds good. I'll remove that then.

micahellison (approved):
Looks great! Thanks for your contributions.

PR #1461, opened 2022-04-27
Give a proper message when trying to use an empty config file

### LINKED ISSUE #1420: Empty config file leads to confusing error message
### Environment
  - `jrnl --diagnostic` output: 
```
jrnl: v2.8.4-beta2
OS: Windows 10
```
  - Install method:  pipx

### Current Behavior

If your config file is empty, you get a stack trace and a `TypeError: 'NoneType' object is not iterable` message.

### Expected Behavior

It would be nice to have an error message that explains that the config file can't be parsed. Maybe show the path of the config file too.

### Repro Steps

* Empty the contents of your `jrnl.yaml` config file and save it
* Run `jrnl`

### Other Information

Stack trace:
```
    exec(code, run_globals)
  File "C:\Users\micah\.local\pipx\venvs\jrnl\lib\site-packages\jrnl\cli.py", line 34, in cli
    return run(args)
  File "C:\Users\micah\.local\pipx\venvs\jrnl\lib\site-packages\jrnl\jrnl.py", line 39, in run
    config = install.load_or_install_jrnl(args.config_file_path)
  File "C:\Users\micah\.local\pipx\venvs\jrnl\lib\site-packages\jrnl\install.py", line 92, in load_or_install_jrnl
    upgrade_config(config, alt_config_path)
  File "C:\Users\micah\.local\pipx\venvs\jrnl\lib\site-packages\jrnl\install.py", line 28, in upgrade_config
    missing_keys = set(default_config).difference(config_data)
TypeError: 'NoneType' object is not iterable
```

A resolution to this issue should have:
- [ ] An easy-to-understand error message when the config file can't be parsed
- [ ] A BDD test that confirms the error message appears with an empty config file

jonakeys:
Hi, I'm new to making commits so all feedback is welcome. Of course I put in my best effort.

This code adds a check for an empty config file, as specified by the issue #1420 . If an empty file is found, the programs prints a message and uses the default config, so the user won't be hindered in doing what he tried to do.

I added a bdd test, but couldn't get it to work properly. The test can't seem to find the new 'empty_file.yaml' config I created. Hopefully someone can help me out here.

### Checklist

  for the same issue.

### DIFF
--- a/jrnl/install.py
+++ b/jrnl/install.py
@@ -75,6 +75,12 @@ def load_or_install_jrnl(alt_config_path):
         logging.debug("Reading configuration from file %s", config_path)
         config = load_config(config_path)
 
+        if config is None:
+            print(
+                f"Configuration file {config_path} is empty, now using default config"
+            )
+            config = get_default_config()
+
         if is_old_version(config_path):
             from jrnl import upgrade
 
--- a/jrnl/install.py
+++ b/jrnl/install.py
@@ -76,10 +76,8 @@ def load_or_install_jrnl(alt_config_path):
         config = load_config(config_path)
 
         if config is None:
-            print(
-                f"Configuration file {config_path} is empty, now using default config"
-            )
-            config = get_default_config()
+            print("Unable to parse config file", file=sys.stderr)
+            sys.exit()
 
         if is_old_version(config_path):
             from jrnl import upgrade
--- a/tests/bdd/features/config_file.feature
+++ b/tests/bdd/features/config_file.feature
@@ -85,7 +85,6 @@ Feature: Multiple journals
         Then the output should contain "Journal encrypted to features/journals/basic_onefile.journal"
         And the config should contain "encrypt: false"
 
-
     Scenario: Don't overwrite main config when decrypting a journal in an alternate config
         Given the config "editor_encrypted.yaml" exists
         And we use the password "bad doggie no biscuit" if prompted
@@ -93,3 +92,8 @@ Feature: Multiple journals
         When we run "jrnl --cf editor_encrypted.yaml --decrypt"
         Then the config should contain "encrypt: true"
         And the output should not contain "Wrong password"
+
+    Scenario: Use default config when configuration file is empty or corrupt
+        Given the config "empty_file.yaml" exists
+        When we run "jrnl --cf empty_file.yaml"
+        Then the output should contain "empty, now using default config"
--- a/tests/bdd/features/config_file.feature
+++ b/tests/bdd/features/config_file.feature
@@ -93,7 +93,13 @@ Feature: Multiple journals
         Then the config should contain "encrypt: true"
         And the output should not contain "Wrong password"
 
-    Scenario: Use default config when configuration file is empty or corrupt
+    Scenario: Show an error message when the config file is empty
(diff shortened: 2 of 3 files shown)


jrnl/install.py
```diff
@@ -75,6 +75,12 @@ def load_or_install_jrnl(alt_config_path):
         logging.debug("Reading configuration from file %s", config_path)
         config = load_config(config_path)
 
+        if config is None:
+            print(
+                f"Configuration file {config_path} is empty, now using default config"
+            )
+            config = get_default_config()
```
micahellison: Rather than reverting the default config, I think jrnl should just exit. We don't know what exactly the user wants and I'd prefer not to make an assumption that they'd want to use their default config at this point.

tests/bdd/features/config_file.feature
```diff
@@ -85,11 +85,21 @@ Feature: Multiple journals
...
+    Scenario: Show an error message when the config file is empty
+        Given we use the config "empty_file.yaml"
+        When we run "jrnl -1"
+        Then the error output should contain "Unable to parse config file"
+
+    Scenario: Show an error message when using --config-file with empty file
+        Given the config "empty_file.yaml" exists
+        And we use the config "basic_onefile.yaml"
+        When we run "jrnl --cf empty_file.yaml"
+        Then the error output should contain "Unable to parse config file"
```
micahellison, line 105: I fixed the test (it needed to have a default config specified) and modified the tests a bit to reflect my intent in the issue.

jrnl/install.py
```diff
@@ -75,6 +75,12 @@ def load_or_install_jrnl(alt_config_path):
         logging.debug("Reading configuration from file %s", config_path)
         config = load_config(config_path)
 
+        if config is None:
+            print(
+                f"Configuration file {config_path} is empty, now using default config"
+            )
```
micahellison: This error message should go to STD_ERR -- just need to add `file=sys.stderr` to the print args.

micahellison (changes requested):
Welcome and thanks for the PR, @jonakeys. I had a slightly different intent for the behavior that I realize I hadn't made clear in the original issue: we still want jrnl to fail if the config can't be parsed, rather than falling back to the default config.

Also, the error message should appear when using a default config that's empty too. I added a test for that situation and fixed the other test so that you can take a test-driven approach to this if you'd like.

jrnl/install.py
jonakeys: You're right about that. It's better to let the user decide what to do and resolve the problem first.

tests/bdd/features/config_file.feature
jonakeys: Thanks for updating the tests, I get what they exactly do now.

jrnl/install.py
jonakeys: Indeed better to direct this to the error output.

jonakeys:
Thanks for your welcoming words and suggestions! I solved the issues and committed the changes. Had to add a condition to the test, because it would always automatically insert a version number and then the file wouldn't be empty.

wren (approved):
Looks great. Thank you!

micahellison (approved):

PR #1468, opened 2022-05-05
Always expand all paths (journals, templates, etc)

### LINKED ISSUE #998: Paths starting with tildes (~) should always expand
### Environment
  - Jrnl version: 
  - Install method: 
  - OS 
Likely all versions.

### Current Behavior

`~/template.md` doesn't work for a template file. Has this error:
`[Could not read template at '~/template.md']`

### Expected Behavior

The `~` should be expanded the same way it is for journal file themselves.
### Other Information

Current workaround is to use the full path to a file (e.g. `/Users/wren/template.md`) instead.

apainintheneck:
This also fixed bugs with relative journal and template paths.

### Checklist

  for the same issue.

This is related to the discussion in issue #998. I took a look at the path expansion code used throughout the project and decided to create a new `path.py` file to hold the three common cases.
1. Getting the path to the home directory.
2. Expanding shell variables and the user path (`~`).
3. Getting the absolute path after expanding shell variables and the user path (`~`). Note: This used to happen in the opposite order which is probably what caused problems.

Additionally, a call to `expand_path` is now made before loading template files which solves the problem with loading files like `~/template`. I'm still not sure though if the error message after failing to load a template should print the original template path string or the expanded one (line 210 in `jrnl/jrnl.py`).

Closes #998.

### DIFF
--- a/jrnl/path.py
+++ b/jrnl/path.py
@@ -1,0 +1,13 @@
+import os.path
+
+
+def home_dir():
+    return os.path.expanduser("~")
+
+
+def expand_path(path):
+    return os.path.expanduser(os.path.expandvars(path))
+
+
+def absolute_path(path):
+    return os.path.abspath(expand_path(path))
--- a/jrnl/Journal.py
+++ b/jrnl/Journal.py
@@ -11,6 +11,7 @@ import sys
 from . import Entry
 from . import time
 from .prompt import yesno
+from .path import expand_path
 
 
 class Tag:
@@ -404,7 +405,7 @@ def open_journal(journal_name, config, legacy=False):
     backwards compatibility with jrnl 1.x
     """
     config = config.copy()
-    config["journal"] = os.path.expanduser(os.path.expandvars(config["journal"]))
+    config["journal"] = expand_path(config["journal"])
 
     if os.path.isdir(config["journal"]):
         if config["encrypt"]:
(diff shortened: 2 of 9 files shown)


jrnl/path.py
```diff
@@ -0,0 +1,13 @@
+import os.path
+
+
+def home_dir():
+    return os.path.expanduser("~")
+
+
+def expand_path(path):
```
wren, line 8: I think we could do with just `home_dir` and `expand_path` as long as `expand_path` also runs `abspath`. Would you mind merging `expand_path` and `absolute_path` into one function?

wren (changes requested):
This looks great. Just one comment about the new functions to fix.

Also, would you mind adding a test or two? I think just a couple of unit tests for the new functions should be more than enough.

jrnl/path.py
apainintheneck: I tried that at first because I assumed that it wouldn't make a difference but I think that's not a great idea for a couple reasons.

1. It breaks the tests as currently written. A few BDD tests check the console output of the application and the output that gets parsed contains relative paths. That means that those tests would have to changed to parse the absolute path which is a bit more complicated.
2. The absolute path is only needed once so it would add extra work to all the other calls.
3. Console error messages are more readable when they display the relative path instead of the absolute path.

apainintheneck:
I'm currently running into some problems creating tests that work for both platforms. Essentially, the call to `os.path.abspath()` internally calls `os.getcwd()`. My first thought was to just mock that in `pytest` but that didn't work on Windows for some reason. Then, I decided to use `os.getcwd()` directly and got `FileNotFoundError`s on *nix.

wren (approved):
I went ahead and updated the tests to fix the issues you were running into.

The rest looks good. Thank you!

PR #1475, opened 2022-05-07
Rework how all output and messaging works in jrnl

### LINKED ISSUE #1024: Modularize console output messages
We have several different ways to output to the user, and some conventions around using STDOUT vs. STDERR, but it could be nice to centralize this behavior, and perhaps unify it with the logging calls.

Some reasons to do this:
* Make it clearer when to use STDOUT vs STDERR, by basing the decision on the intent of each message
* Separate concerns (e.g. issues like color selection are nice keep out of control flow logic)
* Lay a foundation for internationalization (I'll make a separate issue for this)

wren:
The main idea behind all of this is to get all user-facing output centralized into one file. This will make it much easier to translate jrnl into different languages later (since we can just swap out the file for an equivalent one in a different language).

### Checklist

  for the same issue.


micahellison (approved):
💥

PR #1478, opened 2022-05-15
Show name of journal when creating a password/encrypting

jonakeys:
This code intends to make it clearer which journal you are trying to encrypt/generate a password for. The original code said 'new journal', but it would be better if the actual name of the journal is shown at the prompt.

I couldn't find an issue number related to this change. And because it only changes the prompt I didn´t write a test for it.

### Checklist

  for the same issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,14 @@
 
 **Fixed bugs:**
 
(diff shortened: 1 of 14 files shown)


wren (approved):
Looks good. Thank you!

PR #1480, opened 2022-05-23
Editor no longer opens when -not is used without search flags

### LINKED ISSUE #1472: Using -not argument by itself leads to new entry creation
### Environment
  - `jrnl --diagnostic` output: 
```
jrnl: v2.8.4
OS: Windows 10
```
  - Install method:  pipx

### Current Behavior

When using the -not argument without other search terms, I'm sent into my editor to create a new entry.

### Expected Behavior

The -not argument is supposed to search entries for all entries that don't have the tag specified.

### Repro Steps

This should show all entries that don't have @tag in them. However, instead it leads me to create a new entry:
```
jrnl -not @tag
```

However, `-not` is still working fine with other arguments, such as:
```
jrnl -from yesterday -not @tag
```

### Other Information

A resolution to this issue should have:
- [ ] A fix to the bug
- [ ] A new BDD test to test for this behavior

apainintheneck:
### Checklist

  for the same issue.

This is fixes #1472. Essentially the problem was that `-not` gets ignored when it is not accompanied with search flags like `-on`, `-contains`, etc. That means that you could accidentally create a new entry by typing something like `jrnl -not @tag`.

I added a check to the `_is_write_mode` that throws an error if `-not` exists and we're not searching for anything. Ideally this would be able to be handled inside the parser but there are essentially two reasons why that's not possible.
1. `argparse` doesn't natively support dependent arguments.
2. We still have to account for the edge case where a string of tag arguments is interpreted as a search query.

### DIFF
--- a/jrnl/jrnl.py
+++ b/jrnl/jrnl.py
@@ -109,6 +109,19 @@ def _is_write_mode(args, config, **kwargs):
     ):
         write_mode = False
 
+    # If -not is set but we are not searching, throw an error
+    if write_mode and args.excluded:
+        raise JrnlException(
+            Message(
+                MsgText.InvalidArgumentUsage,
+                MsgType.NORMAL,
+                {
+                    "arg": "-not",
+                    "msg": "expected another search parameter like -on, -contains, etc.",
+                },
+            )
+        )
+
     return write_mode
 
 
--- a/jrnl/messages.py
+++ b/jrnl/messages.py
@@ -134,6 +134,9 @@ class MsgText(Enum):
         No entries to modify, because the search returned no results
         """
 
+    # --- Usage --- #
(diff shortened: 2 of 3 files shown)


micahellison:
Thanks for the PR, though issue #1472 is actually calling for the opposite behavior - `jrnl -not @tag` should do a search for all entries without @tag in it, according to the documentation. It doesn't affect other parameters like `-on` or `-contains`, which is by design. That said, I'm not sure I like the design, but it is enshrined in the API and I'm reluctant to change it without a lot of consideration.

Do feel free to file another PR for that behavior, or if you have any feedback on how I could make issues like this a bit clearer, feel free to share.

apainintheneck:
Ah, I totally misread that. But yeah, that behavior makes more sense.

Issue #1293, opened 2021-07-07
Exception when providing folder name of folder that does not exist as journal path


viegasfh:
### Environment
  - `jrnl --diagnostic` output: 
  - Install method: 
jrnl: v2.8.1
Python: 3.8.5 (default, May 27 2021, 13:30:53)
[GCC 9.3.0]
OS: Linux 4.4.0-19041-Microsoft
### Current Behavior

When providing a folder name as opposed to a file name in the journals configuration, I get an IsDirectoryException.

### Expected Behavior

Although the documentation does not provide examples, I have looked into the source code and found some examples of how to use a folder for multiple files grouped by date, month and day, but the examples provided presume that you have created the folder.

When providing a folder name for the journal name, the app should create the folder and then create the journal entry using the year, month and day.

When I executed the jrnl command after the error, everything worked without an issue. So, the program should create the folder and the journal entry without throwing the exception.

### Repro Steps

In the config file add the following entry or similar one. Make sure the / is found in the end of the folder name. This way jrnl treats is a folder as opposed to a file.

```
test: /home/user01/.local/share/jrnl/test/
```

Once this is configured, just run:

```
jrnl test today: My First Entry. This is an entry I am using to test the folder based journal
```

### Other Information

Here is the stack trace of the error:

```
[Directory /home/user01/.local/share/jrnl/test created]
Traceback (most recent call last):
  File "/home/user01/.local/bin/jrnl", line 8, in <module>
    sys.exit(cli())
  File "/home/user01/.local/lib/python3.8/site-packages/jrnl/cli.py", line 35, in cli
    return run(args)
  File "/home/user01/.local/lib/python3.8/site-packages/jrnl/jrnl.py", line 62, in run
    journal = open_journal(args.journal_name, config)
  File "/home/user01/.local/lib/python3.8/site-packages/jrnl/Journal.py", line 418, in open_journal
    return PlainJournal(journal_name, **config).open()
  File "/home/user01/.local/lib/python3.8/site-packages/jrnl/Journal.py", line 85, in open
    self.create_file(filename)
  File "/home/user01/.local/lib/python3.8/site-packages/jrnl/Journal.py", line 110, in create_file
    with open(filename, "w"):
IsADirectoryError: [Errno 21] Is a directory: '/home/user01/.local/share/jrnl/test/'
```

wren:
I was genuinely surprised to hear that we don't mention directory-based journals in the docs. 🤦 

[...]

PR #1482, opened 2022-05-23
Check for folder name ending with (back)slash and act on that

jonakeys:
This PR fixes the issue #1293 When reading the config file, the code now checks if the path ends with a (back)slash and if so, calls the FolderJournal instead of the PlainJournal.

I also added a BDD test for this specific check. There was already a check for a non-existent directory, but didn't check if the path ended with a (back)slash.

### Checklist

  for the same issue.

### DIFF
--- a/jrnl/Journal.py
+++ b/jrnl/Journal.py
@@ -436,6 +436,7 @@ def open_journal(journal_name, config, legacy=False):
             return LegacyJournal(journal_name, **config).open()
         if config["journal"].endswith("/") or config["journal"].endswith("\\"):
             from . import FolderJournal
+
             return FolderJournal.Folder(journal_name, **config).open()
         return PlainJournal(journal_name, **config).open()
 
--- a/jrnl/Journal.py
+++ b/jrnl/Journal.py
@@ -434,6 +434,9 @@ def open_journal(journal_name, config, legacy=False):
     if not config["encrypt"]:
         if legacy:
             return LegacyJournal(journal_name, **config).open()
+        if config["journal"].endswith("/") or config["journal"].endswith("\\"):
+            from . import FolderJournal
+            return FolderJournal.Folder(journal_name, **config).open()
         return PlainJournal(journal_name, **config).open()
 
     from . import EncryptedJournal
--- a/jrnl/Journal.py
+++ b/jrnl/Journal.py
@@ -434,7 +434,7 @@ def open_journal(journal_name, config, legacy=False):
     if not config["encrypt"]:
         if legacy:
             return LegacyJournal(journal_name, **config).open()
-        if config["journal"].endswith("/") or config["journal"].endswith("\\"):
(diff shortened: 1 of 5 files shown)


jrnl/Journal.py
```diff
@@ -434,6 +434,10 @@ def open_journal(journal_name, config, legacy=False):
     if not config["encrypt"]:
         if legacy:
             return LegacyJournal(journal_name, **config).open()
+        if config["journal"].endswith("/") or config["journal"].endswith("\\"):
```
micahellison: Could you use [os.sep](https://docs.python.org/3.9/library/os.html?highlight=os.sep#os.sep) here instead of both conditions? This will keep POSIX systems from accidentally creating folder paths ending with a backslash.

tests/bdd/features/file_storage.feature
```diff
@@ -30,6 +30,14 @@ Feature: Journals iteracting with the file system in a way that users can see
         Then the journal should exist
         When we run "jrnl -99 --short"
         Then the output should contain "This is a new entry in my journal"
+    
+    Scenario: If the directory for a Folder journal ending in a slash ('/' or '\') doesn't exist, then it should be created
```
micahellison: Could you separate into this into two tests, one with forward slashes for POSIX and one with backslashes for Windows? You can use the @on_posix and @on_windows tags to ensure they only run on the appropriate platform.

tests/lib/then_steps.py
```diff
@@ -210,6 +210,22 @@ def journal_should_not_exist(config_on_disk, should_or_should_not):
...
+@then(parse('the journal "{journal_name}" directory {should_or_should_not} exist'))
+def directory_should_not_exist(config_on_disk, should_or_should_not, journal_name):
+    scoped_config = scope_config(config_on_disk, journal_name)
+    expected_path = scoped_config["journal"]
+
+    dir_exists = os.path.isdir(expected_path)
+    if should_or_should_not == "should":
+        assert dir_exists
+    elif should_or_should_not == "should not":
+        assert not dir_exists
```
micahellison: Could you use the parse_should_or_should_not(should_or_should_not) method here? See other methods in this file like config_var_on_disk for an example.

micahellison (changes requested):
Thanks for the PR. This is a nice change. I have just a few little nitpicks about methods in my review. Also, feel free to update the docs in docs/journal-types.md (though if you don't, that's fine, I'll get to it).

jonakeys:
Thanks for your feedback @micahellison
Those are valid points you bring up and I'll insert the changes in the code. I've just recently started contributing to repo's on GitHub and I'm glad you as maintainers are kind and helpful, so I can grow and develop so I can be useful in the open source world.

jonakeys:
I committed the changes. You're welcome to review the current state of the PR.

PR #1495, opened 2022-06-04
Fixed error related to display_format in config file for some values

### LINKED ISSUE #1263: display_format: pretty and display_format: short lead to crash
### Environment
  - `jrnl --diagnostic` output: 
```
jrnl: v2.8.1
OS: Windows 10
```
  - Install method:  pipx

### Current Behavior

Setting the `display_format` configuration operation to `pretty` or `short` causes jrnl to crash when viewing the contents of the journal.

### Expected Behavior

It should use the selected format instead of crashing.

### Other Information

Stack trace is the same whether on `pretty` or `short`:
```
    exec(code, run_globals)
  File "c:\users\micah\.local\pipx\venvs\jrnl\lib\site-packages\jrnl\cli.py", line 35, in cli
    return run(args)
  File "c:\users\micah\.local\pipx\venvs\jrnl\lib\site-packages\jrnl\jrnl.py", line 73, in run
    search_mode(**kwargs)
  File "c:\users\micah\.local\pipx\venvs\jrnl\lib\site-packages\jrnl\jrnl.py", line 179, in search_mode
  File "c:\users\micah\.local\pipx\venvs\jrnl\lib\site-packages\jrnl\jrnl.py", line 340, in _display_search_results
AttributeError: 'NoneType' object has no attribute 'export'
```

apainintheneck:
### Checklist

  for the same issue.

Basically what was happening was that using `display_format: pretty` or `display_format: short` in your config file would cause an error. The reason for this was that even though those arguments were handled correctly they were sourced from the config file after those if statements had already been skipped over.

Now `_display_search_results` tries to source the `display_format` value from the config file if the `export` arg is empty before displaying search results. This is a small change so I didn't include any tests. Let me know if you think they are necessary and I'd be happy to add them.

Fixes #1263.


apainintheneck:
Actually, it looks like this clobbers the `--tags` flag so I'll take a longer look at it.

apainintheneck:
Okay, I had to change it to evaluate `--tags` before `--format`. Also, there was a weird thing that showed up in the unit tests where `mock_args.tags` was evaluating to true so I changed that as well. The funny thing is that that test isn't even testing for tags so I'm not sure why that happened.

wren (approved):
Looks good! Just updated the unit test to fix the tags issue you were running into.

micahellison (approved):
Looks good! I just added a BDD test in case of regressions.

Issue #1184, opened 2021-02-06
Refactor base Journal.py class


micahellison:
### Use Case/Motivation

The Journal class in Journal.py has a lot of responsibilities right now:
* Express the journal in as data in memory
* Reading/writing a journal file or directory
* Parsing a journal file
* Displaying the journal's entries in certain formats

This grab-bag of responsibilities has made it hard to work with, especially when it comes to adding flexibility like plugins, storage formats, and display formats.

Unless there's a compelling reason to keep some of these responsibilities in this class, I think the only concern of the Journal class should be to express the data in memory, like a "POPO" (Plain Old Python Object). Data access and data formats should probably live somewhere else.

### Other information

* [Wikipedia article on Plain Old Java Objects](https://en.wikipedia.org/wiki/Plain_old_Java_object) (which is what the notion of "Plain Old Python Objects" comes from)

micahellison:
I'm thinking about this issue as I look at #1352 -- it would be very nice to have color completely separated from the data access / data representation classes.

PR #1507, opened 2022-06-12
Add linewrap option 'auto'

### LINKED ISSUE #1246: Allow for automatic line wrapping based on detected console width
### Use Case/Motivation

Right now, we have a `linewrap` key in our configure that allows you to specify the max width of a line before wrapping. It's useful for customizing, but it could be automatic instead of manual.  Python can detect the user's terminal size and use that.

### Example Usage

Perhaps simply having an "auto" value for the linewrap key in the config would be enough:
```
linewrap: auto
```
Then jrnl would need to read that key and interpret it appropriately.

### Other information

Relevant python function: os.get_terminal_size() would do the trick. However! It is possible for it to fire an exception if there's no terminal (like perhaps in a headless configuration?) so jrnl would need to catch that and revert to a default value for the duration of the program if it fires.

jonakeys:
This resolves the feature request #1246 It adds the possibility to set 'auto' as linewrap option so jrnl automatically determines the width of the terminal/columns.

I chose to do the check while outputting information. The downside is that when using the debug flag it outputs a line for every entry. But it could also be done earlier when the config file is being read. Would that be better? If so, I'll modify the code to do so.

Fixes #1246 

### Checklist

  for the same issue.


jrnl/Entry.py
```diff
@@ -105,6 +108,18 @@ def pprint(self, short=False):
         )
 
         if not short and self.journal.config["linewrap"]:
+            columns = self.journal.config["linewrap"]
+
+            if columns == "auto":
+                try:
+                    columns = os.get_terminal_size()[0]
+                except:  # noqa: E722
```
micahellison: Could you use `except OSError` here? The docs for `get_terminal_size` say that's the exception that fires when it can't get the terminal size for standard out, I'd like to avoid generic exceptions.

micahellison (changes requested):
I have a small nitpick on the except clause, and I think this also needs some tests, either a unit test or a BDD test. Let us know if you'd like any help with that.

jonakeys:
Thanks for your remarks! I can easily specify the except clause.
But for the tests I need some help. I can create a test for the 'auto' linewrap option. But how do I create a test that checks if the `get_terminal_size` fires the exception?

micahellison:
We're working on the finishing touches of the 3.0 beta and are going to prioritize that now, but I'll get back to this PR to review in a week or so. Thanks for your work so far on it!

jonakeys:
That's no problem, I'm happy to contribute where I can.
And ofcourse the v3.0 has full focus now for you as maintainers. Good luck!

tests/bdd/features/config_file.feature
```diff
@@ -106,3 +106,8 @@ Feature: Multiple journals
         And we use the config "basic_onefile.yaml"
         When we run "jrnl --cf empty_file.yaml"
         Then the error output should contain "Unable to parse config file"
+
```
micahellison, line 109: Could you add a test for a config for format "fancy" and resolve the issue? I'm getting this message:
```
┏━ Error ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃  TypeError                                           ┃
┃  unsupported operand type(s) for -: 'str' and 'int'  ┃
┃                                                      ┃
┃  This is probably a bug. Please file an issue at:    ┃
┃  https://github.com/jrnl-org/jrnl/issues/new/choose  ┃
┗━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┛
```

micahellison (changes requested):
One minor issue - this is broken with the "fancy" format. Could you take a look? It has its own code in jrnl/plugins/fancy_exporter.py. Thanks!

tests/bdd/features/config_file.feature
jonakeys: Thanks for the find. I created a bdd test and fix to resolve the error.

micahellison (approved):
Thanks for making this change. It's working well for me. I don't like seeing the same code in multiple places at once, but this is a more of a problem with jrnl's architecture than with your PR. We eventually want to refactor the Journal/Entry classes as mentioned in #1184 to separate presentation from data travel, and that's way outside of the scope of this PR.

jonakeys:
You're welcome, glad I can contribute! I'm also not a fan of duplicate code myself. But to fix it would need to create a function in a shared file with another import as consequence. That wouldn't be beautiful too. When the refactoring takes place, it will resolve issues like this indeed.

PR #1510, opened 2022-06-13
Display message when no edits take place

### LINKED ISSUE #1501: Blank box displays when not making any changes in --edit
### Environment
  - `jrnl --diagnostic` output: 
```
jrnl: v3.0-beta
OS: Windows 10
```
  - Install method:  pipx

### Current Behavior

When you run `jrnl --edit` and change nothing, a blank box appears.

### Expected Behavior

There should probably be a message saying that nothing was edited. Or it should just be blank instead.

### Repro Steps

```
jrnl --edit
```
Then close your editor.

### Other Information

Before this version of jrnl (3.0 beta), nothing happened when running `jrnl --edit` with no changes.

apainintheneck:
### Checklist

  for the same issue.

This displays an error message when no edits take place after launching the editor with the `jrnl --edit` command. 
Screen Shot 2022-06-13 at 11 36 46 AM
I tried to model the message after `MsgText.NoTextReceived`.

Fixes #1501

### DIFF
--- a/jrnl/editor.py
+++ b/jrnl/editor.py
@@ -43,7 +43,7 @@ def get_text_from_editor(config, template=""):
     os.remove(tmpfile)
 
     if not raw:
-        raise JrnlException(Message(MsgText.NoTextReceived, MsgStyle.ERROR))
+        raise JrnlException(Message(MsgText.NoTextReceived, MsgStyle.NORMAL))
 
     return raw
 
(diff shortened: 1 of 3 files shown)


apainintheneck:
Would it be better to make this a warning instead of an error?

micahellison:
> Would it be better to make this a warning instead of an error?

I think normal would actually be best -- it's basically the same level of message as "1 entry modified" is when you successfully edit an entry.

apainintheneck:
If that's the case, I think the no text received message should also be normal since they are similar messages. This can happen normally when submitting an empty journal entry so it's not always an error.

Currently it looks like this:
Screen Shot 2022-06-18 at 1 14 43 PM

But they both could look like this:
Screen Shot 2022-06-18 at 1 21 57 PM

micahellison (approved):
Thanks for these changes, and for making it more consistent.

PR #1511, opened 2022-06-14
Check for duplicate keys in config file

### LINKED ISSUE #1415: Warn user when there are duplicate keys in the config file

jonakeys:
This PR solves issue #1415 It checks if a config file contains duplicate keys at the same level in the config file, warns the user about that and names the keys it concerns. The program still runs as usual (i.e. no exception is thrown) as stated in the requirements in the issue.
I also added BDD tests for the different cases.

Edit:
I chose to exclude old .json files, because those won't probably be used a lot anymore and besides that: after one run the config will be upgraded to .yaml anyway, so the check will be run.

Fixes #1415 

### Checklist

  for the same issue.


wren (changes requested):
Great addition of tests, and thenk you for taking on this issue, but the implementation itself is a bit brittle here. Instead of parsing the yaml file ourselves, can we make use of the yaml parsing library that jrnl already has loaded (ruamel)?

It looks like we already use `yaml.allow_duplicate_keys` [in config.py](https://github.com/jrnl-org/jrnl/blob/develop/jrnl/config.py#L165), so maybe we can make use of that? It might work to set `yaml.allow_duplicate_keys = False`, catch the exception, then warn the user about it, and reload with `yaml.allow_duplicate_keys = True` to be able to continue.

This would avoid reinventing a yaml parser, and give us a more robust check.

What do you think?

jonakeys:
Thanks for your comments, I really appreciate it!
The solution you propose indeed is simpler/more robust in checking for duplicate keys. Ruamel can check for duplicate keys and than we can act on that. The reason I choose to find another way is because the issue said:

> Maybe the warning could say something like: "Warning: the [key name here] key appears twice at the same level in your configuration file."

And ruamel only returns _that_ the file has duplicates, but not _which_ keys are duplicate. I thought it would be helpful to give the user information about the key(s) it concerns.

If it's ok to warn the user the file has duplicates, but not mentioning them specifically, I'm happy to rewrite the code in the way you described it.

wren:
> If it's ok to warn the user the file has duplicates, but not mentioning them specifically, I'm happy to rewrite the code in the way you described it.

Hrmm, that's a good point. Yes, I think that the increased robustness is worth the more vague warning. That's a shame that the exception doesn't have more info, but if that's the best we've got, then I think it's still worth using the library.

wren:
@jonakeys Actually, I just did a quick test on this with `yaml.allow_duplicate_keys = False` and I see the following exception, which seems to have all the info we want:

```
┏━ Error ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃  DuplicateKeyError                                                                                                                                              ┃
┃          while constructing a mapping                                                                                                                           ┃
┃    in "/Users/jonathan/Projects/Jrnl/test.yaml", line 14, column 5                                                                                              ┃
┃  found duplicate key "journal" with value "~/Projects/Jrnl/test-journals/basic_onefile.journal" (original value:                                                ┃
┃  "~/Projects/Jrnl/test-journals/basic_onefile.journal")                                                                                                         ┃
┃    in "/Users/jonathan/Projects/Jrnl/test.yaml", line 15, column 5                                                                                              ┃
┃                                                                                                                                                                 ┃
┃  To suppress this check see:                                                                                                                                    ┃
┃      http://yaml.readthedocs.io/en/latest/api.html#duplicate-keys                                                                                               ┃
┃                                                                                                                                                                 ┃
┃                                                                                                                                                                 ┃
┃          This is probably a bug. Please file an issue at:                                                                                                       ┃
┃          https://github.com/jrnl-org/jrnl/issues/new/choose                                                                                                     ┃
┗━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┛

```

Are you seeing something different?

jonakeys:
The output you get is the output I get now myself too.. I can't recall seeing this when I started working on the issue. This is indeed enough information and a solution is close by. I'll see how to catch the exception and load the config after that.
Thanks for your ideas and thinking!

jonakeys:
@wren I reworked the code. The warning message now looks like this:

```
(jrnl-aMUFJMGM-py3.9) jonathan@gentoo ~ $ jrnl -1

┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃  Warning: One or more keys appear multiple times at the same level       ┃
┃          in your configuration file.                                     ┃
┃                                                                          ┃
┃          while constructing a mapping                                    ┃
┃    in "/home/jonathan/.config/jrnl/jrnl.yaml", line 2, column 3          ┃
┃  found duplicate key "title" with value "CYAN" (original value: "CYAN")  ┃
┃    in "/home/jonathan/.config/jrnl/jrnl.yaml", line 6, column 3          ┃
┃                                                                          ┃
┃  To suppress this check see:                                             ┃
┃      http://yaml.readthedocs.io/en/latest/api.html#duplicate-keys        ┃
┗━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┛
2022-06-09 20:37 this is a new entry
```



I can't get the second line 'in your configuration...' start at the same indent as the first line. Do you know how to fix this?

wren:
@jonakeys Thank you!

Since there's a beta out right now (v3.0), we're going to wait to fully review and merge this after that version is released. Thanks for your hard work on it! 👍

jonakeys:
Glad I can contribute a little to this project!
I understand the v3.0 is around the corner, so it's wise to focus on that release.

micahellison:
> I can't get the second line 'in your configuration...' start at the same indent as the first line. Do you know how to fix this?

I was just talking over this w/ @wren about this now, who will fix this in a separate PR for all messages by dedenting before building the message with its arguments.

wren (dismissed):
Looks great. Thank you!

micahellison (approved):

PR #1513, opened 2022-06-18
Add bdd tests for jrnl installation

### LINKED ISSUE #1473: Add installation BDD tests
### Use Case/Motivation

We've added tons of BDD tests to jrnl over the years, but it looks like we missed tests for the initial installation step that runs when no config is found.

It would be good to have tests for cases in a new `install.feature` file like:
- [ ] setting the config path correctly based on various parameters, including the tilde (~)
- [ ] confirming that the directory of the new journal was created if it didn't exist before
- [ ] checking if the journal was encrypted based on whether the user wants to encrypt the journal
- [ ] confirming that the `version` config key was set to the current version

### Other information

I'm not sure if this is just a bunch of Gherkin in a new feature file, or if this will also need some changes in `/tests/lib` to deal with the lack of config file.

apainintheneck:
### Checklist

  for the same issue.

This is just a first stab at adding BDD tests related to installation as mentioned in #1473. I couldn't really find a way to make it work with just Gherkin because of some of the required mocking and steps that were added. Is that sorta what you had in mind?

The biggest change I made was to use the pipe character `|` as the equivalent of an empty line of input in the `all_inputs` fixture used by the `when_steps.we_run_jrnl()` method. This change made it a lot easier to test the default journal installation case and I think it makes the BDD test more readable too. Do you all know of a better way to handle that?

Resolves #1473.


tests/bdd/features/install.feature
```diff
@@ -0,0 +1,45 @@
+Feature: Installing jrnl
+
+    Scenario: Install jrnl with default options
+        Given we use no config
+        When we run "jrnl hello world" and enter
+            |
+            |
```
micahellison: Could you use `\n` for this instead of a pipe? I think that will be a bit more Gherkin-appropriate since the [Gherkin reference](https://cucumber.io/docs/gherkin/reference/) has some allowances for this inside of table cells already.

wren:
@apainintheneck Hey, this PR is looking good so far; we just need a little more time to review (had a lot of PRs today, and focusing on the beta). We'll come back to this next week and get merged in. Thank you!

tests/bdd/features/install.feature
apainintheneck: Yeah, I can do that. I can see how that might be confusing.

tests/lib/fixtures.py
```diff
@@ -216,7 +231,9 @@ def mock_console_input(**kwargs):
                 return password_input
 
             if isinstance(user_input, Iterable):
-                return next(user_input)
+                Input_line = next(user_input)
+                # A raw newline is used to indicate deliberate empty input
+                return "" if Input_line == r"\n" else Input_line
```
micahellison: Sorry to be pedantic, but could you rename `Input_line` to `input_line` since it's a local variable?

micahellison (changes requested):
Just one tiny case change, please. Thanks for adding these much-needed tests.

tests/lib/fixtures.py
micahellison: Never mind, I'll just do this, since it's such a tiny change.

micahellison (approved):

PR #1522, opened 2022-06-28
Fixed index out of range error in fancy exporter

apainintheneck:
### Checklist

  for the same issue.

This is just a small bug that shows up when trying to print entries using the fancy exporter that lack titles. This error doesn't show up with any of the other `--format` methods as far as I can tell.

#### Steps to reproduce
1. Use `jrnl edit` to add a post without a title.
3. Enter `jrnl -1`.

This change just makes sure that the `title_lines` list always includes at least one empty string.

### DIFF
--- a/jrnl/plugins/fancy_exporter.py
+++ b/jrnl/plugins/fancy_exporter.py
@@ -50,7 +50,7 @@ class FancyExporter(TextExporter):
             subsequent_indent=cls.border_g + " ",
         )
 
-        title_lines = w.wrap(entry.title)
+        title_lines = w.wrap(entry.title) or [""]
         card.append(
             title_lines[0].ljust(initial_linewrap + 1)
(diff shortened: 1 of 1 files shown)


micahellison:
~Hi @apainintheneck, I'm having a time running the steps to reproduce. Every time I edit my journal, it keeps a title of some sort, or folds the entry into the previous entry. Could you let me know what specific text you're entering into your editor to get this to work? Thanks.~
Never mind! Just figured out - the entry needs to be completely blank except for a space after the timestamp.

Could you add a test for this PR in `format.feature`?

apainintheneck:
I was unable to find a way to reproduce this bug with a BDD test. Not sure if it's really necessary either since it's such a small thing.

wren (approved):
Ok, that makes sense. Thanks!

PR #1524, opened 2022-06-29
Add message showing the number of search results

### LINKED ISSUE #1454: Report number of entries found in a search
### Use Case/Motivation

Right now, if you search for entries in any way (tags, latest # of entries, -from or -to, -contains, etc.) it will simply return the results, or return nothing if no results are found.

However, it's pretty standard behavior for anything with a search feature to report the number of records found. And we already do something like this after editing an entry (it shows the number of entries created, modified, or deleted).

Moreover, this could be a nice way of adding some clearer feedback when nothing is found. For instance, if I run `jrnl -contain aWordThatINeverUse` I get no information from jrnl instead of a message telling me it couldn't find anything.

### Example Usage

```
> jrnl yesterday: I did this yesterday

> jrnl today: This happened today

> jrnl -from yesterday
[2 entries found]
2022-04-22 09:00 I did this yesterday

2022-04-23 09:00 This happened today

> jrnl -from tomorrow
[no entries found]
```
I think it's reasonable to expect this in `--edit` mode too:
```
> jrnl --edit -from yesterday
[2 entries found]
```
Then jrnl waits until I finish my edits, after which I'd expect the usual message when edits are finished, such as:
```
[1 entry modified]
```

### Other information

Whatever the resolution, it's absolutely critical that these messages go to STDERR and *not* STDOUT, since STDOUT is for journal data, not messages about jrnl's operation.

Could be nice to add some color to these messages as well.

apainintheneck:
### Checklist

  for the same issue.

This is a first stab at #1454. The basic idea was to include a message stating how many entries were found when searching.

#### Changes
1. Renamed `_search_journal()` -> `_filter_journal_entries()` to remove the need for a comment explaining what it does.
2. Added `_has_search_args()` which checks if there are any search arguments provided by the user.
3. Added `_print_entries_found_count()` which prints a message based upon the number of entries found and prints warnings if the user was trying to edit or delete and there are no entries.

I decided that it probably doesn't make sense to output the number of entries found when the user is just changing the timestamps on or deleting entries. Currently it only displays that information if the user is editing or searching.

I'm also not sure if this needs new tests but extra checks could probably be added to the current BDD tests that reflect these changes. I just decided to wait on that until I got some feedback.

### DIFF
--- a/jrnl/jrnl.py
+++ b/jrnl/jrnl.py
@@ -173,8 +173,9 @@ def search_mode(args, journal, **kwargs):
         "old_entries": journal.entries,
     }
 
-    # Filters the journal entries in place
-    _search_journal(**kwargs)
+    if _has_search_args(args):
+        _filter_journal_entries(**kwargs)
+        _print_entries_found_count(len(journal), args)
 
     # Where do the search results go?
     if args.edit:
@@ -195,6 +196,10 @@ def search_mode(args, journal, **kwargs):
             journal.entries = selected_entries
 
         _edit_search_results(**kwargs)
+
+    elif not journal:
+        # Bail out if there are no entries and we're not editing
+        return
 
     elif args.change_time:
         _change_time_search_results(**kwargs)
@@ -244,8 +249,28 @@ def _get_editor_template(config, **kwargs):
     return template
 
 
-def _search_journal(args, journal, **kwargs):
-    """Search the journal with the given args"""
+def _has_search_args(args):
+    return any(
+        (
+            args.on_date,
+            args.today_in_history,
+            args.text,
+            args.month,
+            args.day,
+            args.year,
+            args.start_date,
+            args.end_date,
+            args.strict,
+            args.starred,
+            args.excluded,
+            args.contains,
+            args.limit,
+        )
+    )
+
+
+def _filter_journal_entries(args, journal, **kwargs):
+    """Filter journal entries in-place based upon search args"""
     if args.on_date:
         args.start_date = args.end_date = args.on_date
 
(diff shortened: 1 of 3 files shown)


micahellison:
Thanks for the PR @apainintheneck. This implementation looks great to me, though it's noisier than I anticipated. I think the main issue is when I use the `-n` argument -- if I run `jrnl -n 1` or its equivalent `jrnl -1`, I don't need to know that 1 entry was found. However, it could be useful when there are _less_ entries returned than the `-n` value.

Could you suppress this message when both 1) the `-n` argument is used and 2) the number of entries isn't equal to whatever is supplied to `-n`?

apainintheneck:
I get your point that the output is a bit a noisy when `-n` arg is passed to limit the output. That being said I'd almost be more inclined to just omit showing the number of results when that is passed as an option instead of sometimes showing it and sometimes not.

micahellison:
> I get your point that the output is a bit a noisy when `-n` arg is passed to limit the output. That being said I'd almost be more inclined to just omit showing the number of results when that is passed as an option instead of sometimes showing it and sometimes not.

That's a good point. Keeping it simple is nice, and if it ever does become a big issue, we can always change it later.

wren:
@apainintheneck Thanks for your patience! This looks good, just need to add the tests you mentioned and we should be good to go.

apainintheneck:
Okay, I added to some of the existing tests. Let me know what you think.

micahellison (approved):
Looks great to me! Thanks for adding these tests.

wren (approved):
Looks great. Thank you!

PR #1527, opened 2022-07-01
Export to file(s) when first line/title of an entry is very long

### LINKED ISSUE #1272: Exporting Entries with Long First Lines Results in "File name too long" Error
### Environment

```
jrnl: v2.7
Python: 3.9.5 (default, May 24 2021, 12:50:35) 
[GCC 11.1.0]
OS: Linux 5.10.42-1-MANJARO
```

Installed via pipx on Manjaro

### Current Behavior
I am attempting to export an old journal to individual file names, but the first lines in some of these are very long. Running an export with ```jrnl $JOURNAL_FILE --format markdown --file $EXPORT/``` results in the following:

```
[ERROR: $EXPORT/$DATE_$THISFILENAMEISTOOLONG.md File name too long]
```

### Expected Behavior
jrnl should have a limit on exports for when it hits the maximum file name characters so that the export may proceed. The file name should be capped at the maximum characters.

Ideally, in the rare event that an entry is found that already exists due to this capping, having the same name, it should:

* create a new file with a number string appended, rather than overwriting the other file, to maintain both files,
* or prompt the user to review which file to keep or offer to keep both, with a number string appended.

### Repro Steps
* Create a journal entry with a long first line.
* Attempt to export it.

jonakeys:
This PR addresses issue #1272 When a journal has a long first line/title, it now exports to a (truncated) file name without throwing an error.
Also added tests for markdown and text formats.

### Checklist

  for the same issue.

Fixes #1272


jonakeys:
The tests on Windows failed, so back to the drawing board it is.

jonakeys:
All checks have passed! PR is open for reviewing.

tests/data/journals/long_title.journal
```diff
@@ -0,0 +1,9 @@
...
+Nec dui nunc mattis enim ut tellus elementum sagittis vitae. Id porta nibh venenatis cras sed felis. Id leo in vitae turpis. Lorem ipsum dolor sit amet consectetur adipiscing elit pellentesque. Dignissim enim sit amet venenatis urna cursus eget nunc scelerisque. Donec ac odio tempor orci dapibus ultrices in iaculis nunc. Velit laoreet id donec ultrices tincidunt arcu non. Egestas diam in arcu cursus euismod quis. Dolor morbi non arcu risus quis. Nisi lacus sed viverra tellus in hac habitasse platea. Mattis enim ut tellus elementum sagittis vitae et leo. Non pulvinar neque laoreet suspendisse interdum. Libero volutpat sed cras ornare arcu dui vivamus.
```
micahellison: Could you re-write the tests to use the same series of `basic_` config files as the others, then use the "When we run" step to create the long entries, rather than adding this file? That way, we can keep our test data down while also ensuring that this works with every type of journal.

tests/lib/helpers.py
```diff
@@ -17,6 +18,16 @@ def does_directory_contain_files(file_list, directory_path):
     return True
 
 
+def does_directory_contain_n_files(directory_path, number):
```
micahellison, line 20: Could you ensure this only checks the number of files, and doesn't include any subdirectories?

micahellison (changes requested):
Hi @jonakeys, thanks for working on this and for your patience in waiting for a review. I've got just a couple small nitpicks.

tests/lib/helpers.py
jonakeys: Check!

tests/data/journals/long_title.journal
jonakeys: Thanks for pointing that out, better that way! Resolved it by removing unneeded changes to config files/journals.

micahellison (approved):
Looks great to me! Thanks for making the changes.

PR #1563, opened 2022-08-23
Add hint for how to get vi to go to end-of-file

pjz:
### Checklist

- [NA] I have included a link to the relevant issue number.
- [NA] I have checked to ensure there aren't other open [pull requests](../pulls)
  for the same issue.
- [NA] I have written new tests for these changes, as needed.

Added a hint for vi users to the docs.

### DIFF
--- a/docs/tips-and-tricks.md
+++ b/docs/tips-and-tricks.md
@@ -211,3 +211,12 @@ display_format: markdown
 
 For more information on how `jrnl` outputs your entries in Markdown, please visit the [Formats](./formats.md) section.
 
+
+## Jump to end of buffer (with vi)
+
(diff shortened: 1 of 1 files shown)


wren (changes requested):
Hi, and thank you for submitting this tip (it seems very useful)!

I think this should go into the Tips & Tricks section, though, since the External Editors section is more about getting editors to work with `jrnl` (which `vi` already does without any additional flags).

Also, we don't have an explicit style guide, but could you try to match the tone of the rest of the page? In the case of this type of tip, we try to stick with a "How To" style, rather the more Editoral style that this currently has. If you need examples, please look at the other tips on that same page.

wren (approved):
Thank you!

PR #1564, opened 2022-08-23
Fix #1486

### LINKED ISSUE #1486: --format list and error output is incorrect for importing
### Environment
  - `jrnl --diagnostic` output: 
```
jrnl: v2.8.4
OS: Windows 10
```
  - Install method:  pipx

### Current Behavior / Repro Steps / Expected behavior

The list of formats outputted by jrnl is the same whether we're talking about importing or exporting. To see this in action, enter in a nonsensical import format like so:
```
>jrnl --import --format thiswillbreak
usage: jrnl [--debug] [--help] [--version] [--list] [--encrypt] [--decrypt] [--import] [-on DATE] [-today-in-history] [-month DATE] [-day DATE]
            [-year DATE] [-from DATE] [-to DATE] [-contains TEXT] [-and] [-starred] [-n [NUMBER]] [-not [TAG]] [--edit] [--delete]
            [--format TYPE] [--tags] [--short] [--config-override CONFIG_KV_PAIR CONFIG_KV_PAIR] [--config-file CONFIG_FILE_PATH]
            [...]
jrnl: error: argument --format: invalid choice: 'thiswillbreak' (choose from 'boxed', 'dates', 'fancy', 'json', 'markdown', 'md', 'pretty', 'sample', 'short', 'tags', 'text', 'txt', 'xml', 'yaml')
```
All those formats that jrnl offers are export formats and none are import formats (the only import format is jrnl), so that message shouldn't appear.

If you try any of those formats, jrnl crashes with a traceback like so:
```
>jrnl --import --format json
  File "C:\Users\micah\.pyenv\pyenv-win\versions\3.9.7\lib\runpy.py", line 197, in _run_module_as_main
  File "C:\Users\micah\.pyenv\pyenv-win\versions\3.9.7\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "C:\Users\micah\.local\pipx\venvs\jrnl\lib\site-packages\jrnl\cli.py", line 34, in cli
    return run(args)
  File "C:\Users\micah\.local\pipx\venvs\jrnl\lib\site-packages\jrnl\jrnl.py", line 53, in run
    return args.postconfig_cmd(
  File "C:\Users\micah\.local\pipx\venvs\jrnl\lib\site-packages\jrnl\commands.py", line 57, in postconfig_import
    get_importer(format).import_(journal, args.filename)
AttributeError: 'NoneType' object has no attribute 'import_'
```
It should show a user-friendly error message instead, probably explaining that this is only an export format, not an import format.

### Other Information

Found this out after spotting discussion #1476.

pjz:
Fix #1486 

Instead of --import paying attention to --format, leave --format
just for the export-format and give --import itself an optional
arguent that is the import format.  This of course defaults to 'jrnl'.
This allows argparse to do the choice-checking.

### Checklist

  for the same issue.

### DIFF
--- a/jrnl/args.py
+++ b/jrnl/args.py
@@ -121,18 +121,19 @@ def parse_args(args=[]):
     )
     standalone.add_argument(
         "--import",
-        action="store_const",
         metavar="TYPE",
-        const=postconfig_import,
-        dest="postconfig_cmd",
+        dest="import_format",
+        nargs="?",
+        choices=IMPORT_FORMATS,
+        default="jrnl",
         help=f"""
         Import entries from another journal.
+
+        TYPE is the format to import [{util.oxford_list(IMPORT_FORMATS)}] (default: jrnl)
 
         Optional parameters:
 
         --file FILENAME (default: uses stdin)
-
-        --format [{util.oxford_list(IMPORT_FORMATS)}] (default: jrnl)
         """,
     )
     standalone.add_argument(
--- a/jrnl/commands.py
+++ b/jrnl/commands.py
@@ -69,8 +69,7 @@ def postconfig_import(args, config, **kwargs):
     # Requires opening the journal
     journal = open_journal(args.journal_name, config)
 
-    format = args.export if args.export else "jrnl"
-    get_importer(format).import_(journal, args.filename)
+    get_importer(args.import_format).import_(journal, args.filename)
 
 
 def postconfig_encrypt(args, config, original_config, **kwargs):


micahellison:
Hi @pjz, thanks for the PR. While I can understand the reasoning for this, I don't think the resolution should change the structure of the arguments, since it's not a big enough issue to justify an API change. Instead, I think it would be ideal if it showed the correct information about the arguments and didn't crash when using the arguments incorrectly.

pjz:
That would mean that jrnl has to do the choice checking to make sure that --format is valid with --import.  Additionally. there would be no way to specify a separate import from export style, since --format would control both.

micahellison:
> That would mean that jrnl has to do the choice checking to make sure that --format is valid with --import. Additionally. there would be no way to specify a separate import from export style, since --format would control both.

Yes, agreed. It means handling some extra logic outside of the typical argparse path. A little extra work, but it preserves the API for the user. I think a solution would need cross-argument validation as documented in [this SO article](https://stackoverflow.com/questions/49000217/cross-argument-validation-in-argparse). And the `--format` documentation could list both sets of formats.

micahellison:
I'm closing this since it's getting stale, but feel free to re-open if you'd like to continue working on it.

PR #1592, opened 2022-09-08
Add machine readable --list output

### LINKED ISSUE #1445: Add machine-readable format for --list
### Use Case/Motivation

Currently, the only guaranteed way to know the location of your config file and journals is to run `jrnl --list`. Without that command, you'd have to either reimplement the [basedir XDG spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) or use something simpler that's specific to your machine's configuration. Not ideal.

However, the command's output is optimized for human reading and doesn't really have a guarantee that it can be consistently parsed by a machine. We could get around this by adding an argument to view its output in a machine-readable format like JSON or YAML.

The end of #1444 mentions using a feature like this to discover journals when diffing/merging encrypted journals. I think it could be useful for a wide variety of situations, for instance, an "application-aware" backup that backs up configuration along with journal data.

### Example Usage

```
jrnl --list --format json
```
```
jrnl --list --format yaml
```

### Other information

Just a minor note that --format is also used to format journals, but would have a very different meaning in this case. If you disagree with the use of the term "format" here, feel free to chime in with other suggestions.

apainintheneck:
### Checklist

  for the same issue.

This closes #1445.

The following options are now available for the list command. If another format option is chosen that isn't yaml or json, it just returns the default list.

```sh
jrnl --list
jrnl --list --format json
jrnl --list --format yaml
```

### DIFF
--- a/jrnl/output.py
+++ b/jrnl/output.py
@@ -25,30 +25,48 @@ def deprecated_cmd(old_cmd, new_cmd, callback=None, **kwargs):
         callback(**kwargs)
 
 
+def journal_list_to_json(journal_list):
+    import json
+
+    return json.dumps(journal_list)
+
+
+def journal_list_to_yaml(journal_list):
+    from io import StringIO
+
+    from ruamel.yaml import YAML
+
+    output = StringIO()
+    YAML().dump(journal_list, output)
+    return output.getvalue()
+
+
+def journal_list_to_stdout(journal_list):
+    result = f"Journals defined in config ({journal_list['config_path']})\n"
+    ml = min(max(len(k) for k in journal_list["journals"]), 20)
+    for journal, cfg in journal_list["journals"].items():
+        result += " * {:{}} -> {}\n".format(
+            journal, ml, cfg["journal"] if isinstance(cfg, dict) else cfg
+        )
+    return result
+
+
 def list_journals(configuration, format=None):
     from jrnl import config
 
     """List the journals specified in the configuration file"""
+
+    journal_list = {
+        "config_path": config.get_config_path(),
+        "journals": configuration["journals"],
+    }
+
     if format == "json":
-        import json
-
-        return json.dumps(configuration["journals"])
+        return journal_list_to_json(journal_list)
     elif format == "yaml":
-        from io import StringIO
-
-        from ruamel.yaml import YAML
-
-        output = StringIO()
-        YAML().dump(configuration["journals"], output)
-        return output.getvalue()
+        return journal_list_to_yaml(journal_list)
     else:
-        result = f"Journals defined in config ({config.get_config_path()})\n"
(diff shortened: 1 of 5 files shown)


jrnl/output.py
```diff
@@ -25,17 +25,30 @@ def deprecated_cmd(old_cmd, new_cmd, callback=None, **kwargs):
...
+        YAML().dump(configuration["journals"], output)
+        return output.getvalue()
+    else:
+        result = f"Journals defined in config ({config.get_config_path()})\n"
+        ml = min(max(len(k) for k in configuration["journals"]), 20)
+        for journal, cfg in configuration["journals"].items():
+            result += " * {:{}} -> {}\n".format(
+                journal, ml, cfg["journal"] if isinstance(cfg, dict) else cfg
+            )
+        return result
```
micahellison: I see the value in the simplicity of this approach, though I would hope to see the same information regardless of what format is used. In particular, the nested configuration dictionary doesn't flatten the journal paths between journals specified using "journalname: path" and "journal: journalname: path". And while the old approach already combined the concerns of data retrieval and data presentation, it would be good to keep those separate if we're expanding the feature.

Instead, could you try retrieving the journal data first (sort of like the code in the `else` block does), then sending that data to different export functions based on the file format? Probably best as a dictionary object. That way it's all working with the same data while respecting the separation of concerns.

jrnl/output.py
```diff
@@ -25,17 +25,30 @@ def deprecated_cmd(old_cmd, new_cmd, callback=None, **kwargs):
...
+    elif format == "yaml":
+        from io import StringIO
+
+        from ruamel.yaml import YAML
+
+        output = StringIO()
+        YAML().dump(configuration["journals"], output)
+        return output.getvalue()
+    else:
+        result = f"Journals defined in config ({config.get_config_path()})\n"
```
micahellison: Could you include this config path information in the other exports as well? I think `config_path` is a good key name for it.

micahellison (changes requested):
Thanks for the PR! Just a couple notes about separation concerns in the review. Also, let us know if you'd like any help with adding tests for this.

jrnl/output.py
apainintheneck: So something like this with a different function for each output format?

```python
journal_list = {
  "config_path" : config.get_config_path(),
  "journals" : configuration["journals"],
}

if format == "json":
  return journal_list_to_json(journal_list)
elif format == "yaml":
  return journal_list_to_yaml(journal_list)
else:
  return journal_list_to_stdout(journal_list)
```

micahellison: Yes, exactly, that looks great!

micahellison (approved):
Thank you! Looks great!

PR #1594, opened 2022-09-10
Change default config to use journal key

### LINKED ISSUE #1202: Change default config to use journal key
### Use Case/Motivation

When a user runs jrnl for the first time, it creates a config file with a default journal based on the user's input. That shows up under the `journals` key as `default: (path)` like so:
```
journals:
  default: C:\Users\micah\journal.txt
```

This is fine, except it can be a bit confusing for users that want to add per-journal configuration elements, e.g. different templates for different journals as seen in #1152. It would be easier to modify if the `journal` key was used instead, like so:

```
journals:
  default:
    journal: C:\Users\micah\journal.txt
```

micahellison:
Fixes #1202.

### Checklist

  for the same issue.

### DIFF
--- a/jrnl/config.py
+++ b/jrnl/config.py
@@ -91,7 +91,7 @@ def get_config_path():
 def get_default_config():
     return {
         "version": __version__,
-        "journals": {"default": get_default_journal_path()},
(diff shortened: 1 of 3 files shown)


wren (approved):
📒

PR #1595, opened 2022-09-10
Change default timestamp to `%F %r`

### LINKED ISSUE #1100: Change default timeformat config to '%F %r'
### Use Case/Motivation

There are a variety of open issues related to the way jrnl stores and displays date/time formats. We're looking at the possibility of separating date/time storage from date/time display (#1098) and parsing dates more strictly (#630) and supporting time zones (#825).

As we implement these changes, it might help users to have more granular date/time data to upgrade from when it's time to make the switch. One way we could do this is by changing the default `timeformat` to something a bit more specific, perhaps including seconds and time zone / offset if that's supported.

wren:
Fixes #1100

### Checklist

  for the same issue.

### DIFF
--- a/jrnl/config.py
+++ b/jrnl/config.py
@@ -97,7 +97,7 @@ def get_default_config():
         "template": False,
         "default_hour": 9,
         "default_minute": 0,
-        "timeformat": "%Y-%m-%d %H:%M",
+        "timeformat": "%F %r",
         "tagsymbols": "#@",
         "highlight": True,
         "linewrap": 79,


micahellison (approved):
⌚

laikq:
Hej, I have just installed `jrnl` and seen that it uses `%F %r` as default timeformat, but neither `%F` nor `%r` are explained in the [linked python documentation](https://docs.python.org/3/library/time.html#time.strftime) in the [jrnl docs](https://jrnl.sh/en/stable/reference-config-file/#timeformat). I also cannot find any documentation for `%F`/`%r` in the python `datetime` docs.

(Asking because maybe there are more undocumented format keys which I could use, as I like to use ISO week numbers in my timestamps. Not sure if this is the appropriate place to discuss, I can also open a new issue.)

PR #1600, opened 2022-09-20
Don't create empty file when attempting a YAML export to a non-existing folder

### LINKED ISSUE #1593: Zero-length file created when attempting to export YAML to a directory that does not exist
### Diagnostic output

```
jrnl: v3.2
OS: Windows 10
```

### Current Behavior

In a directory where there is no file or subdirectory named "test", run:
```
>jrnl --format yaml --file test
```
I see the error message as expected:
```
┏━ Error ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃  YAML export must be to a directory, not a single file  ┃
┗━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┛
```
But then I find an empty file named "test" in that directory.

### Expected Behavior

jrnl should give me an error message and not do anything esle.

### Repro Steps

In a directory without a file or folder named "test":
```
jrnl --format yaml --file test
```

### Debug output

```
DEBUG    root         Parsed args: Namespace(debug=True, preconfig_cmd=None, postconfig_cmd=None, filename='test', on_date=None, today_in_history=False, month=None, day=None, year=None, start_date=None, end_date=None, contains=None, strict=False, starred=False, limit=None, excluded=[], edit=False, delete=False, change_time=None, export='yaml', tags=False, short=False, config_override=[], config_file_path='', text=[])
DEBUG    root         Using configuration "{'colors': {'body': 'none', 'date': 'black', 'tags': 'none', 'title': 'none'}, 'default_hour': 9, 'default_minute': 0, 'display-format': 'markdown', 'editor': 'subl -w', 'encrypt': False, 'highlight': True, 'indent_character': '', 'journals': {'default': {'journal': 'C:\\Users\\micah\\Documents\\jrnls\\default.txt', 'linewrap': 80}}}, 'linewrap': 79, 'tagsymbols': '@', 'template': False, 'timeformat': '%Y-%m-%d %H:%M', 'version': 'v2.8.4'}"
DEBUG    root         Updating configuration with specific journal overrides {'journal': 'C:\\Users\\micah\\Documents\\jrnls\\default.txt', 'linewrap': 80}
DEBUG    root         opened PlainJournal with 34 entries
┏━ Error ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃  YAML export must be to a directory, not a single file  ┃
┗━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┛
```

### Other Information

_No response_

outa:
### Changes
call `export_journal` which raises the exception before opening the file handle so that no empty file is created

Also I thought I'd add a unit test for that situation, but I don't quite understand how to do that (see the changes in test_export.py; the test fails currently). Would much appreciate some guidance on this.

Closes: #1593

### Checklist

  for the same issue.

### DIFF
--- a/jrnl/plugins/text_exporter.py
+++ b/jrnl/plugins/text_exporter.py
@@ -31,18 +31,19 @@ class TextExporter:
     @classmethod
     def write_file(cls, journal, path):
         """Exports a journal into a single file."""
+        export_str = cls.export_journal(journal)
         with open(path, "w", encoding="utf-8") as f:
-            f.write(cls.export_journal(journal))
-            print_msg(
-                Message(
-                    MsgText.JournalExportedTo,
-                    MsgStyle.NORMAL,
-                    {
-                        "path": path,
-                    },
-                )
+            f.write(export_str)
+        print_msg(
+            Message(
+                MsgText.JournalExportedTo,
+                MsgStyle.NORMAL,
+                {
+                    "path": path,
+                },
             )
-            return ""
+        )
+        return ""
 
     @classmethod
     def make_filename(cls, entry):
--- a/tests/unit/test_export.py
+++ b/tests/unit/test_export.py
@@ -1,14 +1,14 @@
 # Copyright © 2012-2022 jrnl contributors
 # License: https://www.gnu.org/licenses/gpl-3.0.html
 
-from pathlib import Path
-from tempfile import TemporaryDirectory
+from unittest import mock
 
 import pytest
-from ruamel.yaml import YAML
 
 from jrnl.exception import JrnlException
-from jrnl.Journal import PlainJournal
+from jrnl.messages import Message
+from jrnl.messages import MsgStyle
+from jrnl.messages import MsgText
 from jrnl.plugins.fancy_exporter import check_provided_linewrap_viability
 from jrnl.plugins.yaml_exporter import YAMLExporter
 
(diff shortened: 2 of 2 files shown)


wren:
Hi, and thanks for the PR. For the unit test, one way to do it would be this:

1. Patch `cls.export_journal` and `open` (fee here for more info about how to patch: https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch)
2. Force `cls.export_journal` to have a `side_effect` that raises an exception (see here for more info about side effects: https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.side_effect)
3. Run `write_file` method
4. Assert that `open` was never called (probably something like `mock_open.assert_not_called()`: https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_not_called)

That way, you can avoid having to make a new fixture, or write to any files.

Does that makes sense?

outa:
Hi Jonathan, thanks for the explanation! The test passes now.

However I am passing a dummy string to `write_file` now instead of an actual Journal class. Not sure if that's ok?

wren (approved):
Looks good. Thank you!

Issue #1521, opened 2022-06-25
Don't import cryptography package if not needed


micahellison:
### Use Case/Motivation

I noticed the 3.0 beta is running a bit slower than 2.8.4 so I ran some performance testing. It looks like the cryptography package is the slowest to load, and it's taking a lot longer to load on my system with the 3.0 beta than with 2.8.4.

However, cryptography is only needed for encrypted journals. Maybe there's a way to keep it from loading unless it's absolutely needed.

### Other information

See the description of PR #719 for code on testing imports.

SepehrRasouli:
Hi @micahellison, I think this problem could be solved via importing it in an if statement, For example, we can see if the user is using an encrypted journal or not via their config file, if yes, import the library, otherwise continue the program's flow.

Issue #1248, opened 2021-05-22
New encryption method


micahellison:
### Use Case/Motivation

jrnl's current encryption method is getting pretty old, and there's been a lot of innovation out there. Should we consider moving to something new?

Right now we're looking at [Ed25519 encryption](https://ed25519.cr.yp.to/), which is pretty well regarded and performant, but we're open to other ideas and suggestions.

Issue #1149, opened 2021-01-04
Implement proper salts in encrypted journals


KarimPwnz:
Since we're discussing changing the storage format in #1107, perhaps we could implement proper salting. Currently, encrypted journals use a hard-coded salt that is passed into `PBKDF2HMAC`. However, it is not technically a salt but a pepper: a salt is unique for every encryption.

To implement this, we should store the unencrypted, base64-encoded salt in the journal file—it is not meant to be secret.

wren:
This is a good point to bring up. I'm completely open to discussion, so this seems like a good place to do it.

You're absolutely right that the current implementation isn't really a salt. It's fixed for all journals, and in no way increases the difficulty of a pre-computed attack on a journal file.

In the past, I've sided against moving the salt into the config file (to make it actually a salt) for a few reasons:
1. Moving the salt (or even getting rid of it) would break backwards compatibility with jrnl's current storage format. So, then we would have a v1, v2, and v3 storage format to deal with. Not necessarily a deal-breaker, but I would want to see some concrete advantages before committing jrnl to a larger maintenance burden, and increased user confusion.
2. Currently, a user only needs their password in order to be able to decrypt jrnl files. If the salt is moved to the config, then users would need their password and the config file to do the same thing. This isn't bad in and of itself, but it is more difficult to keep track of two things than it is to keep track of one. And historically, keeping track of key files along with passwords has proven to be too difficult for most non-technical users (and many technical ones). For example, look at PGP and it's waning popularity in the security community. If we do use a salt, I would strongly prefer to store it in the journal file itself.
3. I believe this would significantly complicate managing journals in jrnl both for users and for maintainers.
    * For users: a user can currently run `--decrypt` and `--encrypt` over and over again with the same password, and be confident they'll always be able to decrypt that journal with that password, even if they restore from a backup. If there's a salt in the config, it would change each time `--encrypt` is run. So now a user has to keep track of exactly when a backups was taken, and sync the config file with the journal itself. And since the config is most often kept in a separate location from the journal files, we risk a user not backing up the config, and rendering their backups useless.
[...]

Issue #1021, opened 2020-08-05
Can't save password to keychain when running --encrypt on already-encrypted journal


iedd:
### Affected Page(s)
https://jrnl.sh/encryption/#storing-passwords-in-your-keychain

### What Could Be Better?
In the situation where the user has generated a new keychain in place of the old one, it is not sufficient to simply run jrnl --encrypt on an already-encrypted journal. The user must run jrnl --decrypt before encrypting again, in order to access the prompt to store the password on the keychain.

Thanks for your hard work and for this great tool!

wren:
@iedd Thanks for the report, and for the kind words!

Just to clarify, is this the situation you're talking about?

1. Encrypt a journal
2. Store password in keychain
3. Delete keychain
4. Make new blank keychain
5. Attempt to access the encrypted journal

Is that right?

iedd:
That is correct. In the situation you describe, the user will be prompted for the password in order to access the journal. In order to avoid the prompt and add it to the keychain, you must decrypt/re-encrypt.

PR #1602, opened 2022-10-01
Rework Encryption to enable future support of other encryption methods

wren:
This PR is intended to rework how `jrnl` does encryption in such a way that makes it easier for us to support multiple encryption methods. It should also make it easier to enable encryption support on other journal types (directory journal, DayOne, etc).

Fixes:
- #1521
- #1021

It does not fix, but lays the groundwork to fix the following issues in the near future:
- #648
- #1149 
- #1248
- #848

### Checklist

  for the same issue.


micahellison (dismissed):

micahellison:
This looks great. Let's plan on re-running tests then merging this after we drop Python 3.9 in #1611.

micahellison (approved):

PR #1603, opened 2022-10-02
Add documentation about information leaks in Vim/Neovim 

Granddave:
### Checklist

- [ ] I have included a link to the relevant issue number. *(found none)*
  for the same issue.
- [ ] I have written new tests for these changes, as needed. *(not needed)*

This PR aims to increase the knowledge about "leaky editors" and how to mitigate that risk. I have only experience in Vim-like editors, but this could be a base for future docs about other editors as well.

### DIFF
--- a/docs/external-editors.md
+++ b/docs/external-editors.md
@@ -14,6 +14,9 @@ in your [configuration file](./reference-config-file.md#editor)
 If your editor is not in your operating system's `PATH` environment variable,
 then you will have to enter in the full path of your editor.
 
+Please see [this section](./privacy-and-security.md#editor-history) about how
+your editor might leak sensitive information and how to mitigate that risk.
+
 ## Sublime Text
 
 To use [Sublime Text](https://www.sublimetext.com/), install the command line
--- a/docs/external-editors.md
+++ b/docs/external-editors.md
@@ -51,6 +51,17 @@ back to journal. In the case of MacVim, this is `-f`:
 editor: "mvim -f"
 ```
 
+## Vim/Neovim
+
+To use any of the Vim derivatives as editor in Linux, simply set the `editor`
+to the executable:
+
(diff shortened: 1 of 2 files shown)


wren:
@Granddave This is great information. Thank you!

I'm wondering if it's worth pointing out how to disable these settings without completely disabling user config. I've seen some people use an autocommand to set something like:
```vim
autocmd BufNewFile,BufReadPre *.jrnl setlocal noswapfile noundofile nobackup nowritebackup viminfo= shada=
```

Do you think this would be sufficient to mitigate the security concern while not degrading the editor experience as much as `-u NONE`?

Granddave:
Hi, sorry for the delay.

I'm no Vim expert but I think that starting Vim/Neovim with `-i NONE -n` is equivalent with setting `noswapfile viminfo=` for Vim and `noswapfile shada=` for Neovim so there seem to exist other leaky configs as you pointed out. Adding an autocommand might be easier to cover more cases.

But yes, disabling the whole config as well as having the autocommand disable some leaky options might be unnecessary, but IMO good to mention as an option for extra security.

I'll look into crafting a autocommand that we can use for the documentation for Vim and Neovim.

wren:
@Granddave No worries about the delay. Just update here, or open a new PR whenever you're ready. Thank you!

Granddave:
@wren, thanks!

I've pushed a fixup commit on this same branch with updated startup options as well as adding autocommands. Let me know what you think and I'll squash later if we decide to merge. :)

Granddave:
@wren Ping

wren:
> @wren Ping

I completely lost track of this since the PR is closed, sorry!

Would you mind opening a new one so we can go over it on Saturday?

Granddave:
No worries :)

PR #1618, opened 2022-10-20
Remove wrong option in configuration file reference

### LINKED ISSUE #1617: Documentation of "display_format" incorrectly states that it can be used to change date display format
### Diagnostic output

```shell
jrnl: v3.2
Python: 3.9.2 (default, Feb 28 2021, 17:03:44)
[GCC 10.2.1 20210110]
OS: Linux 5.4.72-microsoft-standard-WSL2
```

I want the time and date of my journal entries to be displayed in a format that I am used to (day-month-year). I do not want to change the timeformat of an existing journal file or the config.

### What have you tried?

The documentation of the [Configuration File](https://jrnl.sh/en/stable/reference-config-file/) says with regard to the timeformat:

> If you would just like to change how jrnl displays dates, use display_format instead.

This is what I did, leading to an attribute error when trying the show journal entries.

### Other Information

The documentation also mentions the display_format option in combination with the formatter, so I guess the point is wrong, that it changes the output format of the time and date only. If so: this would be a great new feature.

DSiekmeier:
There is a wrong statement in the documentation about the `display_format` format, which is removed in this commit.

### Checklist

#1617
  for the same issue.

### DIFF
--- a/docs/reference-config-file.md
+++ b/docs/reference-config-file.md
@@ -90,9 +90,6 @@ See the [python docs](http://docs.python.org/library/time.html#time.strftime) fo
 Do not change this for an existing journal, since that might lead
 to data loss.
 
-If you would just like to change how `jrnl` displays dates,
-use display_format instead.
-
 !!! note
     `jrnl` doesn't support the `%z` or `%Z` time zone identifiers.
 


micahellison (approved):
Thanks!

PR #1621, opened 2022-10-22
Fix bug for new `--list --format` options when no default journal is specified

### LINKED ISSUE #1445: Add machine-readable format for --list

wren:
Fixes #1445

While testing the beta, we found an issue where a command like `jrnl --list --format json` wouldn't work if no default journal is defined in the config (even though `jrnl --list` doesn't require a journal name). This fixes that issue, and also fixes seeing that same error when running `jrnl --list` normally.

### Checklist

  for the same issue.


micahellison (approved):
💯

PR #1624, opened 2022-10-29
Document that editors must be blocking processes

### LINKED ISSUE #1456: Add documentation about how the editor must be a blocking process
### Affected Page(s)

 - https://jrnl.sh/en/stable/external-editors/
 - https://jrnl.sh/en/stable/reference-config-file/

### What Could Be Better?

When I saw #1451 about getting neovide to work in jrnl, I realized that it wasn't clear in the docs that the editor needs to be a blocking process, rather than returning control to jrnl immediately.

- [External Editors page](https://jrnl.sh/en/stable/external-editors/)
  - [x] Before going into all the examples on how to make various editors into blocking processes, it should start by mentioning that this is required for any editor.
  - [x] May be worth mentioning that many editors are blocking by default, such as vim.
  - [x] The failure scenario for editing jrnl without a blocking editor is that jrnl just finishes running before you edit the entry. Mention that too, so people know what to look for.

- [Config reference page](https://jrnl.sh/en/stable/reference-config-file/)
  - [x] It states that "jrnl processes the file once the editor is closed" which is only true if the editor is a blocking process. The text should change a bit to reflect that.

micahellison:
Fixes #1456. Updates documentation to specify that editors must be blocking processes.

### Checklist

  for the same issue.

### DIFF
--- a/docs/external-editors.md
+++ b/docs/external-editors.md
@@ -14,7 +14,7 @@ then you will have to enter in the full path of your editor.
 !!! note
     To save and log any entry edits, save and close the file.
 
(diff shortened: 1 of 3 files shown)


wren (approved):
🗺️

PR #1631, opened 2022-10-30
Fix bug where changelog is always slightly out of date on release tags

wren:
As the title says, there's currently a bug in our release process where the release is tagged and published, and then the changelog is updated afterward. You can actually see it if you go through the tags in our repo and click on the changelog file for each one.

This PR makes it a part of the changelog step in CI to update the tag to the proper commit after the changelog is generated (the changelog generator itself requires a tag to work, which is why the tag has to be moved afterward instead of beforehand).

### Checklist

  for the same issue.

### DIFF
--- a/.github/workflows/changelog.yaml
+++ b/.github/workflows/changelog.yaml
@@ -157,7 +157,7 @@ jobs:
         run: |
           # This is a tag build (releases and prereleases)
           # update the tag to include the changelog
-          git tag -fam "${GITHUB_REF##*/}" "${GITHUB_REF##*/}"
+          git tag -fam "$GITHUB_REF_NAME" "$GITHUB_REF_NAME"
           git push --tags --force
 
       - name: Merge to Release branch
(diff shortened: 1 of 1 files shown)


.github/workflows/changelog.yaml
```diff
@@ -151,6 +151,9 @@ jobs:
           git add "$FILENAME"
           git commit -m "Update changelog [ci skip]"
           git push origin "$BRANCH"
+          # update the tag to include the changelog
+          git tag -fam "${GITHUB_REF##*/}" "${GITHUB_REF##*/}"
```
micahellison: A couple thoughts after doing some manual testing and looking at the [GitHub environment var doc](https://docs.github.com/en/actions/learn-github-actions/environment-variables):
* I'm not sure, but I think that this will tag every changelog change from a PR merge the PR # as well. Is that the case? If so, we might need to check `GITHUB_REF_TYPE` for `tag` first, like the "Check branch for new commits" step does
* Looks like we can use `GITHUB_REF_NAME` instead of `GITHUB_REF_##*/`

micahellison (changes requested):

.github/workflows/changelog.yaml
wren: Yup! You're right! This was supposed to be only during full releases (like the section below).

micahellison (approved):
🏷

Issue #1639, opened 2022-11-11
Adding an entry with a combination of flags causes a journal overwrite


richardjs:
### Diagnostic output

jrnl: v3.3
Python: 3.10.8 (main, Nov  1 2022, 14:18:21) [GCC 12.2.0]
OS: Linux 5.15.77-1-lts

[I confirmed this exists in the current `develop` branch.]

### Current Behavior

Run this, when no entries match the parameters:

`jrnl @dummytag -on today --edit --change-time`

And then save the entry. The journal file will only contain the just-added entry.

### Expected Behavior

To me, it seems like it should report no entries match the search. From the flags, the user is clearly expecting to edit an existing entry.

In fact, it appears to give a warning to that effect, but continues processing.

### Repro Steps

Described under "Current Behavior".

### Debug output

[Redacted slightly]

```
(jrnl-a6oBaqxX-py3.10) [username@computer jrnl]$ jrnl --debug @dummytag -on today --edit --change-time                                                                     
[14:08:06] DEBUG    Parsed args:                                                                                                                                     cli.py:41
                    Namespace(debug=True, preconfig_cmd=None, postconfig_cmd=None, filename=None, on_date='today', today_in_history=False, month=None, day=None,              
                    year=None, start_date=None, end_date=None, contains=None, strict=False, starred=False, limit=None, excluded=[], edit=True, delete=False,                  
                    change_time='now', export=False, tags=False, short=False, config_override=[], config_file_path='', text=['@dummytag'])                                    
           DEBUG    Reading configuration from file /home/username/.config/jrnl/jrnl.yaml                                                                        install.py:81
           DEBUG    Using configuration:                                                                                                                        install.py:107
                    "{                                                                                                                                                        
                        'colors': {                                                                                                                                           
                            'body': 'none',                                                                                                                                   
                            'date': 'none',                                                                                                                                   
                            'tags': 'none',                                                                                                                                   
                            'title': 'none'                                                                                                                                   
                        },                                                                                                                                                    
                        'default_hour': 17,                                                                                                                                   
                        'default_minute': 5,                                                                                                                                  
                        'editor': 'emacs -nw',                                                                                                                                
                        'encrypt': False,                                                                                                                                     
                        'highlight': True,                                                                                                                                    
                        'indent_character': '|',                                                                                                                              
                        'journals': {                                                                                                                                         
                            'default': '/home/username/journal.txt',                                                                                                
                                                                                                             
                        },                                                                                                                                                    
[...]

PR #1640, opened 2022-11-11
Fix for #1639 journal overwrite

richardjs:
This fixes the journal overwrite bug in #1639. 

There's a bit of awkward branching to account for existing use cases of calling `jrnl --edit` on empty journals; if we drop those, the cleaner fix in fce5afff0632a273de1713bc3a3f6f90c6b7706a could likely work.

### Checklist

  for the same issue.

### DIFF
--- a/jrnl/jrnl.py
+++ b/jrnl/jrnl.py
@@ -328,7 +328,11 @@ def _other_entries(journal: Journal, entries: list["Entry"]) -> list["Entry"]:
 
 
 def _edit_search_results(
-    config: dict, journal: Journal, old_entries: list["Entry"], **kwargs
+    args: "Namespace",
+    config: dict,
+    journal: Journal,
+    old_entries: list["Entry"],
+    **kwargs
 ) -> None:
     """
     1. Send the given journal entries to the user-configured editor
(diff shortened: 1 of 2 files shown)


micahellison (changes requested):
Instead of returning, jrnl should open the editor like any other empty search result.

For instance, if you run `jrnl --edit -contain StringThatIsNotInYourJournal` and enter an entry, it will save that new entry.

I'd expect that `jrnl --edit -contain StringThatIsNotInYourJournal --change-time` would behave exactly the same.

richardjs:
@micahellison, ok, I've modified the fix to behave in that way. I assumed we'd then want to set the newly created entry's time to the value passed to `--change-time` (or the default `now`). Unfortunately, as the change time operation runs before the edit operation ([see here](https://github.com/jrnl-org/jrnl/pull/1452#discussion_r857041184)), I needed to add a little bit of a special case to `_edit_search_results`.

micahellison:
Thanks @richardjs. @wren and I are chatting about this and realizing that this need for a special case reveals a bigger problem in the original flow that we'd like to fix. We definitely would like to keep your test, but we'll need some more time to think about the rest of it before considering merging this. We'll plan on getting back into this next week.

micahellison:
Hi again @richardjs, just wanted to let you know, we're still working on this, but it's a bit hairier than we thought. We're basically replacing the branching edit/change-time/delete actions with sequential actions, and I think it will be a lot simpler in the long run, but we're not quite there yet.

micahellison:
Closing this PR since it's resolved in our larger structural changes in #1669. Thanks for your work on getting this going, @richardjs.

Issue #1608, opened 2022-10-08
Adding type hints to the codebase


outa:
### Use Case/Motivation

Hi,

how do you feel about adding type hints to functions and methods in the code?

I think that in addition to catching and avoiding bugs, it would also make the code easier to understand for new contributors.

Plus one could add some kind of static type checker to the CI

### Example Usage

n/a

### Other Information

I'm willing to submit a PR

micahellison:
Thanks for filing this, and yes, we're into it! Just a couple things to be mindful of:

Right now we have an open draft PR #1602 that does this in a few places, so it might be worth glancing at that so you don't have to duplicate any work.

The other issue is that type hints have been changing rapidly over the past few Python releases, as we're seeing in that same PR (the tests in Python 3.9 are failing). We plan to merge it when we drop Python 3.9 support later this month, so don't worry about any failing 3.9 tests if you tackle this.

micahellison:
The bulk of the work is done (thanks @outa) though I'm leaving this open because we don't quite have enough static type checking yet. Once we have that in place and without errors, I think we can close this issue.

Looks like we'll probably be picking between mypy and pedantic, or maybe others. Suggestions are welcome.

PR #1614, opened 2022-10-12
Add type hints

outa:
Add type hints to functions and methods
Uses Python 3.10 syntax

Waiting for #1611 and #1602 to be merged.

cc: #1608 

### Checklist

- [X] I have read the [contributing doc](https://github.com/jrnl-org/jrnl/blob/develop/CONTRIBUTING.md).
- [X] I have included a link to the relevant issue number.
- [X] I have checked to ensure there aren't other open [pull requests](../pulls)
  for the same issue.
- [ ] I have written new tests for these changes, as needed.

PR #1642, opened 2022-11-25
Add more type hints

outa:
Add some more type hints now that #1602 is merged.

cc: #1608 

### Checklist

  for the same issue.

### DIFF
--- a/jrnl/Journal.py
+++ b/jrnl/Journal.py
@@ -59,7 +59,7 @@ class Journal:
         return (entry for entry in self.entries)
 
     @classmethod
-    def from_journal(cls, other):
+    def from_journal(cls, other: "Journal") -> "Journal":
         """Creates a new journal by copying configuration and entries from
         another journal object"""
         new_journal = cls(other.name, **other.config)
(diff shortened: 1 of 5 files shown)


jrnl/Journal.py
```diff
@@ -96,7 +96,7 @@ def _encrypt(self, text: str) -> bytes:
 
         return self.encryption_method.encrypt(text)
 
-    def open(self, filename=None):
+    def open(self, filename: str | None = None) -> "Journal":
```
wren, line 99: Can we use `Optional[str]` instead of `str | None`?

jrnl/encryption/__init__.py
```diff
@@ -17,7 +18,7 @@ def __str__(self) -> str:
     JRNLV2 = "Jrnlv2Encryption"
 
 
-def determine_encryption_method(config: str | bool) -> "BaseEncryption":
+def determine_encryption_method(config: str | bool) -> Type["BaseEncryption"]:
```
micahellison, line 22: Nice catch here!

wren (changes requested):
Thanks for this! I think the only change is to use `Optional` instead of `foo | None`.

jrnl/Journal.py
outa: They are equivalent right? So we could, but I have typed this consistently as `str | None` in #1614 
Anything wrong with the new syntax?

wren (approved):
lgtm

Thank you!

PR #1663, opened 2023-01-12
Search for entries with no tags or stars with `-not -starred` and `-not -tagged`

### LINKED ISSUE #1057: Search for entries with no tags or stars with `-not -starred` and `-not -tagged`
### Use Case/Motivation
I've moved to leveraging tags more seriously. 
I'd like be able to view/edit entries that are not tagged to (potentially) tag the entries
Another usage, untagged entries have an implicit "empty" tag, that could be part of a tagging system.
My current usage starts to emulate that, with untagged entries functioning as notes

### Example Usage
``jrnl -untagged``
### Other information
I've hacked together a shelled out (to awk) version, that "works"
``jrnl -s -not $(jrnl --tags | awk '/@/ {printf "%s ", $1}'``
It's ugly, as the resulting command line could be very long.
But this seems like it could/should be a core feature :-)

cjcon90:
This aims to fix jrnl-org/jrnl#1057 - [Filter for entries with no tags](https://github.com/jrnl-org/jrnl/issues/1057)

As discussed in that issue, the preferred syntax is `-not -starred` & `-not -tagged`, as this feels the most intuitive

I aimed to change the existing functionality as little as possible, but making this syntax fit did necessitate some minor changes:
- Add a `-tagged` flag, to show only tagged entries
- Alter `-not` to be allowed to take 0-1 arguments
- Manage parsing in case proceeded by `-starred` or `-tagged`
- If not is not followed by either tag, and is empty, then return the error we would normally get

---

**Note:** While I agree that `-not -starred` etc is the most intuitive syntax (it's what I instinctively tried initially), I'm not crazy about the double usage for the `-not` flag. Ideally each flag should only do one thing, but if we want to go a different route I can take a look at that also :-)

### Checklist

  for the same issue.


jrnl/args.py
```diff
@@ -28,6 +28,43 @@ def _split_lines(self, text: str, width: int) -> list[str]:
...
+    Pass -not without a following string and avoid appending
+    a None value to the exluded list
+    """
+
+    def __call__(self, parser, namespace, values, option_string=None):
+        if values is not None:
+            super().__call__(parser, namespace, values, option_string)
+
+
+def parse_not_tag(
```
micahellison: Could you rename this to `parse_not` or `parse_not_arg`?

jrnl/args.py
```diff
@@ -28,6 +28,43 @@ def _split_lines(self, text: str, width: int) -> list[str]:
         return text
 
 
+class IgnoreNoneAppendAction(argparse._AppendAction):
+    """
+    Pass -not without a following string and avoid appending
+    a None value to the exluded list
```
micahellison: Just a little typo here - "exluded" instead of "excluded"

micahellison (changes requested):
Welcome to jrnl, and thank you for the very complete and thoughtful PR! I like the overall approach you took, and I appreciate the new tests as well.

I just have a couple minor nitpicks around spelling/wording, but I don't think this needs any major changes.

cjcon90:
@micahellison thank you! Updated both.

Happy to be able to contribute - just started using it daily and really love the app :-)

wren:
@cjcon90 Hi! We didn't expect you to respond so quickly, and just now merged another PR that will likely cause another merge conflict. Sorry about that!

As soon as that's fixed, we'll get yours merged in. Thanks for the great work!

cjcon90:
Sorry meant to do this earlier and completely went out of my head! Rebased my changes so hopefully ready to merge :-)

micahellison (approved):

micahellison:
Thanks for the PR and changes! Looks good to me. This still needs to be documented, but I've filled that in a separate issue #1668.

PR #1667, opened 2023-01-23
Add ability to use template with `--template`

alichtman:
You can use this CLI argument in two ways:

```bash
# Pass a filepath directly (relative or absolute)
$ jrnl --template local_file.md
$ jrnl --template /tmp/abs_path_file.md

# Or use the default XDG template directory
$ mkdir -p $XDG_DATA_HOME/jrnl/templates
$ echo "Hi! I'm a template!" > $XDG_DATA_HOME/jrnl/templates/test.md
$ jrnl --template test.md
```

I've updated the docs for the Templates feature.

Also, see https://github.com/alichtman/jrnl-templates

### Checklist

  for the same issue.


jrnl/args.py
```diff
@@ -174,6 +174,7 @@ def parse_args(args: list[str] = []) -> argparse.Namespace:
         "Writing", textwrap.dedent(compose_msg).strip()
     )
     composing.add_argument("text", metavar="", nargs="*")
+    composing.add_argument("--template", dest="template", help="Path to template file to be used, relative to $JRNL_TEMPLATE_DIR")
```
alichtman:
```suggestion
    composing.add_argument("--template", dest="template", help="Path to template file. Can be a local path, absolute path, or a path relative to $JRNL_TEMPLATE_DIR")
```

tests/unit/test_parse_args.py
```diff
@@ -39,6 +39,7 @@ def expected_args(**kwargs):
         "start_date": None,
         "strict": False,
         "tags": False,
+        "template": None,
```
alichtman, line 45: It's odd that the tests fail when this value is set to `False`. I didn't expect that.

jrnl/controller.py
```diff
@@ -218,45 +305,16 @@ def search_mode(args: "Namespace", journal: Journal, **kwargs) -> None:
         _display_search_results(**kwargs)
 
 
-def _write_in_editor(config: dict, template: str | None = None) -> str:
+def _write_in_editor(config: dict, prepopulated_text: str | None = None) -> str:
```
micahellison, line 302: Nice change with `prepopulated_text` change instead of `template`. It's good to root out metonymies like this from the code.

micahellison (changes requested):
Thanks for this PR. I like the `--template` idea and I think it's a lot more fluent and easy to use than the current method of `--config-overide template file`. I also really appreciate the changes to the docs here too.

These are a few different changes, though, and I feel particularly hesitant about the environment variable change. Could you remove that part from this PR? I think we need a broader conversation about how jrnl could handle environment variables, and I think that will take some time, but I don't want it to hold up this nice `--template` change.

I think we could use some BDD tests for `--template` as well. There's a new `template.feature` file in there that could use these tests. Let me know if you'd like any help with that.

jrnl/controller.py
alichtman: > I feel particularly hesitant about the environment variable change. Could you remove that part from this PR? I think we need a broader conversation about how jrnl could handle environment variables, and I think that will take some time, but I don't want it to hold up this nice `--template` change.

Sure. The purpose of the env var was to provide some shorthand for creating a new jrnl from a template (instead of having to `cd` to the template directory and then run `$ jrnl --template LOCAL_TEMPLATE`). Since `jrnl` already supports `XDG`, does it make sense to create `$XDG_CONFIG_HOME/jrnl/templates` if it does not exist, and use that as the shorthand templates directory instead?

> I think we could use some BDD tests for `--template` as well. There's a new `template.feature` file in there that could use these tests. Let me know if you'd like any help with that.

I agree that tests are worthwhile here. I wanted to get your eyes on the PR before I invested time in writing tests.

micahellison: > Since `jrnl` already supports `XDG`, does it make sense to create `$XDG_CONFIG_HOME/jrnl/templates` if it does not exist, and use that as the shorthand templates directory instead?

Sorry for the delay in response here. I think your suggestion is a great compromise and I'm all for it. Though I think we should use `$XDG_DATA_HOME/jrnl/templates` instead since that fits the XDG spec and we already use `$XDG_DATA_HOME/jrnl` as the default path for the default journal.

alichtman:
I've made the default template path `$XDG_DATA_HOME/jrnl/templates` and updated the docs. Everything works based on my manual testing, and I've added some `bdd` tests. 

I'm unsure how to handle two test cases:

1. Testing that involves `$XDG_DATA_HOME/jrnl/templates` without _actually_ messing with the system's `$XDG_DATA_HOME/jrnl/templates` folder. 
2. Testing that involves an arbitrary absolute filepath.

Do you have any ideas about how I can mock that behavior out?

And it's unclear to me why `test_template_local_filepath_should_be_used_in_new_entry` is failing. Will debug more later.

tests/bdd/features/template.feature
```diff
@@ -31,4 +31,64 @@ Feature: Using templates
...
+        Then the error output should contain "Unable to find a template file based on the passed arg"
+
+        Examples: configs
+        | config_file          |
+        | basic_onefile.yaml   |
+        | basic_encrypted.yaml |
+        | basic_folder.yaml    |
+        | basic_dayone.yaml    |
+
+    Scenario Outline: --template local_filepath should be used in new entry
```
alichtman, line 50: This test fails. Not sure why.

micahellison:
This looks nice! Thanks for your work on this. I'll plan on taking a deeper look at this later but wanted to get to your questions for now.

> Testing that involves `$XDG_DATA_HOME/jrnl/templates` without _actually_ messing with the system's `$XDG_DATA_HOME/jrnl/templates` folder.

My instinct would be to create a new fixture in `fixtures.py` patching your new `get_templates_path` method to return a subdirectory of the temp_dir fixture, like the `mock_default_journal_path` fixture, then add it to the cli_run fixture so the patch is applied in the "we run" step. After that, a new given step could populate that directory with a template from the test data folder. Like, "Given we copy the {template_file_name} template to the $XDG_DATA_HOME/jrnl/templates folder".

>  Testing that involves an arbitrary absolute filepath.

I'm not sure. I'm looking through other tests we've done with path inputs, such as `--config-file` and the installation wizard, and I don't see anything using absolute paths. One might argue that absolute vs. relative paths are a concern of the Python libraries rather than jrnl itself and therefore don't need to be tested by us, though it still could be nice if we had an approach for this since path manipulation bugs are common, especially in cross-platform apps. Maybe @wren has some thoughts on this. Either way, I'd be fine accepting this PR without absolute file path tests.

> And it's unclear to me why `test_template_local_filepath_should_be_used_in_new_entry` is failing.

Due to #1653, jrnl now ignores entries that are exactly the same as the template, so the entry isn't being added. You can copy a couple lines from an earlier test in template.feature to add some text to the entry and get it working:
```
        And we append to the editor if opened
            This is an addition to a templated entry
```

alichtman:
> Due to https://github.com/jrnl-org/jrnl/pull/1653, jrnl now ignores entries that are exactly the same as the template, so the entry isn't being added. You can copy a couple lines from an earlier test in template.feature to add some text to the entry and get it working:

Fixed. 

> My instinct would be to create a new fixture in fixtures.py patching your new get_templates_path method to return a subdirectory of the temp_dir fixture, like the mock_default_journal_path fixture, then add it to the cli_run fixture so the patch is applied in the "we run" step. After that, a new given step could populate that directory with a template from the test data folder. Like, "Given we copy the {template_file_name} template to the $XDG_DATA_HOME/jrnl/templates folder".

Makes sense to me. What I've got in the last commit seems like it _should_ work, but the `get_templates_path` function isn't being properly mocked (and is instead returning my real `XDG`-based templates path). Any ideas on what I missed?

> One might argue that absolute vs. relative paths are a concern of the Python libraries rather than jrnl itself and therefore don't need to be tested by us

I think Pathlib does all the work here, and it's acceptable to drop this test entirely. Plus, the risk is low: worst case, a `jrnl` entry won't be prepopulated with a template.

tests/lib/fixtures.py
```diff
@@ -179,6 +181,16 @@ def mock_default_journal_path(temp_dir):
     }
 
 
+@fixture
+def mock_get_templates_path(temp_dir):
```
alichtman: This mock does not work correctly. `jrnl.config.get_templates_path` is the object I want to patch, and I've added this fixture to the `mock_factories` object.

micahellison: Since the method is being called in `controller.py`, you can get it to work by patching `jrnl.controller.get_templates_path` instead of `jrnl.config.get_templates_path`

alichtman: You're right. I misunderstood how this worked. Thanks for the help!

jrnl/controller.py
```diff
@@ -123,9 +124,78 @@ def _is_write_mode(args: "Namespace", config: dict, **kwargs) -> bool:
     return write_mode
 
 
+def _read_template_file(template_arg: str, template_path_from_config: str) -> str:
```
micahellison, line 127: I'm thinking that this function should be split up into shorter functions and at least some of them should be moved elsewhere. `editor.py` has previously been the only file involved with templates and I think some of these concerns belong there. Ideally, the controller is just telling other classes what to do, rather than doing the work itself.

Though let me know if you're not interested in that work -- we're happy to do it as well.

micahellison (changes requested):
Just one note about a long function. Otherwise, it looks good. Thanks for your work on this!

jrnl/controller.py
alichtman: > I'm thinking that this function should be split up into shorter functions and at least some of them should be moved elsewhere.

Seems reasonable enough to me. I don't have a strong opinion on it.

> Though let me know if you're not interested in that work -- we're happy to do it as well.

I'll pass on implementing it myself :) I'm already running a serious open-source time deficit.

alichtman (commented):
I'd be alright shipping this as is, and cutting a follow-up task for the changes you suggested, @micahellison.

Thanks for the reviews!

micahellison (approved):
Thanks for all your work on this!

PR #1680, opened 2023-02-13
Update documentation about privacy and security in VSCode

### LINKED ISSUE #1675: Visual Studio Code may store unencrypted temporary files
### Affected Page(s)

https://jrnl.sh/en/stable/external-editors/#visual-studio-code

### What Could Be Better?

Visual Studio Code supports local file history. A local history entry is a copy of the file at the time it was saved with some associated metadata. Local history entries persist in the `<user-data-dir>/User/History` folder.
Therefore, if Visual Studio Code is configured as external editor, unencrypted temporary files remain stored on disk.
It is possible to exclude files from local file history by using the `workbench.localHistory.exclude` setting with the following exclude pattern: `**/jrnl*.jrnl`.
Do you think it is useful to add this information in the documentation?

### Other Information

_No response_

giuseppedandrea:
Close #1675 

This PR updates the documentation by warning users about two VS Code features (local file history and backup of unsaved files) that may leave unencrypted temporary files stored on disk.

### Checklist

  for the same issue.


docs/privacy-and-security.md
```diff
@@ -73,6 +73,37 @@ Some editors keep usage history stored on disk for future use. This can be a
...
+setting in the [Settings editor](https://code.visualstudio.com/docs/getstarted/settings#_settings-editor).
+
+Alternatively, you can configure a [pattern](https://code.visualstudio.com/docs/editor/codebasics#_advanced-search-options) in the `workbench.localHistory.exclude`
+setting for excluding only specific files from local file history. For excluding
+unencrypted temporary file generated by `jrnl`, you can set the `**/jrnl*.jrnl`
+pattern for the `workbench.localHistory.exclude` setting in the [Settings editor](https://code.visualstudio.com/docs/getstarted/settings#_settings-editor).
+
+!!! note
+    On Windows, the history location is typically found at
+    `%APPDATA%\Code\User\History`.
```
micahellison, line 91: Do you have a link that describes this in the VS Code documentation? I searched for it but couldn't find any information outside of Stack Overflow.

micahellison (changes requested):
Thanks for the thoughtful PR! I made some small changes for tone, and had just one question about backing something up with a link to VS Code's docs.

docs/privacy-and-security.md
giuseppedandrea: Actually, I found this position by manually searching in the local data folder of VS Code. There is no reference in the documentation as to where the location actually is.
Now I have just found a note about this in the [release notes of version 1.66](https://code.visualstudio.com/updates/v1_66#_local-history) (the one in which local history was introduced):

> Local history entries are stored in different locations depending on your use of VS Code. When opening local files, the entries are persisted in the local user data folder and when opening remote files, they will be stored on the remote user data folder. When no file system is available (for example, in certain cases when using VS Code for Web), entries are stored into IndexedDB.

micahellison: Thanks for the info, and for testing this. That works for me. That's interesting about the remote data and no-file-system scenarios, but I think it's fine not to mention since the jrnl temp files are local.

micahellison (approved):

PR #1684, opened 2023-02-20
Document `-tagged`, `-not -tagged`, and `-not -starred`

### LINKED ISSUE #1668: Document `-tagged`, `-not -tagged`, and `-not -starred`
### Affected Page(s)

* [Command Line Reference](https://jrnl.sh/en/stable/reference-command-line/)
* Possibly one of the user guide pages as well.

### What Could Be Better?

#1057 is about to be resolved, but the resolution should be documented.

### Other Information

See notes in #1663 about -not used for both -tagged and -starred

micahellison:
Fixes #1668. Documents the behavior modified in #1663.

In the long run, I think we'll want a more polymorphic approach to documenting `-not`, but since it only supports three different inputs and two of them are already search arguments, I thought it best to keep it all separate for now. This also helps with the table layout on the site, since it can't be widened nicely in our current docs style.

### Checklist

  for the same issue.

### DIFF
--- a/docs/reference-command-line.md
+++ b/docs/reference-command-line.md
@@ -76,8 +76,11 @@ entries, such as `yesterday`, `today`, `Tuesday`, or `2021-08-01`.
 | -contains TEXT | Show entries containing specific text (put quotes around text with spaces) |
 | -and | Show only entries that match all conditions, like saying "x AND y" (default: OR) |
 | -starred | Show only starred entries (marked with *) |
+| -tagged | Show only tagged entries (marked with the [configured tagsymbols](reference-config-file.md#tagsymbols)) |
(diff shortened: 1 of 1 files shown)


wren (changes requested):
Great clarifications!

Can you add this to the help screen at the same time, please?

micahellison:
> Can you add this to the help screen at the same time, please?

Turns out this was already done in the original PR, and looks pretty good to me:
```
  -not [TAG/FLAG]       If passed a string, will exclude entries with that tag.
                        Can be also used before -starred or -tagged flags, to
                        exclude starred or tagged entries respectively.
```

wren (approved):
:label:

PR #1687, opened 2023-02-22
Prompt to include colors in config when first running jrnl

### LINKED ISSUE #1372: Prompt to include colors in config when first running jrnl
### Use Case/Motivation

The colors are a relatively new feature that can be modified in the config file, but we never enabled them in the default configuration. It seems like a good question to ask new users about, with a default to "yes".

### Example Usage

When you first run jrnl and there is no config file found, you encounter a wizard with a couple prompts like this:
```
Path to your journal file (leave blank for /home/micahadmin/.local/share/jrnl/journal.txt):
Do you want to encrypt your journal? You can always change this later [y/N]
```
I imagine this feature adding another line, like:
```
Do you want the jrnl interface to use colors? You can always change this later [Y/n]
```
(that's defaulting to "Y" here)

### Other information

I think this is a good default color config (thanks @wren for these), though I'm open to discussion on this:
```
colors:
  body: none
  date: black
  tags: yellow
  title: cyan
```

Thanks to @dufferzafar for raising this question in #1352.

micahellison:
Fixes #1372.

Default colors are set in the config as:
```
colors:
  body: none
  date: magenta
  tags: yellow
  title: cyan
```

image

However, I'm open to changing this.

### Checklist

  for the same issue.


wren (dismissed):
:crayon:

wren (approved):

PR #1691, opened 2023-02-24
Support mkdocs 1.4.2 and fix its missing breadcrumb

micahellison:
This PR upgrades us to mkdocs 1.4.2 and fixes the "missing" breadcrumb.

It turns out that the problem was a combination of issues:

* a styling issue in jrnl's CSS with a class that's used multiple times for very different types of objects in the mkdocs ReadTheDocs template (sidebar logo and breadcrumb logo) - it turns out that "missing" breadcrumb was actually the white jrnl logo on a white background, using a CSS :before selector
* a change in the mkdocs ReadTheDocs template's top-level breadcrumb link that was broke support for screen readers by removing the tag's content; RTD later fixed this with an `aria-label` but that hasn't made it into the mkdocs source

I'll see about getting that change upstream so we don't have to override their breadcrumbs template and others can benefit from this work, but in the meantime, this PR fixes it in jrnl so we can use the latest mkdocs version.

### Checklist

  for the same issue.


micahellison:
Once [this is fixed upstream](https://github.com/mkdocs/mkdocs/issues/2827), the breadcrumbs.html file can be deleted from docs_theme directory.

wren (dismissed):
Nice! Thank you! :boom:

wren (approved):
:1st_place_medal:

PR #1693, opened 2023-02-27
Use rich instead of ansiwrap to wrap text

### LINKED ISSUE #1191: Drop/replace ansiwrap dependency
### Environment
  - Local env (running pytest)

### Current Behavior
When running pytest, I see a deprecation notice for ansiwrap (`import imp`). When looking into ansiwrap, I find that they already have an issue filed (jonathaneunice/ansiwrap#17), but no response or updates from the project in more than 6 months. The project itself also hasn't been updated in ages, so we should probably look for a replacement.

### Expected Behavior

We don't get deprecation warnings when running our test suite.

### Repro Steps

1. Run `pytest`
2. See warning

```
================================== warnings summary ===================================
../../../Library/Caches/pypoetry/virtualenvs/jrnl-GFqVlfgP-py3.9/lib/python3.9/site-packages/ansiwrap/core.py:6
  /Users/jonathan/Library/Caches/pypoetry/virtualenvs/jrnl-GFqVlfgP-py3.9/lib/python3.9/site-packages/ansiwrap/core.py:6: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
    import imp

-- Docs: https://docs.pytest.org/en/stable/warnings.html
============================ 77 passed, 1 warning in 0.84s ============================
```

micahellison:
Fixes #1191. This PR removes ansiwrap and uses rich instead. It only affects "pretty printing" -- the default display format that jrnl uses. 

⚠ This PR modifies the line wrapping behavior slightly: the old method removed trailing spaces, whereas this new method includes them. We don't consider this a breaking change because this format is meant to be read by humans instead of machines.

### Checklist

  for the same issue.


micahellison:
I am not sure why this last test is failing:
```
tests/bdd/test_features.py::test_printing_a_journal_that_has_multiline_entries_with_tags
```
image

Adding a space to the end of the second line makes it pass. However, when I run the test with manual jrnl commands, I don't see that extra space.

The original test result may seem a bit odd, too: the last word is on its own line. Both the original ansiwrap and the new rich wrappers are giving it its own line because the source data is actually already wrapped, and this is the only line in the source data that is at least as long as the linewrap value (80 characters). In fact, it's exactly the length of the linewrap value. Still, I have no explanation as to why this space isn't being preserved in my manual tests. It would be nice to know what's going on instead of just changing the test to make it pass.

micahellison:
We found that the issue is that rich doesn't remove trailing spaces when it wraps, whereas ansiwrap did. We're staying with rich's default behavior and I've made a note about this in the PR description.

wren (approved):
🫡

PR #1763, opened 2023-07-03
Replace flake8 and isort with ruff linter and add `black --check` to linting step

### LINKED ISSUE #1742: Linting rules aren't enforced the same as format rules
### Diagnostic output

n/a - this is a dev issue

### Current Behavior

Our linter isn't consistently checking for the same rules that black is running.

### Expected Behavior

We expect to catch formatting issues in a `poe test` run or through C/I, rather than always having to remember to run `poe format`.

### Repro Steps

One rule that isn't consistent is the line length rule. Here's how to repro:

Add a line to a .py file in the jrnl folder like so:
```
print("Testing a very long line") # here is a long enough line with a comment at the end, so black will indent it
```
Run `poe lint`

It will pass, but it shouldn't.

Run `poe format` and it will change into a state that should pass:
```
print(
    "Testing a very long line"
)  # here is a long enough line with a comment at the end, so black will indent it
```

### Debug output

n/a

### Other Information

Might be worth switching to a different linter. Maybe [ruff](https://github.com/charliermarsh/ruff), which runs much faster than flakeheaven.

micahellison:
This PR:

* replaces flake8/flakeheaven with ruff
* replaces isort with ruff
* adds `black --check` to the linting step. Fixes #1742.
* fixes line lengths of comments and long strings to comply with ruff's rules

While flakeheaven has worked out well for us, it is an order of magnitude slower than ruff on my machine. Moreover, ruff's configuration feels a lot simpler -- we don't have to install a bunch of packages for various rule sets.

FlakeHeaven was ignoring long lines with comments or strings, whereas ruff is a bit more strict. I reworded comments as necessary and broke long strings into multiple lines. In a couple instances, I added the `noqa` directive to ignore this rule on particular lines since I think they're already most readable in their current forms.

This should only affect developers, and developers shouldn't have to change anything in their process. Just keep running `poe lint` for linting checks and `poe format` to automatically sort imports and allow black to reformat code.

### Checklist

  for the same issue.


wren (approved):
🐩

PR #1768, opened 2023-07-10
Add note about messages going to `stderr` and the implication for piping

micahellison:
This PR adds a note in the "Data Formats" section of the docs about which data goes to `stdout` vs `stderr`.

While this is standard behavior for console apps, we have gotten a few issues about the record count messages appearing to interfere with piping because people weren't aware of this, so I'm hoping that this will help.

### Checklist

  for the same issue.


wren (approved):
🗒️

PR #1697, opened 2023-03-02
Only read text files that look like entries when opening folder journal

micahellison:
Fixes #1692.

### Checklist

- [x] I have read the [contributing doc](https://github.com/jrnl-org/jrnl/blob/develop/CONTRIBUTING.md).
- [x] I have included a link to the relevant issue number.
- [x] I have checked to ensure there aren't other open [pull requests](../pulls)
  for the same issue.
- [x] I have written new tests for these changes, as needed.

I didn't add a new test for this specifically, but rather, I modified the test folder journal to include entries that should be ignored. Without the change in `FolderJournal.py`, the tests blow up quite fantastically.

PR #1789, opened 2023-09-07
Allow custom file extensions for directory-based journal

### LINKED ISSUE #1289: Allow custom file extensions for directory-based journal
### Environment
```
jrnl: v2.7
Python: 3.9.5 (default, May 24 2021, 12:50:35) 
[GCC 11.1.0]
OS: Linux 5.10.42-1-MANJARO
pipx
```

I have my journal saving in a directory instead of a monolithic file. I'd like the new entries to be saved with a .md extension. What is the proper way to set this in the config file?

### What have you tried?
Setting a yaml option as follows:

```
format: markdown
```

```
format: md
```

thw26:
This PR seeks to resolve #1289, which I opened a while back. This PR utilizes a new config option `extension` which permits the user to change the file extensionNow, when a directory-based journal is searched or a new entry is created, jrnl will recognize files with the custom file extension or create a file using the custom file extension, respectively.

This PR makes use of regex for the filename recognition.

A previous version of this code attempted to implement the possibility to use an array of extensions, but eventually I bumped into the rigid structure alluded to in https://github.com/jrnl-org/jrnl/issues/1289#issuecomment-876170381. I reviewed the code in order to understand what was happening. Now that I understand what the code is doing, I see that implementing an array would require a much larger amount of work. A single file extension change should be much more manageable at this time.[^0]

I have tested basic functionality and the code is working, but this PR is flagged as a draft because I have to learn how to handle output from `poe test` and then review that output. I am opening this PR as draft in order to share my work and to receive help.

```
=========== 111 failed, 528 passed, 40 skipped in 72.57s (0:01:12) ===========
```

When setting "extension: txt", I get:

```
=========== 111 failed, 528 passed, 40 skipped in 74.00s (0:01:14) ===========
```

Several of these appear to be related to me having to pass along additional positional arguments. Hopefully I can resolve these with time.

### Checklist

  for the same issue.

---

[^0]: For posterity's sake, when implementing an array of extensions, I found that the code worked—except and unless there are multiple files on the same day with different extensions. The code is looking for a single file extension in part because once the entries are read from the files, the file paths are no longer passed along to the rest of the code, rather the file paths are retrieved and then reassigned. And so when attempting to use such functionality, the `--delete` option more or less no longer functions properly—not on .txt files or the first extension set in the config file—but on additional extensions beyond these. Without being able to specify the file, the delete option ceases working as intended. Perhaps to summarize, jrnl's internal code operates off of journal entries, not off of files.

### DIFF
--- a/jrnl/journals/FolderJournal.py
+++ b/jrnl/journals/FolderJournal.py
@@ -4,6 +4,7 @@
 import codecs
 import os
 import pathlib
+import re
 from typing import TYPE_CHECKING
 
 from jrnl import time
@@ -17,7 +18,7 @@ if TYPE_CHECKING:
 DIGIT_PATTERN = "[0123456789]"
 YEAR_PATTERN = DIGIT_PATTERN * 4
 MONTH_PATTERN = "[01]" + DIGIT_PATTERN
-DAY_PATTERN = "[0123]" + DIGIT_PATTERN + ".txt"
+DAY_PATTERN = "[0-3][0-9].(txt|md)"
 
 
 class Folder(Journal):
@@ -144,9 +145,11 @@ class Folder(Journal):
 
     @staticmethod
     def _get_day_files(path: pathlib.Path) -> list[str]:
-        for child in path.glob(DAY_PATTERN):
+        for child in path.iterdir():
+            match = re.search(DAY_PATTERN, str(child))
             if (
-                int(child.stem) > 0
+                match is not None
+                and int(child.stem) > 0
                 and int(child.stem) <= 31
                 and time.is_valid_date(
                     year=int(path.parent.name),
--- a/jrnl/journals/FolderJournal.py
+++ b/jrnl/journals/FolderJournal.py
@@ -18,7 +18,6 @@ if TYPE_CHECKING:
 DIGIT_PATTERN = "[0123456789]"
 YEAR_PATTERN = DIGIT_PATTERN * 4
 MONTH_PATTERN = "[01]" + DIGIT_PATTERN
-DAY_PATTERN = "[0-3][0-9].(txt|md)"
 
 
 class Folder(Journal):
@@ -35,7 +34,7 @@ class Folder(Journal):
         self.entries = []
 
         if os.path.exists(self.config["journal"]):
-            filenames = Folder._get_files(self.config["journal"])
+            filenames = Folder._get_files(self, self.config["journal"])
             for filename in filenames:
                 with codecs.open(filename, "r", "utf-8") as f:
                     journal = f.read()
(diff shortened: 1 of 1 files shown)


jrnl/journals/FolderJournal.py
```diff
@@ -45,6 +45,8 @@ def open(self) -> "Folder":
 
     def write(self) -> None:
         """Writes only the entries that have been modified into proper files."""
+        if self.config["extension"] is not None:
```
micahellison, line 48: This line is one of the reasons the tests are failing -- the "extension" key doesn't always exist in every config, so there's a KeyError when trying to retrieve it. Checking `self.config.get("extension")` instead of `self.config["extension"]` should do the trick.

jrnl/journals/FolderJournal.py
```diff
@@ -45,6 +45,8 @@ def open(self) -> "Folder":
...
+            EXTENSION = self.config["extension"]
```
micahellison, line 49: A little style note - could you keep this extension variable lower case? Upper case is meant for constants.

jrnl/journals/FolderJournal.py
```diff
@@ -121,12 +123,14 @@ def parse_editable_str(self, edited: str) -> None:
         self.entries = mod_entries
 
     @staticmethod
-    def _get_files(journal_path: str) -> list[str]:
+    def _get_files(self, journal_path: str) -> list[str]:
```
micahellison, line 126: Since this is a static method, it shouldn't take self. Instead, I think it would be better to pass the extension as a argument with a default value of None, so the `_get_files` method doesn't have to be "aware" of how the config works.

jrnl/journals/FolderJournal.py
```diff
@@ -143,10 +147,14 @@ def _get_month_folders(path: pathlib.Path) -> list[pathlib.Path]:
         return
 
     @staticmethod
-    def _get_day_files(path: pathlib.Path) -> list[str]:
-        for child in path.glob(DAY_PATTERN):
+    def _get_day_files(self, extension, path: pathlib.Path) -> list[str]:
+        EXTENSION = extension
+        DAY_PATTERN = "[0-3][0-9]." + EXTENSION
+        for child in path.iterdir():
```
micahellison, line 153: I would prefer to keep the glob method rather than using iterdir and regex, for performance reasons. We've seen issues in the past where unpredictable things can happen for users when pointing jrnl to a large directory, and iterdir has been the culprit since it iterates through all files instead of just the expected files.

We can still use the glob approach here by adding the configured extension the day pattern at the start of this method, though.

micahellison (changes requested):
Thanks for the PR. I added some initial comments below that I hope should help. Let me know if you have any questions about any of it.

We'll also need new tests to accept this PR. I recommend taking a look at `/jrnl/tests/bdd/features/file_storage.feature` and testing for a new extension. It will probably need a new sample config and sample folder journal data as well. I did the same thing in a PR #1697 a few months ago, which I think should help show what's needed, but again, if you have any questions, feel free to respond here and I'll be happy to help.

micahellison:
Closing this PR since #1873 takes it further.

PR #1812, opened 2023-10-18
Pin ReadTheDocs Python version to 3.11

micahellison:
ReadTheDocs is failing to build again, this time during a `pip install` step. My best guess is that this is due to some dependency it has that isn't working in Python 3.12, so I'm pinning the version to 3.11.

### Checklist

  for the same issue.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -9,7 +9,7 @@ version: 2
 build:
   os: ubuntu-22.04
   tools:
-    python: "3"
+    python: "3.11"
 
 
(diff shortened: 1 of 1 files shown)


micahellison:
That did it. I don't think this is a problem, either. RTD has been using Python 3.7 to build the docs this whole time anyway, so it's not like we're losing out on functionality we would otherwise want by using 3.11.

micahellison:
Looks like ReadTheDocs handled this on their end. I'm sure lots of other projects had the same problem. Closing this with no need to merge.

PR #1873, opened 2024-03-23
Allow custom file extensions for directory journal

comkieffer:
This PR builds on https://github.com/jrnl-org/jrnl/pull/1789 (Closes issue #1289). 

In short, it adds a new configuration option `extension` that controls the extension for the files created with a folder journal. 

### Checklist

- [X] I have read the [contributing doc](https://github.com/jrnl-org/jrnl/blob/develop/docs/contributing.md).
  for the same issue.

### DIFF
--- a/jrnl/journals/FolderJournal.py
+++ b/jrnl/journals/FolderJournal.py
@@ -17,6 +17,7 @@ if TYPE_CHECKING:
 DIGIT_PATTERN = "[0123456789]"
 YEAR_PATTERN = DIGIT_PATTERN * 4
 MONTH_PATTERN = "[01]" + DIGIT_PATTERN
+DAY_PATTERN = "[0-3][0-9]"
 
 DEFAULT_EXTENSION = ".txt"
 
@@ -148,15 +149,15 @@ class Folder(Journal):
 
     @staticmethod
     def _get_day_files(path: pathlib.Path, extension: str) -> list[str]:
-        DAY_PATTERN = "[0-3][0-9]" + extension
-        for child in path.glob(DAY_PATTERN):
+        for child in path.glob(DAY_PATTERN + extension):
+            day = str(child.name).replace(extension, "")
             if (
-                int(child.stem) > 0
-                and int(child.stem) <= 31
+                int(day) > 0
+                and int(day) <= 31
                 and time.is_valid_date(
                     year=int(path.parent.name),
                     month=int(path.name),
-                    day=int(child.stem),
+                    day=int(day),
                 )
                 and child.is_file()
             ):
--- a/jrnl/journals/FolderJournal.py
+++ b/jrnl/journals/FolderJournal.py
@@ -17,7 +17,8 @@ if TYPE_CHECKING:
 DIGIT_PATTERN = "[0123456789]"
 YEAR_PATTERN = DIGIT_PATTERN * 4
 MONTH_PATTERN = "[01]" + DIGIT_PATTERN
-DAY_PATTERN = "[0123]" + DIGIT_PATTERN + ".txt"
+
+DEFAULT_EXTENSION = ".txt"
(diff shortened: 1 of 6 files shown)


thw26:
Thanks for building on my PR! I have not had the time to return to it since the review.

tests/bdd/features/file_storage.feature
```diff
@@ -10,13 +10,21 @@ Feature: Journals iteracting with the file system in a way that users can see
...
+            2013/07/23.md
+
     Scenario: Adding multiple entries to a Folder journal should generate multiple date files
         Given we use the config "empty_folder.yaml"
         When we run "jrnl 23 July 2013: Testing folder journal."
         And we run "jrnl 3/7/2014: Second entry of journal."
         Then we should get no error
         And the journal directory should contain
             2013/07/23.txt
+            2014/07/23.txt
```
micahellison: Date doesn't line up with the test case in default US date layout -- either this should be changed to 2014/03/07.txt or the second `we run` line should be changed to: `And we run "jrnl 7/23/2014: Second entry of journal."`

tests/bdd/features/file_storage.feature
```diff
@@ -10,13 +10,21 @@ Feature: Journals iteracting with the file system in a way that users can see
         And the journal directory should contain
             2013/07/23.txt
 
+    Scenario: Adding entries to a Folder journal with a custom extension should generate date files
+        Given we use the config "empty_folder_with_extension.yaml"
+        When we run "jrnl 23 July 2013: Testing folder journal."
+        Then we should get no error
```
micahellison, line 17: Congratulations, your test uncovered an existing bug! 🎉 I filed it in #1894.

Looks like `empty_folder.yaml` got past this issue by having an "empty" file in it: `tests/data/journals/empty_folder/empty`

Feel free to copy that "empty" file to `tests/data/journals/empty_folder_with_extension/empty` to get this working.

micahellison (changes requested):
Sorry for taking so long to get to this PR, but also, thank you for the excellent PR. I really appreciate that you're modifying documentation, adding tests, and adding thoughtful changes within the norms of the existing code.

The GitHub Actions tests were broken when your PR was first submitted, but they should be workable now if you merge in develop. Otherwise, I added notes about a couple small tasks that should get these new tests passing.

tests/bdd/features/file_storage.feature
comkieffer: Fixed, I've updated the dates.

tests/bdd/features/file_storage.feature
comkieffer: I can't replicate this. I'm not sure what might have caused the issue. 

I re-ran the tests without the `empty` file in `test/data/journals/empty_folder` and the tests still passed.

micahellison: Looks like bug #1894 is Windows-only, and this is combined with a limitation of git: since we can only commit files (not directories), that `empty_folder` doesn't exist on the remote repository until we push a file in it.

jrnl/journals/FolderJournal.py
```diff
@@ -143,15 +148,16 @@ def _get_month_folders(path: pathlib.Path) -> list[pathlib.Path]:
         return
 
     @staticmethod
-    def _get_day_files(path: pathlib.Path) -> list[str]:
-        for child in path.glob(DAY_PATTERN):
+    def _get_day_files(path: pathlib.Path, extension: str) -> list[str]:
+        for child in path.glob(DAY_PATTERN + extension):
```
micahellison, line 152: I think the only thing left on this PR is supporting a lack of dots in the extension. For instance, I'd expect this to work the same whether my config value is `extension: md` or `extension: .md`

PR #1874, opened 2024-03-23
Correctly extract titles with multiple line breaks

### LINKED ISSUE #1676: Starting an entry with multiple line breaks splits the end of the first line by the number of line breaks
### Diagnostic output

jrnl: v3.3
Python: 3.11.1 (tags/v3.11.1:a7a450f, Dec  6 2022, 19:58:39) [MSC v.1934 64 bit (AMD64)]
OS: Windows 10

### Current Behavior

Run `jrnl` by itself, then, in your editor, add at least two line breaks and type a sentence. After you save, it will insert a single line break before the end of your sentence. The distance before the end is the number of prepended line breaks minus 1.

### Expected Behavior

The entry shouldn't be cut off just because of prepended line breaks.

### Repro Steps

* Run `jrnl`
* In your editor, press enter twice, then type "entry"
* Run `jrnl -1`
* You'll see something like this:
```
[2023-02-11 11:47:35 AM] entr
y
```

* Run `jrnl`
* In your editor, press enter **three times**, then type "entry"
* Run `jrnl -1`
* You'll see something like this:
```
[2023-02-11 11:48:04 AM] ent
ry
```

### Debug output

```
DEBUG    root         Parsed args: Namespace(debug=True, preconfig_cmd=None, postconfig_cmd=None, filename=None, on_date=None, today_in_history=False, month=None, day=None, year=None, start_date=None, end_date=None, contains=None, strict=False, starred=False, limit=None, excluded=[], edit=False, delete=False, change_time=None, export=False, tags=False, short=False, config_override=[], config_file_path='', text=[])
DEBUG    root         Using configuration "{'colors': {'body': 'none', 'date': 'none', 'tags': 'none', 'title': 'none'}, 'default_hour': 9, 'default_minute': 0, 'display_format': 'text', 'editor': 'subl -w', 'encrypt': False, 'highlight': True, 'indent_character': '|', 'journals': {**************}, 'linewrap': 79, 'tagsymbols': '#@', 'template': False, 'timeformat': '%F %r', 'version': 'v3.3'}"
DEBUG    root         Updating configuration with specific journal overrides {'encrypt': False, 'journal': 'C:\\Users\\micah\\Documents\\jrnl\\default.txt'}
DEBUG    root         opened PlainJournal with 11 entries
DEBUG    root         Write mode: starting
DEBUG    root         Write mode: opening editor
DEBUG    root         Write mode: loading template for entry
DEBUG    root         Write mode: no template configured
DEBUG    root         Write mode: appending raw text to journal "default":

entry
DEBUG    root         Write mode: completed journal.write()
```

### Other Information

I also tested this with the default stdin editor and saw the behavior. @wren reproduced this on a Mac also so it's not Windows-specific.

comkieffer:
Closes #1676 (Starting an entry with multiple line breaks splits the end of the first line by the number of line breaks)

Since the location of the separator was computed on the `text.lstrip()` but the function returned the substring of `text` the title would be cut off.

### Checklist

- [X] I have read the [contributing doc](https://github.com/jrnl-org/jrnl/blob/develop/docs/contributing.md).
  for the same issue.

### DIFF
--- a/jrnl/journals/Entry.py
+++ b/jrnl/journals/Entry.py
@@ -223,7 +223,8 @@ SENTENCE_SPLITTER_ONLY_NEWLINE = re.compile("\n")
 
 def split_title(text: str) -> tuple[str, str]:
     """Splits the first sentence off from a text."""
-    sep = SENTENCE_SPLITTER_ONLY_NEWLINE.search(text.lstrip())
+    text = text.lstrip()
+    sep = SENTENCE_SPLITTER_ONLY_NEWLINE.search(text)
     if not sep:
         sep = SENTENCE_SPLITTER.search(text)
         if not sep:


comkieffer:
I've rebased this onto the latest `develop/`. Let's see if the tests pass.

Issue #1875, opened 2024-03-24
Tests failing on develop branch starting with pytest-bdd 7.1.2


micahellison:
### Diagnostic output

n/a

### Current Behavior

Around a week ago, pytest-bdd 7.1.2 was released and all of our PRs started failing tests with this error on `tests/lib/when_steps.py`:
```
ImportError: Error importing plugin "tests.lib.when_steps": cannot import name 'inject_fixture' from 'pytest_bdd.steps'
```

### Expected Behavior

Tests should be running fine. This patch release of pytest-bdd is a breaking change for us, though.

### Repro Steps

* Open pyproject.toml, pin pytest-bdd version: `pytest-bdd = "7.1.1"`
* Run `poetry install` and `poetry run pytest`. Everything passes.
* Open pyproject.toml, pin pytest-bdd version: `pytest-bdd = "7.1.2"`
* Run `poetry install` and `poetry run pytest`. Everything fails.

### Debug output

n/a

### Other Information

I don't have a lot of time to look at this now, and am tempted to pin pytest to 7.1.1 for now just to be able to get through PRs, though if a contributor has a fix for this, it would be welcome.

micahellison:
It looks like version pinning is a bit trickier than I'd thought, I think due to irregularities between how poetry and tox handle pytest plugin versions. Probably better to spend time on just fixing the root cause.

musicinmybrain:
https://github.com/pytest-dev/pytest-bdd/issues/684

musicinmybrain:
The immediate workaround would be something like:

```python
try:
    from pytest_bdd.compat import inject_fixture  # since pytest_bdd 7.1.2
except ImportError:
    from pytest_bdd.steps import inject_fixture  # before pytest_bdd 7.1.2
```

but perhaps it would be worth considering, in the medium term, whether there is an alternative to using an unsupported function that could move again or disappear.

musicinmybrain:
Alternatively, I guess you could just `from pytest_bdd.compat import inject_fixture` and update to `pytest_bdd>=7.1.2`.

PR #1876, opened 2024-03-24
Pin pytest-bdd to keep tests working

micahellison:
This PR pins pytest-bdd to 7.1.1 to keep tests from breaking due to #1875. It doesn't fix the problem but it will at least allow tests to run on PRs.

### Checklist

  for the same issue.


micahellison:
It looks like tox and poetry use different version resolvers for pytest plugins in particular. I'm not sure how to actually make this work in our test runner. Closing this for now to avoid confusion. I think it's probably a better use of time to just fix the root cause by importing fixtures differently to appease pytest-bdd 7.1.2.

PR #1889, opened 2024-05-07
Recommend pipx as easiest installation method for all OSes and remove warning about apt

### LINKED ISSUE #1886: Remove documentation recommendation to install pipx through brew or pip
### Affected Page(s)

https://jrnl.sh/en/stable/installation/

### What Could Be Better?

I'd like to challenge the advice "pipx should be installed through brew or pip, not apt".  As far as I can see, the only sensible way to install jrnl on Debian is to install pipx through apt and jrnl with pipx.  Installing pipx with pip on Debian is not possible (it just refuses). Using brew on Debian seems like a weird and dangerous thing to do.

pipx's job is to manage obtaining the dependencies of a package in a safe, isolated way. I would be surprised if any dependency problems emerged (and it worked perfectly for me on bookworm).  At a guess the note in the docs dates from a time when having an old version of pipx itself could cause problems, but that is very stable now.

I'm not a heavy python developer and the tooling always confuses me, so I might be completely wrong, but for me the only choice for installing jrnl that I was happy with was the one that is specifically advised against.

### Other Information

_No response_

### LINKED ISSUE #1888: Recommend pipx as default installation method
### Affected Page(s)

https://jrnl.sh/en/stable/installation/

### What Could Be Better?

I think it's time to recommend pipx over brew for all OSes.

It's a different world than when the installation docs were written. pipx has become easier to install and hasn't had any packaging issues with jrnl in years, whereas brew takes a long time to install and its lack of isolation has led to frequent issues for jrnl users. Moreover, the maintenance burden of supporting pipx is practically nothing, whereas the added complexities introduced by brew have led to a lot of choring for us rather than working on features.

This isn't to knock brew, which of course has found its place in the big wide world of package managers, but for our purposes, and especially for new users, I think pipx is the answer moving forward.

### Other Information

Hat tip to #1886 which also hints at this change.

micahellison:
Fixes #1886 and #1888.

### Checklist

  for the same issue.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -20,11 +20,6 @@ On other platforms, install `jrnl` using [Python](https://www.python.org/) 3.10+
 pipx install jrnl
 ```
 
-!!! note
-     `pipx` should be installed through either `brew` or `pip`. Missing dependencies and other issues
-     may occur when installing `pipx` through `apt` or another package manager. Further installation
-     instructions can be found in [pipx's documentation](https://pipxproject.github.io/pipx/installation/).
-     
 !!! tip
      Do not use `sudo` while installing `jrnl`. This may lead to path issues.
 
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -7,14 +7,9 @@ License: https://www.gnu.org/licenses/gpl-3.0.html
 
 ## Installation
 
-On Mac and Linux, the easiest way to install `jrnl` is using
-[Homebrew](http://brew.sh/):
-
-``` sh
-brew install jrnl
-```
-
-On other platforms, install `jrnl` using [Python](https://www.python.org/) 3.10+ and [pipx](https://pipxproject.github.io/pipx/):
+The easiest way to install `jrnl` is using
+[pipx](https://pipx.pypa.io/stable/installation/)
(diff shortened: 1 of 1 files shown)


wren (approved):
🍺

philoserf:
Is `pipx` really easier?

```shell
brew install pipx
pipx ensurepath
pipx install jrnl
```

over

```shell
brew install jrnl
```

cc: https://github.com/jrnl-org/jrnl/issues/1886 and https://github.com/jrnl-org/jrnl/issues/1888.

wren:
> Is `pipx` really easier?
> 
> ```shell
> brew install pipx
> pipx ensurepath
> pipx install jrnl
> ```
> 
> over
> 
> ```shell
> brew install jrnl
> ```
> 
> cc: https://github.com/jrnl-org/jrnl/issues/1886 and https://github.com/jrnl-org/jrnl/issues/1888.

Yup! If you already have `brew` installed, then that will, of course, be easier. But the same is true if you already have any package manager installed (including `pipx`).

We've found `pipx` to be the most reliable way to install Python apps across OSes (remember that `jrnl` supports MacOS, *Nix, and Windows).

PR #1914, opened 2024-08-18
Issue 1913/docs add more datetime examples

### LINKED ISSUE #1913: Have more examples of supported humanized datetime formats
### Affected Page(s)

https://jrnl.sh/en/stable/usage/#composing-entries section "Specifying Date and Time"

### What Could Be Better?

Adding a more extensive list of supported humanized datetime formats

### Other Information

_No response_

chwt:
Closes #1913 

To have more examples of supported supported humanized datetime strings, added a file extracted from unit tests of the used datetime parsing dependency (https://github.com/bear/parsedatetime).

Questions/comments:
- I rather wanted to have as many examples as possible, but is the current list too much?
  - The file is located in a new directory `docs/res/`. Should it be moved somewhere else?
  - First, I had the grep command, which is used to extract the file, in the file itself as header. Then thought that it will rather confuse the documentation users. Currently it is part of the commit message. What would be alternatives?
- Please provide any suggestions for better phrasing of the added documentation change.

### Checklist

  for the same issue.
- [ ] ~I have written new tests for these changes, as needed.~ (docs only)

### DIFF
--- a/docs/res/datetime-examples-raw.txt
+++ b/docs/res/datetime-examples-raw.txt
@@ -1,0 +1,408 @@
+sunday
+sun
+su
+Monday
+mon
+mond
+tuesday
+tues
+tue
+wednesday
+wedn
+wed
+thursday
+thu
+thur
+thurs
+friday
+fri
+frid
+saturday
+sat
+sa
+Aug 05, 2014 4:15 AM
+Aug 05, 2003 3:15 AM
+Aug 05, 2003 03:15 AM
+June 30th 12PM
+June 30th 12:00
+December 30th 23PM
+December 30th 23:02
+08/25/2006 5pm
+5pm on 08.25.2006
+5pm August 25, 2006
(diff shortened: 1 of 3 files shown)


micahellison (changes requested):
Hi @chwt, thanks for the PR. I appreciate the intent to add more examples, though I don't want to add this large list to the documentation. I also want to minimize how much we're documenting dependencies' behavior -- that's usually better done by the dependencies themselves.

I think we might be better off just linking to the [part of parsedatetime's code that deals with this logic](https://github.com/bear/parsedatetime/tree/2a4b16901794f8dc71721d51aef6e1ecd643868b/parsedatetime/pdt_locales). That would also have the nice side effect of showing what other options are available in non-English locales.

chwt:
Understandable. Potentially, those interested in the supported syntax will find this pull request for reference.

PR #1918, opened 2024-09-24
Feature - add bracketed date support when composing new entry (#1915)

### LINKED ISSUE #1915: Support bracketed date style when composing new entry
### Use Case/Motivation

Currently, you can use command-line style dates to specify a new entry's date when composing new entries, but not the bracketed dates that jrnl uses in its storage format and when editing entries.

### Example Usage

Run `jrnl` with no arguments. Your editor or the standard input editor will appear.

You can already do this, which is good and worth keeping:
```
tomorrow: Here is my entry
```
The entry will be written with tomorrow's date.

However, you should also be able to do this:
```
[2024-08-26 08:00] Here is my entry on August 26
```
Currently, that would create a new entry with today's time, and the entry would contain `[2024-08-26 08:00] Here is my entry on August 26`. This enhancement should instead parse that bracketed value and created a new entry on the specified date, where the entry only contains: `Here is my entry on August 26`.

### Other Information

Hat tip @alichtman for filing #1906 which is one step ahead of this enhancement.

A resolution to this should:
- [ ] Support this format in compose mode in a way that is the same as in the edit mode
- [ ]  Add new tests as necessary

chwt:
Bracketed dates like `[2020-01-01]` at the beginning of an entry will be parsed and used as the entry date, thus be removed from the entry text.

Colon `:` and starred `*` usage is mostly covered too, see test cases.

Fixes #1915 

### Checklist

  for the same issue.

### DIFF
--- a/jrnl/journals/Journal.py
+++ b/jrnl/journals/Journal.py
@@ -187,6 +187,7 @@ class Journal:
                 new_date = time.parse(date_blob, bracketed=True)
 
             if new_date:
+                # Fill in text of entry created in previous loop
                 if entries:
                     entries[-1].text = journal_txt[last_entry_pos : match.start()]
                 last_entry_pos = match.end()
@@ -363,7 +364,7 @@ class Journal:
         raw = raw.replace("\\n ", "\n").replace("\\n", "\n")
         # Split raw text into title and body
         sep = re.search(r"\n|[?!.]+ +\n?", raw)
-        first_line = raw[: sep.end()].strip() if sep else raw
+        first_line = raw[:sep.end()].strip() if sep else raw
         starred = False
 
         if not date:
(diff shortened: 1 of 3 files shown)


jrnl/journals/Journal.py
```diff
@@ -376,7 +377,29 @@ def new_entry(self, raw: str, date=None, sort: bool = True) -> Entry:
                 )
                 if date:  # Parsed successfully, strip that from the raw text
                     starred = raw[:colon_pos].strip().endswith("*")
-                    raw = raw[colon_pos + 1 :].strip()
+                    raw = raw[(colon_pos + 1):].strip()
+            else:
+                date_blob_re = re.compile("(?:^|\n)\\[([^\\]]+)\\] ")
```
chwt, line 382: **note:** code is based on `Journal._parse` with slight adaptions. E.g. the `default_hour` and `default_minute` handling is added.

tests/bdd/features/datetime.feature
```diff
@@ -51,6 +51,13 @@ Feature: Reading and writing to journal with custom date formats
         | little_endian_dates.yaml | 2032-02-01: Test.            | 01.02.2032 09:00 Test.            |
         | little_endian_dates.yaml | 2020-01-01: Test.            | 01.01.2020 09:00 Test.            |
         | little_endian_dates.yaml | 2020-12-31: Test.            | 31.12.2020 09:00 Test.            |
+        # @todo: is it fine that default time be used here and not 00:00?
```
chwt, line 54: **question:** is this behavior fine?

tests/bdd/features/datetime.feature
```diff
@@ -78,6 +85,14 @@ Feature: Reading and writing to journal with custom date formats
         Then the output should contain "10.05.2013 09:00 I saw Elvis."
         And the output should contain "He's alive."
 
+    Scenario: Writing an entry at the prompt with custom date in bracket format
```
chwt, line 88: **note:** could be integrated into test above as config

tests/bdd/features/star.feature
```diff
@@ -16,6 +16,31 @@ Feature: Starring entries
...
+        When we run "jrnl <command>"
+        Then we should get no error
+        When we run "jrnl -on 2013-07-20 -starred"
+        Then the output should contain "2013-07-20 09:00 Best day of my life!"
+
+        Examples: configs
+        | config_file       | command |
+        | simple.yaml       | [2013-07-20] Best day of my life! * |
+        | empty_folder.yaml | [2013-07-20] Best day of my life! * |
+        # Note: this one fail due to whitespace, cmp. next config
```
chwt, line 30: **question:** I assume the dayone Journal should also be supported without issues? Will have to check that

PR #2028, opened 2025-08-15
For Python 3.11, use `tomllib` for tests

### LINKED ISSUE #2027: Feature Report: Drop test dependency on PyPI `toml` package
### Use Case/Motivation

Since Python 3.11, the test dependency on [`toml`](https://pypi.org/project/toml/) is no longer needed because [`tomllib`](https://docs.python.org/3/library/tomllib.html) was added to the standard library.

### Example Usage

Testing on Python 3.11 and later should not require installing `toml`.

### Other Information

While `jrnl` does still support Python 3.10, where the `toml` dependency is needed, conditionalizing the dependency for Python 3.11 or later would make testing slightly lighter and would help with downstream packaging, e.g. in Fedora where the `python-toml` package has been [deprecated for some time](https://fedoraproject.org/wiki/Changes/DeprecatePythonToml) and will soon be removed.

I plan to open a PR that would use `tomllib` instead of `toml` on Python 3.11 and later.

musicinmybrain:
Avoid a test dependency on the PyPI [`toml` package](https://pypi.org/project/toml/).

Fixes #2027.

### Checklist

- [x] I have included a link to the relevant issue number. **N/A**
  for the same issue.
- [x] I have written new tests for these changes, as needed. **Covered by existing tests.**


musicinmybrain:
The `toml` case is written as

```python
    pyproject_contents = toml.load(pyproject_path)
```

in this PR, which is concise and similar to what we had before. We could have instead written

```python
    with open(pyproject_path) as pyproject:
        pyproject_contents = toml.load(pyproject)
```

which would more directly parallel the `tomllib` case. However, it’s slightly more verbose, and it wouldn’t allow us to share the part that opens the file, since `tomllib` wants the file opened in binary mode and `toml` doesn’t, so there’s no real advantage in writing it that way.

micahellison (approved):

Issue #1905, opened 2024-06-06
jrnl-org/nostr?


ghost:

micahellison:
I don't want to add any network code to jrnl for a few reasons, mainly because it's a huge maintenance burden to support the shifting landscape of protocols and security issues. We would have to release much more frequently, and we'd still likely do a worse job than any of these third-party tools. Also, I think one of the things that makes jrnl trustworthy is that it has no networking capability at all. I think any new network feature would compromise that trust.

However, if you want to do this with jrnl, the feature that would enable this would be application hooks as described in #921. You'd use whatever command line tool you use to post to nostr using data fed from a hook on a new entry. Or, for a less resilient but immediately usable solution, you can use a shell script wrapped by an alias as described in [this comment](https://github.com/jrnl-org/jrnl/issues/921#issuecomment-1793845901).

PR #2058, opened 2026-02-22
feat: add semantic search for journal entries

keithagroves:
## Motivation

Journal entries often use varied language to describe similar experiences. Keyword search (`-contains`) requires guessing exact words, which misses semantically related entries. For example, searching for "feeling anxious" will not find entries about "stress" or "worry". Semantic search closes this gap by matching entries by meaning.

There is no existing issue for this feature — happy to open one if preferred.

## What this PR does

Adds natural-language semantic search using [fastembed](https://github.com/qdrant/fastembed) (ONNX) embeddings with a SQLite-backed index. Entries are matched by meaning rather than exact keywords.

### New CLI flags
- `--index-semantic` — Build/rebuild the semantic search index
- `-semantic QUERY` — Search entries by semantic similarity

### New modules
- `jrnl/search/` — Embedding, indexing, crypto, and entry ID logic
  - `embeddings.py` — fastembed model loading and text encoding
  - `index.py` — SQLite-backed vector index with brute-force cosine similarity
  - `crypto.py` — Per-row Fernet encryption for encrypted journal support
  - `entry_id.py` — Stable entry identification via SHA256

### Config keys (`semantic_search.*`)
| Key | Default | Description |
|-----|---------|-------------|
| `model` | `BAAI/bge-small-en-v1.5` | Embedding model |
| `threshold` | `0.4` | Minimum similarity score |
| `top_k` | `10` | Max results |
| `auto_index` | `false` | Auto-update index on write |

### Optional dependency
```
pip install jrnl[semantic]
```
Uses fastembed (ONNX runtime) — no PyTorch required, ~50 MB model download on first use.

### Encrypted journals
Semantic index is encrypted with a separate key derived from the journal password via PBKDF2 with a distinct salt, so embeddings are never stored in plain text.

## Example usage

```sh
# Build the index
jrnl --index-semantic

# Search by meaning
jrnl -semantic "outdoor adventures"
jrnl -semantic "feeling stressed" -n 5
jrnl -semantic "travel" -from 2025
```

## Tests

- 4 new unit test files covering index sync, search, entry ID, and crypto
- All 165 existing unit tests continue to pass

## Checklist

- [ ] I have included a link to the relevant issue number. *(No existing issue — happy to create one if needed.)*
- [x] I have checked to ensure there are not other open pull requests for the same issue.


micahellison:
Hi there, I think this is a cool idea, but I'm very resistant to adding this to jrnl's code base, for a variety of reasons:

* I'm not interested in spending the time to maintain it
* It adds a whole new persistence layer that increases security risks
* The download-on-demand feature effectively adds network functionality to jrnl, which I [oppose](https://github.com/jrnl-org/jrnl/issues/1905)
* This moves jrnl further away from the [Unix philosophy](https://en.wikipedia.org/wiki/Unix_philosophy), and while we don't always stick to it, we have always paid many hours in bug fixes and support whenever we've deviated for it

However, I don't want to discourage you from creating something like this. If you want something like this to grow and thrive in the world, it will have to be downstream from jrnl, instead of inside jrnl. In other words, you can make your own jrnl semantic search project using jrnl as a dependency. If you'd like to take such an approach, I wish you well.

keithagroves:
Hey, I completely understand. Seems a little heavy for jrnl so I figured the odds were low. It was mostly for my personal use case but wanted to see if there was interest. I'll make my downstream version available if anyone wants to use it.

PR #2064, opened 2026-03-04
Feature: warning for similar tags

hanbeex:
### Description

This PR adds a fuzzy tag similarity warning when writing a new journal entry. If any tag in the new entry closely resembles an existing tag in the journal, jrnl prints a warning suggesting the user double-check for typos.

**Motivation:** Tag typos silently fragment what should be a single tag group. For example, accidentally writing `@works` instead of `@work` results in two separate tags that are hard to notice and tedious to fix later. Surfacing this immediately after writing to make it easy to catch and correct.

**What it does:**
- Before creating the new entry, the current set of existing tags is captured
- After saving, the new entry's tags are compared against existing ones using `difflib.get_close_matches`  with a similarity cutoff of 0.8
- A warning is printed for each new tag that is suspiciously close to an existing one
- Only tags sharing the same leading symbol are compared, so intentionally distinct tag namespaces (e.g. `@` for people, `#` for topics) don't produce false positives — this also means custom `tagsymbols` from the config file are handled correctly
- The warning is non-blocking: the entry is always saved, no confirmation prompt is shown

**Only outputs a warning and doesn't ask 'Are you sure?' to the user** Some similar-looking tags are perfectly meaningful to keep separate. For example, `@health` and `@wealth`, `@friend` and `@fiend`, or `@hiking` and `@biking` are all ≥80% similar but represent distinct concepts. Asking "are you sure?" every time would be more annoying than helpful.

**Example output:**

```
$ jrnl Today I finished the @works report

┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃  The tag '@works' is similar to: @work                                                                                                                       ┃
┃                                                                                                                                                                                            ┃
┃  Was this intentional? If not, re-open the entry with 'jrnl --edit' to fix it.                                                              ┃
┗━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┛
```

**Files changed:**
- `jrnl/controller.py` - added `_check_similar_tags()` helper and hooked it into `append_mode`
- `jrnl/messages/MsgText.py` - added `SimilarTagsFound` warning message
- `tests/unit/test_controller.py` - added `TestCheckSimilarTags` with 8 unit tests
- `docs/usage.md` - documented the warning and explained why no confirmation is shown
- `jrnl/__version__.py` / `pyproject.toml` - fixed a pre-existing version string mismatch (`v4.3` vs `4.3`) that was causing 6 BDD test failures on the `develop` branch

### Checklist

- [x] I have included a link to the relevant issue number. https://github.com/jrnl-org/jrnl/issues/2065
  for the same issue.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,7 +36,7 @@ Funding = "https://opencollective.com/jrnl"
 jrnl = 'jrnl.main:run'
 
 [tool.poetry]
-version = "4.3"
+version = "v4.3"
 classifiers = [
   "Topic :: Office/Business :: News/Diary",
   "Environment :: Console",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,7 +36,7 @@ Funding = "https://opencollective.com/jrnl"
 jrnl = 'jrnl.main:run'
 
 [tool.poetry]
-version = "v4.3"
+version = "4.3"
 classifiers = [
   "Topic :: Office/Business :: News/Diary",
   "Environment :: Console",
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -77,6 +77,28 @@ case-insensitive.
 
 There is no limit to how many tags you can use in an entry.
 
+#### Similar Tag Warning ####
+
+When you write an entry, `jrnl` checks whether any of its tags closely resemble
+tags you've used before. If so, it prints a warning:
+
+```
+The tag '@works' is similar to: @work
+
+Was this intentional? If not, re-open the entry with 'jrnl --edit' to fix it.
+```
+
+This is just a warning — `jrnl` will not ask for confirmation or block the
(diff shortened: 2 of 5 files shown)


pyproject.toml
```diff
@@ -36,7 +36,7 @@ Funding = "https://opencollective.com/jrnl"
 jrnl = 'jrnl.main:run'
 
 [tool.poetry]
-version = "4.3"
```
alichtman, line 39: drop this change

Issue #2021, opened 2025-08-10
jrnl crashes when `colors.tags` is not configured


matthiasbeyer:
### Diagnostic output

```
┏━ Error ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃  KeyError                                            ┃
┃  'tags'                                              ┃
┃                                                      ┃
┃  This is probably a bug. Please file an issue at:    ┃
┃  https://github.com/jrnl-org/jrnl/issues/new/choose  ┃
┗━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┛
```

### Current Behavior

Crash as listed above

### Expected Behavior

Tags are not colored, or an error about missing configuration is printed

### Repro Steps

1. Use the default configuration when creating a new jrnl and no config is present
2. `jrnl "Some @thing"`
3. `jrnl -n 1`

### Debug output

Relevant part from the `--debug` output:

```
│ /nix/store/8l0rdxb3ysqr41sp8jpbhm3wv9jf0lv8-jrnl-4.2/lib/python3.12/site-packages/jrnl/color.py: │
│ 56 in colorized_text_generator                                                                   │
│                                                                                                  │
│   55 │   │   │   elif part:                                                                      │
│ ❱ 56 │   │   │   │   yield colorize(part, config["colors"]["tags"], bold=True), part             │
│   57                                                                                             │
```

matthiasbeyer:
Hm. After checking the source code, I found that the configuration is indeed generated!

Probably here: https://github.com/jrnl-org/jrnl/blob/develop/jrnl/config.py#L99
(I am not a python person).

Anyways, a missing configuration key should not make the application crash!

PR #2079, opened 2026-05-16
fix: use safe dict access for color/tag config keys (#2021)

miloquinn:
## Problem

When `colors.tags`, `tagsymbols`, `highlight`, or `colors` are missing from the config, jrnl crashes with `KeyError`. This happens with default config on first run or manually edited configs.

```
KeyError: \tags'


alichtman:
This bug should be fixed on `develop` already! Thanks for the contribution :)

Issue #1853, opened 2024-02-02
Add 1password support for saving and retrieving jrnl passwords


micahellison:
### Use Case/Motivation

Currently, jrnl uses [keyring](https://pypi.org/project/keyring/) to save passwords on the keyring backend supported by the operating system (e.g. Keychain on macOS or Windows Credential Locker on Windows). However, keyring has support for some [third-party backends](https://github.com/jaraco/keyring?tab=readme-ov-file#third-party-backends), including the popular password manager 1password through the [onepassword-keyring](https://pypi.org/project/onepassword-keyring/) project. It would be nice to implement this in jrnl.

### Example Usage

I'm not sure how best to configure this. Maybe we add a new keyring field in the config file.

Otherwise, it shouldn't act too differently than how Keychain works with jrnl on MacOS: it would prompt the user to unlock the vault with a password, then the vault would return the specific jrnl password to jrnl.

PR #2090, opened 2026-07-15
feat: add 1Password keyring backend support

alichtman:
## Needs more cleanup and deeper review from me!

Adds an opt-in 1Password keyring backend, on top of several fixes to existing keyring handling that surfaced during development.

## 1Password backend

- New jrnl/encryption/OnePasswordBackend.py implements keyring.backend.KeyringBackend by shelling out to the `op` CLI directly (list-arg subprocess calls, no shell string building; exit-code checks instead of stderr text matching). Each (service, username) pair maps to its own item titled 'jrnl/<journal_name>', so journals never collide.
- Opt-in only, via `keyring_backend: onepassword` in config. The backend is never registered through keyring's own priority-based auto-discovery, so it can't tie with or shadow a user's OS keychain.
- jrnl/prompt.py offers to use 1Password instead of the OS keychain when `op` is detected and no keyring_backend is set yet; the choice is persisted so the question isn't asked again.
- Before overwriting an existing 1Password item, the user is asked to confirm.

## Keyring handling fixes

- jrnl/keyring.py: get/set failures now report the journal, backend, and underlying reason (previously a single generic "Failed to retrieve keyring" string covered both get and set failures). Successful get/set now also print a visible confirmation instead of being silent.
- jrnl/main.py: stop suppressing keyring.backend's own logger to ERROR, so `--debug` shows which backends keyring discovers and selects.
- BasePasswordEncryption.decrypt() now offers to save a manually entered password to the keyring after a successful decrypt, not only when a password is first created via --encrypt. Opt out with `prompt_to_add_to_keyring_on_successful_decrypt: false`.
- Fixed: postconfig_encrypt/postconfig_decrypt persisted only a hardcoded {"encrypt": ...} to config, ignoring Journal._pending_config_updates. This silently dropped any keyring_backend choice made during --encrypt/ --decrypt (it worked for the rest of that process, but never reached disk). Journal._encrypt()/_decrypt() now queue keyring_backend changes into _pending_config_updates, and both postconfig commands merge it in.

## Tests

- New: test_onepassword_backend.py, test_keyring.py, test_prompt.py, test_commands.py, test_journal_keyring_backend_persistence.py, test_decrypt_keyring_offer.py.
- New BDD scenarios for the decrypt-time save offer (accept/decline).
- tests/lib/fixtures.py: OnePasswordBackend.is_available() and the new decrypt-offer prompt default to off in BDD runs, so existing scenarios don't depend on the test machine's PATH or need to script answers to prompts they don't care about.

Addresses #1853

### Checklist

  for the same issue.

### DIFF
--- a/docs/privacy-and-security.md
+++ b/docs/privacy-and-security.md
@@ -39,9 +39,35 @@ want to "store the password in your keychain." This keychain is accessed using
 the [Python keyring library](https://pypi.org/project/keyring/), which has different
 behavior depending on your operating system.
 
+If your password isn't found in the keychain and you're prompted to type it in
+manually — for example while just reading a journal, not only when first
+encrypting it — `jrnl` will also offer to save that password to the keychain
+once it's confirmed to work. To turn this off, set
+`prompt_to_add_to_keyring_on_successful_decrypt: false` in your config.
+
 In Windows, the keychain is the Windows Credential Manager (WCM), which can't be locked
 and can be accessed by any other application running under your username. If this is
 a concern for you, you may not want to store your password.
+
+### Using 1Password instead of your OS keychain
+
+Set `keyring_backend: onepassword` in your config (globally, or as a
+per-journal override) to store passwords as Login items in
+[1Password](https://1password.com) instead of your OS keychain. This requires
+the [1Password CLI](https://developer.1password.com/docs/cli/get-started/)
+(`op`) to be installed and signed in. Each journal's password is stored under
+its own item, titled `jrnl/<journal name>`, so multiple journals don't
+collide.
+
+If `keyring_backend` isn't already set and `jrnl` detects the `op` CLI on your
+system, it will ask whether you'd like to use 1Password the next time you
+create a new password; answering yes saves the choice to your config so
+you're not asked again.
+
+If `keyring_backend` is unset (the default), or if `op` isn't installed,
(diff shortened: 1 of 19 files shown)


docs/privacy-and-security.md
```diff
@@ -39,10 +39,36 @@ want to "store the password in your keychain." This keychain is accessed using
...
+
+Set `keyring_backend: onepassword` in your config (globally, or as a
+per-journal override) to store passwords as Login items in
+[1Password](https://1password.com) instead of your OS keychain. This requires
+the [1Password CLI](https://developer.1password.com/docs/cli/get-started/)
+(`op`) to be installed and signed in. Each journal's password is stored under
+its own item, titled `jrnl/<journal name>`, so multiple journals don't
+collide.
+
+If `keyring_backend` isn't already set and `jrnl` detects the `op` CLI on your
```
wren, line 62: Encryption config currently exists on a per journal level, but this new config key would be global. How does this implementation handle journals with different requirements (one journal uses keyring, and another uses 1password)?

Maybe `keyring` can be folded into the `encryption` config key? Or in the journal config?

```yaml
journals:
  default:
    journal: ~/Documents/Journals/notes-archive.jrnl
    encrypt: true
    keyring: 1password
```

alichtman: Another option would be putting the keyring settings in the jrnlv3 encrypted file header.

PR #2105, opened 2026-07-20
fix: don't treat runs made up only of tagsymbols as tags

### LINKED ISSUE #1700: Text containing only tagsymbols should not be treated as tags
### Diagnostic output

jrnl: v3.3
Python: 3.11.1 (tags/v3.11.1:a7a450f, Dec  6 2022, 19:58:39) [MSC v.1934 64 bit (AMD64)]
OS: Windows 10

### Current Behavior

With the default tagsymbols `#` and `@`, it's easy to enter text that isn't meant to be a tag but is still registered as one. For instance, a markdown heading starting with `###` adds `###` as a tag.

It is doing it in some cases, but it's strangely inconsistent.

### Expected Behavior

When running `jrnl --tags`, no text should be treated as a tag if it's composed entirely of tagsymbols.

### Repro Steps

I ensure my tagsymbols are set to the default `#@`.

I create a new journal entry with the following:
```
@@
##
@#@
#@#
#@#@#
@#@#@
###
@@@
####
```
I run `jrnl --tags`:
```
@#                   : 1
####                 : 1
###                  : 1
##                   : 1
```
None of those should appear as tags.

### Debug output

```
DEBUG    root         Parsed args: Namespace(debug=True, preconfig_cmd=None, postconfig_cmd=None, filename=None, on_date=None, today_in_history=False, month=None, day=None, year=None, start_date=None, end_date=None, contains=None, strict=False, starred=False, limit=None, excluded=[], edit=False, delete=False, change_time=None, export=False, tags=True, short=False, config_override=[], config_file_path='', text=[])
DEBUG    root         Using configuration "{'colors': {'body': 'none', 'date': 'black', 'tags': 'yellow', 'title': 'cyan'}, 'default_hour': 9, 'default_minute': 0, 'display_format': None, 'editor': 'subl -w', 'encrypt': False, 'highlight': True, 'indent_character': '| ', 'journals': {***********}, 'linewrap': 80, 'tagsymbols': '#@', 'template': False, 'timeformat': '%F %r', 'version': 'v3.3'}"
DEBUG    root         Updating configuration with specific journal overrides {'encrypt': False, 'journal': 'C:\\Users\\micah\\Documents\\jrnl\\default.txt'}
DEBUG    root         opened PlainJournal with 1 entries
@#                   : 1
####                 : 1
###                  : 1
##                   : 1
```

### Other Information

It's important that this only applies to tokens that are entirely composed of `tagsymbols`. I think a tag like `#abc#def` should still be read as one tag, even though it contains the tag symbol twice.

rahul188:
Fixes #1700.

### What's happening

`Entry.tag_regex()` builds `(?<!\S)([{tagsymbols}][-+*#/\w]+)`. With the default
tagsymbols `#@`, the trailing character class `[-+*#/\w]` itself contains `#`, so
a run made up purely of tagsymbols matches the pattern. A markdown heading like
`###` — or any of the reporter's cases (`@@`, `@#@`, `####`) — was collected as a
tag and listed by `jrnl --tags`.

The inconsistency the reporter noticed comes from the same place: only runs
starting with a symbol that is *also* in the trailing class can match, which is
why `###` showed up but `@@@` did not.

### The change

`_parse_tags()` now drops a match when every one of its characters is a
tagsymbol. That's deliberately the narrowest rule that fixes the report — as
called out in the issue, a tag that merely *contains* a tagsymbol (`#abc#def`)
is still parsed as one tag, and it's covered by a test so it stays that way.

### Testing

New `tests/unit/test_entry.py` with three cases: the reporter's exact repro block
yields no tags, real tags alongside a `##` heading still parse, and `#abc#def`
stays a single tag. The first two fail on `main` and pass with the fix; the third
passes both ways and guards against over-correcting.

Full unit suite: 187 passed, 27 skipped. `ruff check` and `black --check` clean.

### Checklist

- [x] I have read the [contributing doc](https://github.com/jrnl-org/jrnl/blob/main/docs/contributing.md).
- [x] I have checked to ensure there aren't other open [pull requests](../pulls) for the same issue.


rahul188:
Hi 👋 Gentle nudge on this PR — it's **mergeable and CI is passing**, just waiting on a review. Would really appreciate a look whenever you have bandwidth, and I'm glad to make any changes you'd like. Thanks for your time and for maintaining this project! 🙏

MohammedAlkindi:
Reproduced and confirmed on Windows 11 / Python 3.14.7. Probe: https://gist.github.com/MohammedAlkindi/9e2161d587c805fea4bc4f41b1c38feb

On `main` at `90de494`, then on this branch at `83a1247`:

```
input                       main            this branch
'### heading'               ['###']         []
'@#@'                       ['@#']          []
'####'                      ['####']        []
'## heading with #realtag'  ['##','#realtag'] ['#realtag']
'#realtag and @person'      unchanged       unchanged
```

Full suite on this branch is 694 passed, 44 skipped, and `tests/unit/test_entry.py` is 3 passed.

One correction to the description rather than the code: `@@` is listed as an affected case, but it already returns `[]` on `main`. The trailing class is `[-+*#/\w]`, which contains `#` but not `@`, so `@@` never matched the pattern in the first place. The other three cases do reproduce exactly as described.

